PR #7, opened 2017-04-09
filter is not deprecated

dpsanders:


dpsanders:
It does give deprecation warnings.

PR #12, opened 2017-04-23
0.6 / 0.7 deprecations

dpsanders:


dpsanders:
There doesn't seem to be any way of using `@compat` to use the new style of syntax.
I think this will have to wait until we drop support for 0.5.

lbenet:
I agree with you, and I think that is the general trend: keep the old syntax until dropping support for Julia 0.5.

This is related to https://github.com/JuliaIntervals/ValidatedNumerics.jl/pull/258

lbenet:
Shall I push to this branch some commits which use the `where` syntax in the functions?

lbenet:
Or shall I use a different PR?

dpsanders:
Push to this PR is fine.

lbenet:
I just pushed 4 new commits to have the `where` syntax in the functions. I'll let travis do its work, and if everything is ok, we should merge.

lbenet:
Tests are passing. Merging this...

lbenet:
I think everything is ready for 0.11.0...

dpsanders:
Thanks!

dpsanders:
Tagged 0.11.0!

PR #20, opened 2017-04-28
Fix IntervalBox subtyping

dpsanders:


dpsanders:
@lbenet This is urgent to be merged so that tests work on 0.6.
It is uncontroversial except that it is a bit ugly.

Once 0.6 is released, I suggest we just drop support for 0.5...

src/multidim/intervalbox.jl
```diff
@@ -1,12 +1,22 @@
...
+        data::NTuple{N,Interval{T}}
+    end
+else
+    doc"""An `IntervalBox` is an $N$-dimensional rectangular box, given
+    by a Cartesian product of $N$ `Interval`s.
+    """
+    immutable IntervalBox{N,T} <: StaticVector{Interval{T}}
+        data::NTuple{N,Interval{T}}
+    end
 end
```
lbenet: It seems odd to me that for Julia 0.5 and 0.6, the parameterization of `StaticVector` is different, though I admit that tests pass in Travis (but not locally in my machine, for Julia 0.6). 

Is this related with the requirement of StaticArrays 0.3? I think StaticArrays v0.4.0 deals with some syntax improvements related to Julia 0.6 changes. Shouldn't we change `REQUIRE` to reflect what is needed of StaticArrays?

dpsanders: There is (currently)  no way to specify in REQUIRE that  different versions of StaticArrays.jl should be used on 0.5 and 0.6 without making separate branches. 

We can specify in the docs which version is required and impose the correct minimum version once we drop support for 0.5.

What problem are you having locally?

dpsanders: For 0.6, you need a pretty recent master, and StaticArrays at least 0.4.

dpsanders: (Recent master of Julia 0.6 I mean)

lbenet:
Taking back what I said about the problems. Things are ok again...

dpsanders:
Yes please!

lbenet:
Great! Thanks!

PR #22, opened 2017-04-30
Fix tan

dpsanders:


dpsanders:
Tests are failing for values right on quadrant boundaries, but I don't see what to do about that.

lbenet:
Tests are passing, so I think we should also merge this.

lbenet:
I'm reopening this because the issue isn't really solved. 

Current [master](https://github.com/JuliaIntervals/IntervalArithmetic.jl/commit/2ecefe5864c9e3c0c187a214c8caa84d85f71f54) yields the following:

```julia
julia>    # This examples appears in the tests
x = 2.0^1000
1.0715086071862673e301

julia> Interval(x)
Interval(1.0715086071862673e301, 1.0715086071862673e301)

julia> big(2.0)^1000 ∈ ans   # OK
true

julia> # This seems to be a corner case
x = Interval(2.0)^100
Interval(1.2676506002282294e30, 1.2676506002282294e30)

julia> tan(x)
Interval(-1.7829551493767193, -1.782955149376719)

julia> big(2.0)^100 ∈ ans    # WRONG
false

julia> # Another case
x = Interval(2.0)^60
Interval(1.152921504606847e18, 1.152921504606847e18)

julia> tan(x)  # USEFUL ?
Interval(-Inf, Inf)

julia> prevfloat(tan(x.lo)), nextfloat(tan(x.hi))
(1.4918373962705977, 1.4918373962705977)

julia> tan(big(2.0)^60) # this is contained by the bounds above 
1.491837396270597700887517480636152973049592658670418036396829287651786928528278

```

The problem seems to be in `find_quadrant`, more concretely in [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/trigonometric.jl#L23), which for intervals with large values returns results that are not accurate enough to be useful. (Note that currently, `temp` in that function corresponds to `Interval(Inf,Inf)`, for `Interval(2.0)^1000` in the tests, which makes no sense.)

dpsanders:
I agree that there's an issue with large values of the argument, but I have no idea how to fix it! 

I believe that Siegfried Rump has a method for this.

lbenet:
For the case of `tan` I think I understand where our method is weak; that weekness should somehow also appear elsewhere.

dpsanders:
All the trig functions have problems with large arguments. But I don't think it's worth spending that much effort on myself.

PR #24, opened 2017-05-01
Fix mid using specialized version for alpha=0.5

dpsanders:


lbenet:
@dpsanders It seems that I must [re-tag](https://github.com/JuliaLang/METADATA.jl/pull/9394#issuecomment-302759893); sorry for the mess. 
What I want to bring up is that then we could merge the couple of commits I somehow forgot (this one, #30, and maybe #22) and before the new tag.

PR #25, opened 2017-05-01
Use FastRounding for speed

dpsanders:
Use `FastRounding.jl` as a drop-in replacement for changing the rounding mode. This gives a 3-4x speed-up for basic arithmetic operations.

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -322,7 +322,8 @@ function mid{T}(a::Interval{T}, α)
 
     @assert 0 ≤ α ≤ 1
 
-    return (1-α) * a.lo + α * a.hi  # rounds to nearest
+    # return (1-α) * a.lo + α * a.hi  # rounds to nearest
+    return α*(a.hi - a.lo) + a.lo  # rounds to nearest
 end
 
 
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -326,8 +326,7 @@ function mid{T}(a::Interval{T}, α)
     return α*(a.hi - a.lo) + a.lo  # rounds to nearest
 end
 
-
-function mid{T}(a::Interval{T})  # specialized version for α=0.5
+function mid{T}(a::Interval{T})
 
     isempty(a) && return convert(T, NaN)
     isentire(a) && return zero(a.lo)
@@ -335,12 +334,16 @@ function mid{T}(a::Interval{T})  # specialized version for α=0.5
     a.lo == -∞ && return nextfloat(a.lo)
     a.hi == +∞ && return prevfloat(a.hi)
 
-    return 0.5 * (a.lo + a.hi)  # rounds to nearest
-end
+    # @assert 0 ≤ α ≤ 1
 
+    return simple_mid(a)
+end
 
 mid{T}(a::Interval{Rational{T}}) = (1//2) * (a.lo + a.hi)
 
+function simple_mid(a::Interval)
+    return 0.5*(a.lo + a.hi)
+end
 
 doc"""
     diam(a::Interval)
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -336,14 +336,11 @@ function mid{T}(a::Interval{T})
 
     # @assert 0 ≤ α ≤ 1
 
-    return simple_mid(a)
+    return 0.5 * (a.lo + a.hi)
 end
 
 mid{T}(a::Interval{Rational{T}}) = (1//2) * (a.lo + a.hi)
 
-function simple_mid(a::Interval)
-    return 0.5*(a.lo + a.hi)
-end
 
 doc"""
     diam(a::Interval)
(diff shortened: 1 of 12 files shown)


src/intervals/arithmetic.jl
```diff
@@ -322,24 +322,28 @@ function mid{T}(a::Interval{T}, α)
...
-    return 0.5 * (a.lo + a.hi)  # rounds to nearest
-end
+    # @assert 0 ≤ α ≤ 1
 
+    return simple_mid(a)
+end
 
 mid{T}(a::Interval{Rational{T}}) = (1//2) * (a.lo + a.hi)
 
+function simple_mid(a::Interval)
```
lbenet: Do we really need `simple_mid` as a new function? I think its sole use is inside that method for `mid`.

In any case, it is now good timing to add docstrings and tests.

src/intervals/rounding.jl
```diff
@@ -9,6 +9,7 @@ interval rounding types, e.g.
 +(IntervalRounding{:none}, a, b, RoundDown)
 
 The current allowed rounding types are
+- :errorfree # use errorfree arithmetic via FastRounding.jl
```
lbenet: I suggest the following changes for naming the rounding modes: 
- `:errorfree` should be named `:fast`
-`:fast` should be named `:wider` (or `:wide`?)

lbenet: It should also be mentioned somewhere (the docstrings and perhaps a warning too) that `:errorfree` works only with `Float64` and `Float32`.

lbenet: I shall still refer to `:errorfree` instead of (the new) `:fast` to avoid confusions.

src/intervals/rounding.jl
```diff
@@ -157,23 +199,31 @@ doc"""
     setrounding(Interval, rounding_type::Symbol)
 
 Set the rounding type used for all interval calculations on Julia v0.6 and above.
-Valid `rounding_type`s are `:correct`, `:fast` and `:none`.
+Valid `rounding_type`s are `:correct`, `:fast` and `:none`, `:errorfree`.
```
lbenet: I think this should read as "Valid `rounding_type`s are `:correct`, `:fast`, `:none`, and `:errorfree`".

lbenet:
I forgot to say that this should probably be rebased against current master; the tests related to `cancelminus` and `cancelplus` should all pass.

src/intervals/arithmetic.jl
dpsanders: I think you're right that we no longer need `simple_mid`.

src/intervals/rounding.jl
dpsanders: Agree with `:fast` and `:wide`, thanks for the suggestions.

src/intervals/rounding.jl
dpsanders: Good point.

lbenet:
I just read in the standard (cf page 9, definition of tightness) about what we are discussing as rounding modes. They use the nomenclature "tightest", "accurate" and "valid", which correspond respectively to `:correct`, `:fast` and `:wide`. I mention this in case we want to change the nomenclature to what appears in the standard.

src/intervals/rounding.jl
dpsanders: Following the standard, how about `:errorfree` -> `:fast`; `:fast` -> `:accurate`; `:correct`-> `:tight`?

dpsanders: The problem is that this implies that `:fast` is not `:tight`, which is incorrect. (Errorfree arithmetic does give tight bounds.)

lbenet:
What is faster, `:errorfree` (now `:fast`) or `:fast` (now `:accurate`)? 

I actually thought that the bounds provided by `:errorfree` were tighter than using `prevfloat` and `nextfloat`, and almost as good as `:correct` (now `:tight`).

I suggest to have the docstring in decreasing order of tightness.

dpsanders:
What was `:errorfree` (now `:fast`) is *exactly as accurate as `:tight`* (what was `:correct`). 
In fact, `:tight` (using changes of rounding mode) is now obsolete -- except to check that `:fast` is correct!

dpsanders:
So maybe we can replace `:errorfree` / `:fast` with `:tight`, and `:correct` / `:tight` with `:slow` ;)

lbenet:
Maybe you are right; give me a bit of time, since I thought i had some counterexample.

lbenet:
> So maybe we can replace :errorfree / :fast with :tight, and :correct / :tight with :slow ;)

I just run all ITF1788 suite tests and they pass with :errorfree / :fast. So, I agree: it should be renamed as `:tight`, and :correct / :tight with `:slow`, and the other that involves prevfloat/nextfloat should be `:fast`.

Sorry for being so skeptic...

dpsanders:
I was quite sceptical actually that this would ever be possible. But thanks to @JeffreySarnoff and some hard work fixing difficult corner cases, it seems that it is finally finished!

dpsanders:
So far, I have `:tight`, `:accurate` (the one just using `prevfloat` and `nextfloat`, for which you suggested `:fast`), `:slow` and `:none`. Is that OK with you?

JeffreySarnoff:
!

lbenet:
That's ok with me!

Once travis shows the green lights, I'll merge this.

dpsanders:
Everything is passing!

lbenet:
This is really nice! Merging...

dpsanders:
Thanks!

PR #26, opened 2017-05-01
Make Interval constructor fast by not including checks

dpsanders:
The `IA_VALID` environment variable now determines if validity checks are used. E.g. run Julia with
```
IA_VALID=true julia --compilecache=no
```
to turn on validity checks; by default, they are off.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -34,13 +34,6 @@ end
     end
 end
 
-@benchgroup "Elementary functions" begin
-    a = Interval(1, 2)
-
-    for op in (exp, log, sin, tan)
-        @bench string(op) $(op)($a)
-    end
-end
 
 @benchgroup "Sum" begin
 
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -34,6 +34,11 @@ end
     end
 end
 
+@benchgroup "Elementary functions" begin
+    for op in (exp, log, sin, tan)
+        @bench string(op) $(op)($a)
+    end
+end
 
 @benchgroup "Sum" begin
 
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -5,7 +5,11 @@
 
 ## Interval type
 
-const validity_check = false
+if haskey(ENV, "IA_VALID")
+    const validity_check = true
+else
+    const validity_check = false
+end
 
 abstract type AbstractInterval{T} <: Real end
 
@@ -21,7 +25,7 @@ struct Interval{T<:Real} <: AbstractInterval{T}
                 new(a, b)
 
             else
-                throw(ArgumentError("Must have a ≤ b to construct interval(a, b)."))
+                throw(ArgumentError("Interval of form [$a, $b] not allowed. Must have a ≤ b to construct interval(a, b)."))
             end
 
         end
@@ -60,6 +64,9 @@ Interval{T}(x::Interval) where T = convert(Interval{T}, x)
 Check if `(a, b)` constitute a valid interval
 """
 function isvalid(a::Real, b::Real)
+
+    # println("isvalid()")
+
     if isnan(a) || isnan(b)
         return true
     end
@@ -70,6 +77,10 @@ function isvalid(a::Real, b::Real)
         else
             return false
         end
+    end
+
+    if a == Inf || b == -Inf
+        return false
     end
 
     return true
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -5,7 +5,7 @@
 
 ## Interval type
 
-if haskey(ENV, "IA_VALID") || haskey(ENV, "CI") == true  # CI is for Travis
+if haskey(ENV, "IA_VALID") == true
     const validity_check = true
 else
     const validity_check = false
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -17,7 +17,7 @@ struct Interval{T<:Real} <: AbstractInterval{T}
     lo :: T
     hi :: T
 
-    function Interval{T}(a::T, b::T) where T<:Real
+    function Interval{T}(a::Real, b::Real) where T<:Real
 
         if validity_check
 
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -5,7 +5,7 @@
 
 ## Interval type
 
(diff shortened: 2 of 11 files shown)


lbenet:
What should we do about this? This is already included through the `:accurate` rounding mode.

src/intervals/intervals.jl
```diff
@@ -5,37 +5,34 @@
 
 ## Interval type
 
+const validity_check = false
```
lbenet: I think the default value should be `true`, since that is the case that (eventually) will ensure compliance with the standard. This could be changed by the `IA_VALID` environment variable, which I guess can be set also in the .juliarc file.

src/intervals/intervals.jl
```diff
@@ -57,6 +54,38 @@ Interval{T}(x) where T = Interval(convert(T, x))
...
+function interval(a::Real, b::Real)
+    if !isvalid(a, b)
+        throw(ArgumentError("Must have a ≤ b to construct interval(a, b)."))
+    end
+
+    return Interval(a, b)
+end
+
+interval(a::Real) = interval(a, a)
+
```
lbenet: If I understand `interval` *always* checks the validity of the construction, while `Interval` does not (currently as defualt) unless `IA_VALID=true`. Is this correct?

test/interval_tests/construction.jl
```diff
@@ -43,11 +43,11 @@ using Base.Test
...
-    @test_throws ArgumentError Interval(2, 1)
-    @test_throws ArgumentError Interval(big(2), big(1))
-    @test_throws ArgumentError Interval(BigInt(1), 1//10)
-    @test_throws ArgumentError Interval(1, 0.1)
-    @test_throws ArgumentError Interval(big(1), big(0.1))
+    @test_throws ArgumentError interval(2, 1)
+    @test_throws ArgumentError interval(big(2), big(1))
+    @test_throws ArgumentError interval(BigInt(1), 1//10)
+    @test_throws ArgumentError interval(1, 0.1)
+    @test_throws ArgumentError interval(big(1), big(0.1))
```
lbenet: If my comment about `interval` is correct (`interval` checks the validity and `Interval may not), I think it is a good idea to test the differences among `Interval` and `interval` when their behavior differs.

test/interval_tests/construction.jl
```diff
@@ -253,14 +253,14 @@ end
...
-
-end
+# @testset "Disallow a single NaN in an interval" begin
+#     a = interval(NaN, 2)
+#     @test isnan(a.lo) && isnan(a.hi)
+#
+#     a = interval(Inf, NaN)
+#     @test isnan(a.lo) && isnan(a.hi)
+#
+# end
```
lbenet: Why are these tests disallowed?

src/intervals/intervals.jl
lbenet: Can you also add some docstrings here?

src/intervals/intervals.jl
lbenet: Another reason to change this default is the fact that many users may construct directly the intervals using the constructor instead of `interval`...

benchmark/benchmarks.jl
```diff
@@ -34,13 +34,6 @@ end
     end
 end
 
-@benchgroup "Elementary functions" begin
-    a = Interval(1, 2)
-
-    for op in (exp, log, sin, tan)
-        @bench string(op) $(op)($a)
-    end
-end
```
lbenet: Why is this erased?

src/intervals/intervals.jl
```diff
@@ -5,37 +5,38 @@
 
 ## Interval type
 
+if haskey(ENV, "IA_VALID")
+    const validity_check = true
+else
+    const validity_check = false
+end
+
 abstract type AbstractInterval{T} <: Real end
```
lbenet: (This was not there when I began reviewing.)
My suggestion (see elsewhere) is that the default is to have the validity being checked.

test/interval_tests/construction.jl
dpsanders: I'm not sure, I think they should be added back, thanks.

benchmark/benchmarks.jl
dpsanders: Rebase problem, sorry.

lbenet:
I think that docstrings are still missing. 

I still do not understand the need for the ENV variable.

While we agree to not export `Interval`, why not `AbstractInterval`? I thought the idea of having `AbstractInterval` was to allow other (external?) types use the same functionality. I know it is not yet implemented that way, so currently I agree to not export it; I just want to write this somewhere to be aware of it.

dpsanders:
The ENV variable is the only method I have found to turn on and off the validity checks, such that the code is compiled with *no* validity checks if `validity_check` is `false`.

dpsanders:
@lbenet It would be great if you could have another final look over this. I think I have addressed all the comments and I'm pretty happy with how it's looking.

lbenet:
I've checked and everything seems consistent and nice. I'm actually ready to merge, but I'd like to add the following suggestions:

1. Merging this PR implies that we will have as the default behavior (no `IA_VALID` ENV variable, or `IA_VALID` being whatever except "true") that `validity_check` is `false`. The user *a priori* has to use `interval`, which always checks the validity of the construction of the interval (complying with the standard); that's really nice. Then, why should there be a key `IA_VALID` at all? My suggestion is to get rid completely of it and also of `validity_check`.

2. Following with the previous point, we can also get rid of the current inner constructor (since there is no `validity_check`), and change the current outer constructors to be methods of `interval`. In that way, if the user really wants to use the `Interval` constructor, it will have to be called as `IntervalArithmetic.Interval{T}(a,b)`, with the appropriate `T`, which is nasty enough to be done frequently.

 What do you think about this?

lbenet:
A priori, I am in favor of merging this, and also incorporate my two suggestions above.

lbenet:
Or should the above suggestions be submitted as a different PR?

dpsanders:
I think it makes sense to have the *possibility* of using `validity_check` with every construction of an `Interval`, in case any more bugs in our library code are found.

It's not actually clear to me if unexporting `Interval` is a good idea, since that currently makes using downstream packages (`IntervalRootFinding`) more difficult; but probably we just need to update those in the right way.

lbenet:
Merging, then.

Regarding your comments, if `validity_check` is true, then we are passing through `is_valid_interval` twice. I agree that we should update the rest of the packages so `interval` is used.

PR #30, opened 2017-05-12
Fix mid for IntervalBox to return SVector

dpsanders:


src/multidim/intervalbox.jl
```diff
@@ -31,7 +31,7 @@ IntervalBox(x::Interval) = IntervalBox( (x,) )  # single interval treated as tup
 ## arithmetic operations
 # Note that standard arithmetic operations are implemented automatically by FixedSizeArrays.jl
 
-mid(X::IntervalBox) = [mid(x) for x in X]
+mid(X::IntervalBox) = mid.(X)
```
lbenet: Nice!

PR #31, opened 2017-05-15
WIP: Upgrade documentation and use Documenter.jl

lbenet:
This upgrades the documentation to use Documenter.jl in html format, including the use of doctests. Currently some doctests are not passing (e.g., due `@intervalbox`), but the idea is to have the framework to upgrade it gradually.


lbenet:
I implemented the`TaylorSeries.jl` docs using the repl mode to get the answers automatic. It is certainly possible to change the testing of docstring to the repl mode here, but being in a time of many changes, I think it is safer to test the docstrings.

dpsanders:
I suggest adding an auto-generated API page.

dpsanders:
I suggest just removing the `@intervalbox` stuff from the docs.

lbenet:
Good points! I'll do that and ping you back.

lbenet:
Just updated the docs following your suggestions. The auto-generated API page has an index at the beginning (which we may eventually delete or put elsewhere). The idea is to have a glance of what the docstrings cover, which is not too much (e.g., `Interval` is not documented!).

I will add cross-links in the docs to the relevant entries in the API page.

lbenet:
I just pushed another commit with few crossed links and corrections (thanks to the doctests. 

I just noticed that some entries (in the API autogenerated page) appear repeated). I'll take a look on this later.

dpsanders:
That seems to be due to having several methods (e.g. for intervals and decorated intervals, I guess). I'm not sure if there's a way to consolidate them somehow.

This is looking very good now, thanks! We can merge and fix that later if you prefer.

lbenet:
Sorry for the late answer! I'm merging...

dpsanders:
The online documentation doesn't seem to have worked: https://juliaintervals.github.io/IntervalArithmetic.jl/latest

dpsanders:
There are errors in the doc build, which is presumably why the docs were not deployed: check https://travis-ci.org/JuliaIntervals/IntervalArithmetic.jl/jobs/232963885 and open the second `Documenter` tab (a right-pointing triangle) at the bottom.

lbenet:
Locally, I can't reproduce the problem; maybe it is a Documenter.jl version mistmatch. I'm checking this...

lbenet:
I had an older Documenter.jl version (but not too much); in any case, locally, the doctests pass...

PR #32, opened 2017-05-18
Update coverage badge

lbenet:
Doing it this way to trigger travis and see if the problem with the docs persist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # IntervalArithmetic.jl #
 
 [![Build Status](https://travis-ci.org/JuliaIntervals/IntervalArithmetic.jl.svg?branch=master)](https://travis-ci.org/JuliaIntervals/IntervalArithmetic.jl)
-[![Coverage Status](https://coveralls.io/repos/JuliaIntervals/IntervalArithmetic.jl/badge.svg?branch=master&service=github)](https://coveralls.io/github/JuliaIntervals/IntervalArithmetic.jl?branch=master)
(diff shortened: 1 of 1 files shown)


lbenet:
Here I have only fixed the coveralls badge. The problem was actually not related to the badge itself, but to the settings of coveralls, which I have now fixed. 

I have also [deployed the ssh keys](https://juliadocs.github.io/Documenter.jl/stable/man/hosting.html#Hosting-Documentation-1) and set them in travis (this is related to #31, [this comment](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/31#issuecomment-301942994)). Documenter should build the documentation now, as far as I can see, once this is merged. Note that the docs should be build *after* merging (or committing to master), and *only* from the travis-build using linux and 0.5.

@dpsanders Shall we merge this?

dpsanders:
Thanks, let's hope this works!

lbenet:
It seems everything (badge and docs) finally work: see [here](https://juliaintervals.github.io/IntervalArithmetic.jl/latest/).

The branch gh-pages needs to be cleaned, and perhaps we should tag a new patch version (0.9.1) , so the stable docs are also generated. (Also, the gh-pages branch will need to be cleaned to remove the old docs.)

lbenet:
@dpsanders Do you agree about tagging a patch (0.9.1) version for IntervalArithmetic? The idea is simply to generate the stable docs, and then latter clean-up the gh-pages branch.

dpsanders:
Sure, go ahead.

PR #33, opened 2017-05-19
Remove @intervalbox macro

dpsanders:
I think we should include this in 0.9.1, @lbenet.


lbenet:
This could be merged right away, but I think we should merge first https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/8, which requires some tests and docstrings, and then this one.

dpsanders:
I'm actually not sure that the `@intervalbox` macro is that useful / necessary if all the other functions are written in the correct way.

lbenet:
Then, I'll merge this, and we keep the discussion about having or not `@intervalbox` in https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/8

lbenet:
Thanks, by the way!

dpsanders:
OK, great, thanks!

Issue #34, opened 2017-05-24
Speed up `a..b`


dpsanders:
Don't use complicated conversions -- leave that for `@interval`.
Just do `Interval(prevfloat(a), nextfloat(b))`.

*But* can treat integers exactly:

```jl
..(a::Integer, b::Integer) = Interval(a, b)
..(a::Integer, b::Float64) = Interval(a, nextfloat(b))
..(a::Float64, b::Int) = Interval(prevfloat(a), b)

..(a::Float64, b::Float64) = Interval(prevfloat(a), nextfloat(b))
```

This will take care of most of the useful cases, such as writing `0..\infty`.

dpsanders:
This gives
```
julia> 0..1
[0, 1]

julia> 0.0..1.0
[-4.94066e-324, 1.00001]

```

dpsanders:
See PR #37

PR #37, opened 2017-05-27
Refactor ..

dpsanders:
See #34 for the idea.
There are quite a few tests using `..` and `±` that change slightly.


test/interval_tests/complex.jl
```diff
@@ -14,7 +14,7 @@ end
 
 @testset "Complex functions" begin
```
dpsanders: Yes, we should probably do that for the moment.

src/intervals/intervals.jl
```diff
@@ -67,7 +67,13 @@ include("hyperbolic.jl")
...
 # Syntax for intervals
 
-a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
+# a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
+
+..(a::Integer, b::Integer) = Interval(a, b)
+..(a::Integer, b::Real) = Interval(a, nextfloat(float(b)))
+..(a::Real, b::Integer) = Interval(prevfloat(float(a)), b)
+
+..(a::Real, b::Real) = Interval(prevfloat(float(a)), nextfloat(float(b)))
```
dpsanders: You're probably right. Could you give an example?

test/interval_tests/complex.jl
lbenet: Using this PR I get the following:
```julia
julia> 0.2..0.2
Interval(0.19999999999999998, 0.20000000000000004)

julia> 0.2 ± 0
Interval(0.19999999999999998, 0.20000000000000004)

julia> @interval(0.2)
Interval(0.19999999999999998, 0.2)
```

Current master (for all three versions) yields `Interval(0.19999999999999998, 0.2)
`. My point is that `0.2 ± 0.0` should behave as `@interval` and not as `0.2..0.2`.

dpsanders: The problem is that `@interval` is very slow.

dpsanders: And we only lose one ulp.

lbenet: (I just noticed I answered not in the correct place... Sorry.)

I agree that `a..b` is fast, that at most we loose an ulp, and that `@interval` is slow. My point is that the `±` notation is common in interval arithmetic and has a precise meaning, and thus we have to return the tightest interval possible. The `a..b` notation is sugar, and works fast and correctly to build examples where rounding is not important, like `0..1` or `0..∞`, or whenever getting the tightest interval is not crucial.

dpsanders: OK, I see your point about `±` - I'll change it as you suggest to use `@round_up` and `@round_down`.

dpsanders:
Hmm, I've been thinking about this.

If I write e.g.
```
2.4 ± 0.1
```
then I should actually treat `2.4` as being the interval of floating point numbers nearest to 2.4 (which is one ulp wide), and `0.1` as the interval nearest 0.1.

So then then correct answer would be "2.4 - 0.1", minus one ulp, and "2.4 + 0.1", plus one ulp.
Maybe that is actually equivalent to the rounded version you are suggesting.

dpsanders:
Having played around with this quite a bit, I think the definition in this PR is actually the correct one (thinking about the interpretation treating floating-point numbers in terms of intervals, as above -- cf. unums).

For example, the definition
```
Interval(@round_down(a-b), @round_up(a+b))
```
fails exactly for the case `0.2 ± 0`:

```jl
julia> using IntervalArithmetic; using IntervalArithmetic: @round_up, @round_down

julia> a, b = 0.2, 0.0
(0.2, 0.0)

julia> @format full
6

julia> Interval(@round_down(a-b), @round_up(a+b))
Interval(0.2, 0.2)
```
which does not contain the true `0.2`.

dpsanders:
By the way, I don't actually think that `±` is a common operator in interval arithmetic packages (mainly since most programming languages don't allow it), nor that it has a fixed meaning.

lbenet:
> By the way, I don't actually think that ± is a common operator in interval arithmetic packages (mainly since most programming languages don't allow it), nor that it has a fixed meaning.

I agree, it is not a common operator. Yet, it corresponds to the "midpoint-radius" notation, and therefore it should return the tightest interval.

Regarding your example that involves `@round_down` and `@round_up`, they actually behave differently than `@interval`. The former macros use rounding modes, with the usual subtleties shown above in your examples; `@interval` involves `make_interval` and `transform` which eventually end up with a `convert`.

dpsanders:
So what is your proposal for `±` to make it performant? (Thus having to avoid using `@interval` in its definition.)

I still think that both from a performance point of view and from interpretation point of view where writing `2.3` is not well defined (since that really means the interval (2.3-0.5eps, 2.3+0.5eps), the definition in this PR is the most natural.

dpsanders:
Have we reached a conclusion here? Make `a..b` fast using `accurate` rounding, and keep `a \pm b` as it currently is?

lbenet:
Just to put it somewhere explicitly, essentially we have three options for `m ± r`:
- `(m-r)..(m+r)`, which is fast with wide rigorous bounds (current behavior);
- `@interval(m-r, m+r)`, which is slow ans yields tight rigorous bounds; and
- `Interval(m-r, m+r)`, which is fast, but not necessarily includes the edges meant (due to round off).

I'm not quite convinced whether the current behavior is what it should, but let's see it how the use dictates its behavior. So, let's leave it as it is now, and merge!

Thanks a lot!

PR #39, opened 2017-05-27
Suppress redefinition warnings in setrounding(Interval)

dpsanders:
This suppresses the annoying redefinition warnings spewed out by changing the Interval rounding type via `setrounding(Interval, r)`.

It is careful to suppress *only* those warnings; anything else should immediately throw an error.

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -1,6 +1,6 @@
 # This file is part of the IntervalArithmetic.jl package; MIT licensed
 
-#__precompile__(true)
+__precompile__(true)
 
 module IntervalArithmetic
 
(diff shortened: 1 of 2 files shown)


lbenet:
I understand that the solution proposed here is to dump the warnings to a file, and ignore those related with the redefinition. Isn't it input-output slow in general? Why not allow the user (though some kind of flag) to decide about this?

dpsanders:
True, but I think the input output is fast enough here. This is not a time critical operation, nor one that will happen very often, after all. 

It is possible that the output could be sent to a string instead of a file, which would presumably be faster, but I'm not sure how to do so.

dpsanders:
I'm going to merge this for now.

lbenet:
Fine with me !

dpsanders:
It makes the test output so much nicer to read!

PR #42, opened 2017-05-27
Add pow for interval^real and interval^interval

dpsanders:


lbenet:
Tests are missing; otherwise, I think this is ready to be merged.

dpsanders:
Yes, you're right, it needs tests.

dpsanders:
Rebased. This is now ready.

lbenet:
Yes, I agree that we should merge this; let's just allow the test to complete. Thanks!

lbenet:
Simple curiosity: have you tried to see how many tests of the ITF1788 pass and how many fail, when using `pow` as a replacement for `^`?

dpsanders:
No I haven't tried, but I expect that almost all will fail...

PR #44, opened 2017-05-31
Reimplement cancelminus (without changes of precision)

lbenet:


src/intervals/arithmetic.jl
```diff
@@ -375,6 +375,9 @@ end
     cancelminus(a, b)
 
 Return the unique interval `c` such that `b+c=a`.
+
+See Section 12.12.5 of the IEEE-1788 Standard for
+Interval Arithmetic.
 """
 function cancelminus(a::Interval, b::Interval)
     T = promote_type(eltype(a), eltype(b))
```
dpsanders: This can be written (I believe)
```jl
cancelminus(a::Interval, b::Interval) = cancelminus(promote(a, b))

function cancelminus{T}(a::Interval{T}, b::Interval{T})
   ...
end
```

src/intervals/arithmetic.jl
```diff
@@ -383,19 +386,26 @@ function cancelminus(a::Interval, b::Interval)
...
-            diam(@biginterval(a)) < diam(@biginterval(b))
-        end
-    end
-    ans && return entireinterval(T)
+    c_lo = @round_down(a.lo - b.lo)
+    c_hi = @round_up(a.hi - b.hi)
+
+    c_lo > c_hi && return entireinterval(T)
+
+    c_lo == T(Inf) && return Interval(prevfloat(c_lo), c_hi)
```
dpsanders: No need for `T` here.

src/intervals/arithmetic.jl
lbenet: Sure. I'll change that.

src/intervals/arithmetic.jl
lbenet: Do you get the proper promotion automatically, e.g., for `BigFloat`s?

dpsanders: You mean this?
```jl
julia> Float64(Inf) == BigFloat(Inf)
true
```

lbenet: Ups... that's what I meant. I clean this now.

lbenet:
Just noticed that I generated an ambiguity in cancelminus, which 
> I have to say I don't really understand the logic (and it doesn't seem to be well specified in the standard, from what I can see), but if the tests pass then seems good to me, thanks!

I agree that the standard is not too clear on this.

I'll try to explain what I implemented based on 12.12.5: `cancelminus(a,b)` should return the *tightest* interval `c` such that `a = b+c`. There are some specific cases that return either an empty interval, or an entire one, depending if `a` and `b` are empty or unbounded. A crucial property is that `diam(a)` cannot be smaller than `diam(b)`. The conditionals involving `c_lo` and `c_hi` (which are the calculated bounds for `c`) are included to return `[prevfloat(Inf),Inf]` or `[-Inf,nextfloat(-Inf)]`, if `a` and `b` are both bounded but `c` is too large; without these cases the results would be of the form `[Inf,Inf]` (or the negative). The last if checks if the interval obtained from `b+c` contains `a`, and is as tight as possible; the latter is subtle, and hence the inner conditional. Actually, one of the ITF1788 checks such a case; it turns out that it is possible to construct a `c` interval such that `a ⊆ b+c`, but it is not the tightest possible (hence, an entire interval).

lbenet:
Tests pass in travis; coverage is slightly reduced.

dpsanders:
Thanks!

PR #47, opened 2017-06-03
Refactor Interval{Float64}^x

lbenet:
This PR uses only `Float64` arithmetic to implement a method for `^(a::Interval{Float}, n::Integer)`, i.e., avoids using `big53`.

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -391,9 +391,6 @@ function cancelminus{T<:Real}(a::Interval{T}, b::Interval{T})
 
     c_lo > c_hi && return entireinterval(T)
 
-    c_lo == Inf && return Interval(prevfloat(c_lo), c_hi)
-    c_hi == -Inf && return Interval(c_lo, nextfloat(c_hi))
-
     a_lo = @round_down(b.lo + c_lo)
     a_hi = @round_up(b.hi + c_hi)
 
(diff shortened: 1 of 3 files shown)


dpsanders:
The problem is that the rounded versions of ^ still use BigFloat internally. 

There is some kind of power available in CRlibm that I didn't wrap, but it is known that correctly rounding the power function is very difficult. 

That is why the only solution I see is not to demand tight correct rounding.

lbenet:
> The problem is that the rounded versions of ^ still use BigFloat internally.

This PR replaces the version that uses `big53` in `^(a::Interval{Float64}, n::Integer)` by operations purely with `Float64`, which is the case considered by `pow`. The tests pass, so I would say it is a question of benchmarking against the previous version and `pow`. This version provides tight bounds as far as the ITF1788 tests are concerned.

At the moment I am playing with ``^(a::Interval{Float64}, n::Float64)` to see how far can I get.

lbenet:
This is ready for review; test pass locally!

dpsanders:
I was referring to [these lines](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/b6f352152a99d049bef7cc477ff71512dc2d8f06/src/intervals/rounding.jl#L106-110), where `^(x, y, RoundDown)` is defined in terms of `BigFloat`s.

lbenet:
You are totally right; I forgot the changes in `rounding.jl`...

lbenet:
Closing this, to think this over...

Issue #45, opened 2017-06-01
Bug creating intervals from large floats


dpsanders:
```jl
julia> x = I"1e300"
[∞, ∞]

julia> @interval(1e300)
[∞, ∞]
```

dpsanders:
This comes from
```
julia> convert(Interval, 1e300)
[∞, ∞]
```

lbenet:
Solved by #51. Closing.

PR #51, opened 2017-06-07
Modify constructor to avoid Interval(Inf,Inf) or Interval(-Inf,-Inf)

lbenet:
Solves #45.

### DIFF
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -24,9 +24,12 @@ function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
     # use @round_up and @round_down here?
 end
 
-convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64) =
-    Interval(parse(T, string(x), RoundDown), parse(T, string(x), RoundUp))
-
+function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
+    isinf(x) && return Interval{T}(x)
+    x_str = string(x)
+    return Interval{T}(parse(T, x_str, RoundDown),
+        parse(T, x_str, RoundUp))
+end
 
 convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Interval{T}) = x
 
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -25,10 +25,18 @@ function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
 end
 
 function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
-    isinf(x) && return Interval{T}(x)
-    x_str = string(x)
-    return Interval{T}(parse(T, x_str, RoundDown),
-        parse(T, x_str, RoundUp))
+    isinf(x) && return Interval{T}(prevfloat(x), nextfloat(x))
+
+    xrat = rationalize(x)
+
+    # This prevents that xrat returns a 0//1 when x is very small
+    # or 1//0 when x is too large but finite
+    if (x != zero(x) && xrat == 0) || isinf(xrat)
+        xstr = string(x)
+        return Interval(parse(T, xstr, RoundDown), parse(T, xstr, RoundUp))
+    end
+
+    return convert(Interval{T}, xrat)
 end
 
 convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Interval{T}) = x
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -2,12 +2,6 @@
 
 ## Promotion rules
 
-# Avoid ambiguity with ForwardDiff:
-
-# promote_rule{T<:Real, N, R<:Real}(::Type{Interval{T}},
-#     ::Type{ForwardDiff.Dual{N,R}}) = ForwardDiff.Dual{N, Interval{promote_type(T,R)}}
-
-
 promote_rule{T<:Real, S<:Real}(::Type{Interval{T}}, ::Type{Interval{S}}) =
     Interval{promote_type(T, S)}
 
@@ -30,15 +24,8 @@ function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
     # use @round_up and @round_down here?
 end
 
-function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
-    II = convert(Interval{T}, rationalize(x))
-    # This prevents that rationalize(x) returns a zero when x is very small
-    if x != zero(x) && II == zero(Interval{T})
-        return Interval(parse(T, string(x), RoundDown),
-                        parse(T, string(x), RoundUp))
-    end
-    II
-end
+convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64) =
+    Interval(parse(T, string(x), RoundDown), parse(T, string(x), RoundUp))
 
 
 convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Interval{T}) = x
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -19,13 +19,17 @@ convert{T<:AbstractFloat}(::Type{Interval{T}}, x::AbstractString) =
     parse(Interval{T}, x)
 
 function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
+    isinf(x) && return wideinterval(T(x))
+    # isinf(x) && return Interval{T}(prevfloat(T(x)), nextfloat(T(x)))
+
     Interval{T}( T(x, RoundDown), T(x, RoundUp) )
     # the rounding up could be done as nextfloat of the rounded down one?
     # use @round_up and @round_down here?
 end
 
 function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
-    isinf(x) && return Interval{T}(prevfloat(x), nextfloat(x))
+    isinf(x) && return wideinterval(x)#Interval{T}(prevfloat(T(x)), nextfloat(T(x)))
+    # isinf(x) && return Interval{T}(prevfloat(x), nextfloat(x))
 
     xrat = rationalize(x)
 
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -18,12 +18,18 @@ immutable Interval{T<:Real} <: AbstractInterval
         end
 
         if a > b
-            (isinf(a) && isinf(b)) && return new(a, b)  # empty interval = [∞,-∞]
-
+            # empty interval = [∞,-∞]
+            (isinf(a) && isinf(b)) && return new(a, b)
             throw(ArgumentError("Must have a ≤ b to construct Interval(a, b)."))
         end
 
-
+        # The IEEE Std 1788-2015 states that a < Inf and b > -Inf;
+        # see page 6, "bounds".
+        if a == Inf
+            return new(prevfloat(a), b)
+        elseif b == -Inf
+            return new(a, nextfloat(b))
+        end
 
         new(a, b)
     end
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -25,13 +25,13 @@ immutable Interval{T<:Real} <: AbstractInterval
 
         # The IEEE Std 1788-2015 states that a < Inf and b > -Inf;
         # see page 6, "bounds".
-        if a == Inf
-            return new(prevfloat(a), b)
-        elseif b == -Inf
-            return new(a, nextfloat(b))
+        if a == Inf || b == -Inf
+            throw(ArgumentError(
+                "Must satisfy `a < Inf` and `b > -Inf` to construct Interval(a, b)."))
         end
 
-        new(a, b)
+        return new(a,b)
+
     end
 end
 
--- a/test/interval_tests/construction.jl
+++ b/test/interval_tests/construction.jl
@@ -35,6 +35,10 @@ using Base.Test
     @test Interval{BigFloat}(1) == Interval{BigFloat}(big(1.0), big(1.0))
     @test Interval{BigFloat}(pi) == Interval{BigFloat}(big(pi), big(pi))
 
+    # a < Inf and b > -Inf
+    @test Interval(Inf, Inf) == Interval(prevfloat(Inf), Inf)
+    @test Interval(-Inf, -Inf) == Interval(-Inf, nextfloat(-Inf))
+
     # Disallowed conversions with a > b
 
     @test_throws ArgumentError Interval(2, 1)
--- a/test/interval_tests/construction.jl
+++ b/test/interval_tests/construction.jl
@@ -36,8 +36,10 @@ using Base.Test
     @test Interval{BigFloat}(pi) == Interval{BigFloat}(big(pi), big(pi))
 
     # a < Inf and b > -Inf
-    @test Interval(Inf, Inf) == Interval(prevfloat(Inf), Inf)
-    @test Interval(-Inf, -Inf) == Interval(-Inf, nextfloat(-Inf))
+    @test @interval(1e300) == Interval(9.999999999999999e299, 1.0e300)
+    @test @interval(-1e307) == Interval(-1.0000000000000001e307, -1.0e307)
+    @test @interval(Inf) == Interval(realmax(), Inf)
+    @test Interval(-Inf, -Inf) == Interval(-Inf, -realmax())
 
     # Disallowed conversions with a > b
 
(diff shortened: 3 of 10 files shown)


dpsanders:
I think `Interval(Inf, Inf)` should actually be an error and should be dealt with in the `Interval` constructor itself (or whatever replaces it for checks when constructing an interval).

lbenet:
There are some examples in the ITF1788 test suite that go in a slightly different direction. One involves `cancelminus` (with an equivalent for `cancelplus`) and the other involves `atanh`. Both are already considered, though contrary to what you propose, they do not lead to an error.

The first one is equivalent to #45. In that case it is correct to have `Interval(realmax(),Inf)`.

The second one corresponds to `atanh(1.0, Inf)`. The domain for `atanh` is the interval [-1,1], and `atanh(1.0)` is `Inf`. So, by intersecting with the natural domain `atanh([1.0,Inf])` becomes `atanh([1.0,1.0])` which is the interval `[Inf,Inf]`, hence the requirement that it yields the emptyinterval. This is precisely considered here.

dpsanders:
Ah, I see: `Interval(Inf, Inf)` contains no real numbers, and is hence the empty interval.
In that case, that can also be done right in the constructor of `Interval`, and not by special casing those particular functions.

lbenet:
The case of `atanh` is the only (so far) treated as a special case, since in that case the answer should not be `Interval(prevfloat(Inf),Inf)` as would be done by the constructor, but the empty interval.

lbenet:
I just noticed some inconsistency; hold on a little bit...

lbenet:
Actually, the problem of #45 is not the constructor, but `rationalize`, since `rationalize(1e300)` yields `1//0`.

lbenet:
This PR has a problem: `@interval(1e300)` yields `Interval(1.7976931348623157e308, Inf)`, which is obviously wrong. 

The problem lies in [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/conversion.jl#L33) of a convert method, since `rationalize` for large numbers returns `1//0`, and thus the wrong result.

test/interval_tests/construction.jl
```diff
@@ -35,6 +35,10 @@ using Base.Test
     @test Interval{BigFloat}(1) == Interval{BigFloat}(big(1.0), big(1.0))
     @test Interval{BigFloat}(pi) == Interval{BigFloat}(big(pi), big(pi))
 
+    # a < Inf and b > -Inf
+    @test Interval(Inf, Inf) == Interval(prevfloat(Inf), Inf)
```
dpsanders: I don't understand this line.
`Interval(Inf, Inf)` should either return the empty interval or error.

dpsanders: It is an error in the octave interval package.

lbenet: I agree it should probably yield an error or an empty interval; however, due to `rationalize` in the convert method which is used, we get Interval(Inf,Inf) when it shouldn't. So for the time being I think we should focus in getting around rationalize.

lbenet:
I just pushed another commit (after rebasing to current master). 

Except for three of our tests, all tests pass! I am skipping checking those cases (a test comparing float intervals with the corresponding rationals, a test involving pi, and a test involving large values and `tan`) for the moment, so I can analyze carefully what is going on there.

src/intervals/conversion.jl
```diff
@@ -30,15 +24,8 @@ function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
...
-function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
-    II = convert(Interval{T}, rationalize(x))
-    # This prevents that rationalize(x) returns a zero when x is very small
-    if x != zero(x) && II == zero(Interval{T})
-        return Interval(parse(T, string(x), RoundDown),
-                        parse(T, string(x), RoundUp))
-    end
-    II
-end
+convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64) =
```
dpsanders: This is just far too slow for such a basic operation that is used all over the place, e.g. in `Interval(0.1) * 0.3`:

```
julia> @btime Interval(0.1) * 0.3
  700.447 ns (10 allocations: 416 bytes)
Interval(0.03, 0.03000000000000001)

julia> @btime Interval(0.1)
  7.612 ns (0 allocations: 0 bytes)
Interval(0.1, 0.1)

julia> @btime convert(Interval{Float64}, 0.3)
  610.783 ns (10 allocations: 416 bytes)
Interval(0.3, 0.30000000000000004)

julia> @btime Interval(prevfloat(0.3), nextfloat(0.3))
  12.497 ns (0 allocations: 0 bytes)
```
(A factor of 50 difference!)

I suggest we use `parse` for something like `convert(Interval, "0.3")` (when a string is given), and simply use `convert(::Type{Interval{Float64}, x) = Interval(prevfloat(x), nextfloat(x))` for floats.

lbenet: The only thing I did is to eliminate `rationalize` and the conditional, so that method directly converts to a string which is parsed and rounded.

dpsanders: Unfortunately the conversion to and from a string is very slow, that's what I'm pointing out. (And probably `rationalize` was too.)

dpsanders: Hmm, it seems maybe I'm wrong and this is not the function that is actually used?

lbenet:
I think there are two issues in this discussion we should separate: the construction of an interval (which contains whatever checks we wish), and the use of `convert`.

As for the construction, i.e., `Interval(a,b)` there is essentially no difference. 

As you write, we should improve the `convert` method, which is used whenever we round, i.e., always. Your suggestion is not viable because it does not respect the tightest character that the interval must have.

src/intervals/conversion.jl
dpsanders: No, I'm right.

lbenet: Rationalize has two problems: sometimes it yields a zero for too small floating points, and sometimes it yields Inf for too large values. I'll see if I can improve the speed...

dpsanders:
Well, part of the problem as usual is that it is not clear what one means when writing 0.3 -- do you mean the decimal number "0.3", or the floating point number that is printed as 0.3.

As far as I can tell, the standard only talks about what to do with strings, in which case certainly we must give the tightest interval, however long that takes: that is, if the user uses a string, it is a sign that (s)he wants the tightest possible interval.

With floats there is, as far as I can tell, basically no way of making it performant to get the tightest interval if you interpret 0.3 as "the exact decimal number 0.3". If you interpret 0.3 as "some 
real number that is closest to the floating-point number 0.3" (i.e. an uncertain unum-like object) then it actually represents a small interval of width eps(0.3)/2 (roughly), and so it makes total sense to do outward rounding with `prevfloat` and `nextfloat`.

lbenet:
I disagree: the purpose of the standard is to set-up definitions and rules to get the same; returning a tightest interval is part of it.

Quoting from the standard (italics are mine): 

"The aim of the standard is to improve the availability of reliable computing in modern hardware and software environments by defining the basic building blocks needed for performing interval arithmetic. There are presently many systems for interval arithmetic in use; *lack of a standard inhibits development, portability, and ability to verify correctness of codes*."

" This standard specifies:
...
Constructors for intervals from numeric and character sequence data"

dpsanders:
The relevant section of the standard is 10.5.8 (pg. 38): there are only two required constructors, namely (in our notation):

- `Interval(a, b)` for floats `a` and `b`
- `textToInterval(s)`, for a text string `s`

(We certainly need to work on the second one for it to be compliant with the standard.)

dpsanders:
Here's another relevant piece (pg. 30):
"An implementation may support an extended form of literals, e.g., using number literals in the syntax of the host language of the implementation. *It may restrict the support of literals at Level 2, by relaxing conversion accuracy of hard cases*: rational number literals, long strings, etc. What extensions and restrictions of this kind are permitted is flavor-defined."

(Italics are mine.)

lbenet:
All clause 10 is related to Level 1 description, i.e., to the mathematical theory; the level 2 description (finite precision) corresponds to the clause 12. The measures of accuracy is treated in section 12.10. The mode we are discussing is called `accurate` and is the intermediate one with respect to the tightness.

lbenet:
Perhaps we should use traits also for the constructors...

dpsanders:
Yes, that is certainly a possibility. What did you have in mind? Something like
```
Interval(:tight, 0.3)
```
vs 
```
Interval(0.3)
```
which may not be tight, could work.

Or
```
Interval(Tight(), 0.3)
```
to use dispatch.

dpsanders:
Also pg. 60:
"An implementation may provide more than one version of some operations for a given type. For instance, it may provide an “accurate” version in addition to a required “tightest” one, to offer a trade-off of accuracy versus speed or code size. How such a facility is provided is language- or implementation-defined."

lbenet:
Regarding traits, I was thinking in something like `Interval{T, RM}`, where `RM` is a rounding mode (e.g., `:tight`), but you know all these things better than I do.

Regarding the last quotation you provided, the key part for me is "... in addition to a required “tightest” one,...". That's the reason I think we should focus on this one, and the allow for others.

lbenet:
It's not ready, actually. As it is now, it breaks some tests (`cancelminus`) and maybe more... I have solved the cancelminus problem, but I'm trying to make that it is a bit faster... Difficult...

dpsanders:
OK, thanks, no problem.

lbenet:
I just push the current status. There are three tests that don't pass (currently using `@test_skip`), due to the changes here. 

The one I am most concerned is what you fixed by #22... As far as I can trace it, the reason that #22 works is because at some point you get `Interval(Inf,Inf)`, which now is `Interval(realmax(),Inf)`, and that change yields to a much wider interval that results in an entireinterval...

dpsanders:
I'm not sure why the result for `tan` changed (do you understand why?), but I see no problem with returning `[-Inf, Inf]` for `x = Interval(2)^1000`. (But apparently that causes other issues?)

dpsanders:
Could you give the code sequence that gives `Interval(Inf, Inf)`? I don't understand where that is.

lbenet:
> I'm not sure why the result for tan changed (do you understand why?), but I see no problem with returning [-Inf, Inf] for x = Interval(2)^1000. (But apparently that causes other issues?)

I agree that `[-Inf,Inf]` contains the solution, but in master the test case returns a tighter interval which contains the solution. That is what I wanted to reproduce.

> Could you give the code sequence that gives Interval(Inf, Inf)? I don't understand where that is.

This is my understanding: `find_quadrants` creates the interval `temp` and rounds to the previous (smaller) integer the bounds of the interval. In master, the line `temp = x / half_pi(x)` returns `Interval(Inf,Inf)`, which is wrong(!). Then, `lo_quadrant_mod` and `hi_quadrant_mod` which are `Inf` succeed skipping all if's and then the returned interval is evaluated as `@round(tan(a.lo), tan(a.hi))`, which happens to be a tighter interval (than the entire one), seemingly containing the true answer. 

Yet, this PR is excluding precisely intervals of the form `Interval(Inf,Inf)`. Then, `temp` becomes `Interval(6.821435655968254e300, 6.821435655968257e300)`, which is such a wide interval that the interval returned is the entire interval.

The current result `[-Inf,Inf]` is correct, but maybe we can do better...

dpsanders:
The line with temp only ever acts on floats, not on intervals?

lbenet:
One of the tests that is not passing (due to this PR) is [this one](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/near_infinity/test/interval_tests/numeric.jl#L151). This, I think, is related to the changes in `convert`, and more specifically, to the fact that it doesn't use anymore `rationalize`. 

What troubles me is the following:

```julia
julia> x = 1/3
0.3333333333333333

julia> @interval(x)
Interval(0.33333333333333326, 0.3333333333333333)

julia> big(1)/3 ∈ ans  # AUCH!!!
false
```

On the other hand:
```julia
julia> big(1)/3
3.333333333333333333333333333333333333333333333333333333333333333333333333333348e-01

julia> @interval(1/3)
Interval(0.3333333333333333, 0.33333333333333337)

julia> big(1)/3 ∈ ans
true
```

So rationalize was somehow helping. Any idea?

To go back to `rationalize` means to have another conditional when `rationalize` yields `1//0` which still can be handled as a Float64 number, aside from the performance issue.

lbenet:
> The line with temp only ever acts on floats, not on intervals?

I agree, but it creates an interval; in master we get:
```julia
julia> x = 2.0^1000
1.0715086071862673e301

julia> x/IntervalArithmetic.half_pi(x)
Interval(Inf, Inf)
```

dpsanders:
Not at my computer currently but I thought I tried that and did not get that result. That would definitely be a bug, in division apparently?

lbenet:
i haven't checked the division, but I think the problem is related to #45; somewhere we must rely on `convert(Interval{Float64}, x)` with `x` a large enough number.

dpsanders:
Ah yes, good catch: since x is divided by an Interval,  the first thing that happens is that it gets converted to an Interval.

src/intervals/conversion.jl
```diff
@@ -31,16 +25,12 @@ function convert{T<:AbstractFloat, S<:Real}(::Type{Interval{T}}, x::S)
...
 
 function convert{T<:AbstractFloat}(::Type{Interval{T}}, x::Float64)
-    II = convert(Interval{T}, rationalize(x))
-    # This prevents that rationalize(x) returns a zero when x is very small
-    if x != zero(x) && II == zero(Interval{T})
-        return Interval(parse(T, string(x), RoundDown),
-                        parse(T, string(x), RoundUp))
-    end
-    II
+    isinf(x) && return Interval{T}(x)
```
dpsanders: If x is infinite, then it should be an error to convert it to an Interval (though this can happen in the constructor itself).

src/intervals/intervals.jl
```diff
@@ -18,12 +18,18 @@ immutable Interval{T<:Real} <: AbstractInterval
...
+            # empty interval = [∞,-∞]
+            (isinf(a) && isinf(b)) && return new(a, b)
             throw(ArgumentError("Must have a ≤ b to construct Interval(a, b)."))
         end
 
-
+        # The IEEE Std 1788-2015 states that a < Inf and b > -Inf;
+        # see page 6, "bounds".
+        if a == Inf
+            return new(prevfloat(a), b)
```
dpsanders: a == Inf should give an error.

src/intervals/conversion.jl
lbenet: Good point. Currently this was returning one of the edge intervals, e.g. `Interval(realmax(), Inf)`, which was handled at the constructor. The idea to include that here was to skip the rest of the method. (I will change the constructor as you suggest below, and the adapt this.)

src/intervals/intervals.jl
lbenet: There is a case (`cancelminus` I believe) where the subtlety is important. 

In any case, I think you are right. I will change this to throw an error (ArgumentError?) and treat the special case separately.

lbenet:
@dpsanders Thanks for the review. I will incorporate the changes. Should I also clean-up (cherry-pick and push -f) this PR?

dpsanders:
Do whatever needs doing. Is `rebase` not enough?

lbenet:
Pushing what I have now; it's messy (with respect to the end result) but I think it works fine. I want to let travis do its job.

lbenet:
Tests are passing, including few additions :)

So this is a summary of what I did: 
- I modified the constructor as we discussed, so `Interval(Inf,Inf)` throws an error;
- `rationalize` is back, so `h=1/3; @interval(h * h)` yields an interval containing `big(1)/9`;
- I defined the (unexported) function `wideinterval(x)` which returns the interval constructed with `prevfloat(x)` and `nextfloat(x)`;
- I included conditionals of the form `(a.lo == Inf || a.hi ==-Inf) && return wideinterval(a.lo)`, which is sometimes needed (e.g. `^`).

The following is a feature: `@interval(Inf) == wideinterval(Inf)`.

lbenet:
Incidentally, `@round(a, b)` may lead to headaches in case `a` and `b` may become *both* `Inf`...

lbenet:
Only the test introduced in #22 is skipped; the result of `tan( Interval(2.0)^1000 )` is the entire interval...

dpsanders:
This is looking very nice, thanks!

What was the problem with the power with infinite limits?

In my opinion this is ready to merge. If you agree, then please go ahead!

lbenet:
The problem is that, sometimes, `@round(a,b)` may lead to `Interval(Inf,Inf)`; this was throwing an error (triggered by the constructor) despite that it could be to a computable interval. That's the reason I had to split the operations and use `@round_down` and `@round_up`. (I tried to do it directly in the macros, but was unable.) We should keep this in mind.

Another comment: as far as I can see, `rationalize` is really required by `@interval`, but seemingly not when using `@round`. The only tests that were failing without it were things like `@interval( h*h )` for `h=1/3`. Maybe we can skip using it in the `convert` method where it is now by adding a specialize convert method which `@interval` will use. What do you think about this?

Aside from the last comment, which may be left for another PR, I agree that this should be merged.

lbenet:
Merging!

dpsanders:
Great, thanks!

PR #54, opened 2017-06-08
Add basic benchmarks

dpsanders:


dpsanders:
This is ready.

lbenet:
Good idea to add this! I suggest to include benchmarks such as `@interval(0.1)`, since everything related to rounding would also be included, and perhaps also one related to the calculation for pi.

PR #57, opened 2017-06-11
Update itf1788 tests to use Base.Test

lbenet:
I just benchmarked (using Julia 0.6.0-rc3) the ITF1788 tests using `FactCheck.jl` and using `Base.Test`, and the latter is much shorter, almost by a factor 2! This motivates this PR. 

@dpsanders Can you confirm this?


lbenet:
It seems that the problem with Base.Test persists in Julia0.5; so we should live this PR in the fridge until we drop Julia 0.5 support. Hopefully soon...

dpsanders:
Wow, I confirm that `Base.Test` is now fast (roughly twice as fast, as you say)!

We could release a version that requires Julia 0.6-dev.

PR #66, opened 2017-06-22
Try changing all >= zero(T) to use signbit in multiplication

dpsanders:


dpsanders:
This doesn't  seem to make any difference in performance.

PR #68, opened 2017-06-22
Speed up multiplication and division

dpsanders:
This change makes multiplication of 2 intervals almost twice as fast! And division 30% faster (24 ns instead of 35).

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -113,7 +113,7 @@ function /(a::Interval{T}, b::Interval{T}) where T<:Real
 
     S = typeof(a.lo / b.lo)
     (isempty(a) || isempty(b)) && return emptyinterval(S)
-    b == zero(b) && return emptyinterval(S)
+    iszero(b) && return emptyinterval(S)
 
     if b.lo > zero(T) # b strictly positive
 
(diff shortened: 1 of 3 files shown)


dpsanders:
One of the reasons for the slowness was again the time required to construct an `Interval`, here `zero(a)`.

src/intervals/arithmetic.jl
```diff
@@ -78,7 +78,7 @@ end
 function *(a::Interval{T}, b::Interval{T}) where T<:Real
     (isempty(a) || isempty(b)) && return emptyinterval(T)
 
-    (a == zero(a) || b == zero(b)) && return zero(a)
+    (iszero(a) || iszero(b)) && return zero(Interval{T})
```
lbenet: Neat!

src/intervals/special.jl
```diff
@@ -75,6 +75,8 @@ This occurs when the interval is empty, or when the upper bound equals the lower
 """
 isatomic(x::Interval) = isempty(x) || (x.hi == x.lo) || (x.hi == nextfloat(x.lo))
 
+Base.iszero(x::Interval) = iszero(x.lo) && iszero(x.hi)
```
lbenet: Have you tried `@inline` here? This may yield further performance...

lbenet:
Thank you! Very nice that such a small change has that impact! Have you checked whether `a == zero(a)` is used elsewhere, to profit and get rid of that?

src/intervals/special.jl
dpsanders: A small function like this should be automatically inlined I believe.

dpsanders:
I haven't checked that; it's a good idea!

lbenet:
Thanks! Waiting for travis to complete its job...

lbenet:
Tests pass; merging this! Thanks!

PR #76, opened 2017-06-26
No need for version references

ChrisRackauckas:
These seem like they were from a long time ago, and so they don't need to be referenced in the current docs. Doing different releases of the docs will handle this more naturally in the future.


dpsanders:
Thanks!

PR #78, opened 2017-06-27
Make convert fast and non-allocating

dpsanders:


lbenet:
Am I right saying that this change induces loosing tight rounded intervals?

dpsanders:
Yes, there is no way (as far as I can see) to make those performant. 

This can be one of those things that we can allow the usrler to set via an option (though I don't know what the API should look like).

dpsanders:
Doing the opposite in #97.

Issue #72, opened 2017-06-25
prev_float(::BigFloat) is missing


kalmarek:
at https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/rounding.jl#L131
arithmetic operations are defined for all `T<:AbstractFloats`, but `prev_float` (in `AdjacentFloats.jl`) is defined only for `Float{16, 32, 64}`.
This gets emanated in 

```
ERROR: LoadError: MethodError: no method matching prev_float(::BigFloat)
Closest candidates are:
  prev_float(!Matched::Float16) at /home/kalmar/.julia/v0.5/AdjacentFloats/src/AdjacentFloats.jl:35
  prev_float(!Matched::Float32) at /home/kalmar/.julia/v0.5/AdjacentFloats/src/AdjacentFloats.jl:31
  prev_float(!Matched::Float64) at /home/kalmar/.julia/v0.5/AdjacentFloats/src/AdjacentFloats.jl:21
  ...
 in *(::IntervalArithmetic.Interval{BigFloat}, ::IntervalArithmetic.Interval{BigFloat}) at /home/kalmar/.julia/v0.5/IntervalArithmetic/src/intervals/arithmetic.jl:84
```
(this used to work on v0.9.1) 

```
Julia Version 0.5.2
Commit f4c6c9d4bb* (2017-05-06 16:34 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (NO_LAPACK NO_LAPACKE NO_AFFINITY HASWELL)
  LAPACK: liblapack
  LIBM: libm
  LLVM: libLLVM-3.9.1 (ORCJIT, skylake-avx512)
```
IntervalArithmetic v0.10.0

dpsanders:
Oops, thanks for catching this. I'll fix tomorrow

dpsanders:
This just needs a definition
`AdjacentFloats.prev_float(x) = prevfloat(x)`

kalmarek:
Thanks. Don't forget to define `nextfloat` as well ;-)

dpsanders:
What context did you find this in? Can this be turned into a test case?
These should probably be defined directly in the AdjacentFloats package.

dpsanders:
Yes that would be great, thanks!

[...]

PR #86, opened 2017-07-13
catch missing prev_float

kalmarek:
Following https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/72


dpsanders:
Thanks for the contribution!

PR #88, opened 2017-08-03
Fix issue with method redefinitions

dpsanders:


dpsanders:
Changes on nightly led to extra warnings for deprecations, which were not taken into account. This fixes that to still report the warnings, but not error.

lbenet:
Thanks! Merged!

PR #91, opened 2017-12-15
hull for IntervalBoxes

AKS1996:
analogous to hull for intervals

### DIFF
--- a/src/intervals/rounding.jl
+++ b/src/intervals/rounding.jl
@@ -268,9 +268,24 @@ function setrounding(::Type{Interval}, rounding_type::Symbol)
 
     # check
     lines = readlines(path)
+
+    # print(lines)
+
     for line in lines
+
+        if line == ""
+            continue
+        end
+
         if !startswith(line, "WARNING: Method definition")
-            error(line)
+
+            if startswith(line, "WARNING") || startswith(line, "Use")
+                warn(line)
+
+            else
+                println("Error on line: ", line)
+                error(line)
+            end
         end
     end
 
@@ -279,6 +294,7 @@ function setrounding(::Type{Interval}, rounding_type::Symbol)
 end
 
 rounding(Interval) = current_rounding_type[]
+
 
 
 # default: correct rounding


dpsanders:
Thanks very much for the PR! This is definitely functionality that we need.
I will add some comments on the code.

You will also need to add some tests.

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
 #
 #     return results_list
 # end
+
+import IntervalArithmetic.hull
```
dpsanders: You don't need this, since we're still in the `IntervalArithmetic` module. You're just adding another method to the `hull` function.

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
...
+doc"""
+    hull(a, b)
+
+Returns the `box hull` of the IntervalBoxes `a` and `b`, considered as
+(extended) $N$ dimensional sets of real numbers, i.e. the smallest region that contains
```
dpsanders: I would say "the smallest `IntervalBox` that contains".

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
 #
 #     return results_list
 # end
+
+import IntervalArithmetic.hull
+doc"""
+    hull(a, b)
+
+Returns the `box hull` of the IntervalBoxes `a` and `b`, considered as
```
dpsanders: Backticks (`` ` ``) are used for code, so `IntervalBox` should have backticks. "Box hull" could have normal quotes.

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
...
+all of `a` and `b`.
+"""
+function hull(A::IntervalBox{N,T}, B::IntervalBox{N,T}) where {N,T}
+    result_list = Vector{Interval{T}}()
+
+    global_lo = convert(T,0)
+    global_hi = convert(T,0)
+
+    for dimension in 1:N
+        global_lo = min(Base.getindex(A,dimension).lo,Base.getindex(B,dimension).lo)
```
dpsanders: Note that `Base.getindex(A, dimension)` can (and should) be written as just `A[dimension]`.

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
...
+(extended) $N$ dimensional sets of real numbers, i.e. the smallest region that contains
+all of `a` and `b`.
+"""
+function hull(A::IntervalBox{N,T}, B::IntervalBox{N,T}) where {N,T}
+    result_list = Vector{Interval{T}}()
+
+    global_lo = convert(T,0)
+    global_hi = convert(T,0)
+
+    for dimension in 1:N
```
dpsanders: In this loop, you are calculating the `hull` of each pair of intervals, so it can be just written as `hull(A[i], B[i])`. I would use `i` or `n` instead of `dimension` for the loop variable just for readability.

src/multidim/setdiff.jl
```diff
@@ -105,3 +105,25 @@ end
...
+import IntervalArithmetic.hull
+doc"""
+    hull(a, b)
+
+Returns the `box hull` of the IntervalBoxes `a` and `b`, considered as
+(extended) $N$ dimensional sets of real numbers, i.e. the smallest region that contains
+all of `a` and `b`.
+"""
+function hull(A::IntervalBox{N,T}, B::IntervalBox{N,T}) where {N,T}
+    result_list = Vector{Interval{T}}()
```
dpsanders: Here you are creating an intermediate array that it turns out is actually unnecessary. There is a function `ntuple` that makes a tuple out of a function that you can use here. This is used in a few places in the `IntervalBox` code.

dpsanders:
Also, the `setdiff.jl` file was really meant for the `setdiff` functionality.
Maybe this should be moved to `intervalbox.jl`?  Or maybe that file should now be split up more.

dpsanders:
I have just looked at the code again, and I think this functionality actually already exists, just written as `∪`. Could you please confirm that?

Surely the documentation could be improved on this point, though!

dpsanders:
On this line: https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/d442c3f1414219972ee56c36681f7f041ac3dee3/src/multidim/intervalbox.jl#L32

AKS1996:
True. The union `U` does the same thing

AKS1996:
But should `U`, **mathematically, not computationally concerned**, return a disjoint set of `IntervalBox`es not conatining the extra space as in the hull
untitled

dpsanders:
Mathematically yes, I agree. This can be achieved using the setdiff function.

On the other hand, it is not clear why you would want this, computationally speaking - what would it be useful for? (I'm not saying that it isn't useful, I'm just not sure what the context would be.)

AKS1996:
I don't know it's use. Just took my interest

PR #92, opened 2017-12-16
gH difference as alternative to cancelminus, cancelplus 

AKS1996:
Ref [Properties of interval-valued function space under the gH-difference, Springer Open](https://doi.org/10.1186/s13662-016-0759-9)

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -424,3 +424,36 @@ isinteger(a::Interval) = (a.lo == a.hi) && isinteger(a.lo)
 
 convert(::Type{Integer}, a::Interval) = isinteger(a) ?
         convert(Integer, a.lo) : throw(ArgumentError("Cannot convert $a to integer"))
+
+        doc"""
+        	a⊖b
+        Returns the gH(generalized Hukuhara) difference of the compact `Interval`s `a` and `b`
+        Note that in general if x<:Interval, x-x ≠ 0, but x⊖x = 0
+        """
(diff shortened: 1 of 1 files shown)


dpsanders:
Thanks for the PR! I haven't had a chance to look at the article yet (but thanks for the reference). Could you contact on how exactly these are different from cancelplus and cancelminus? 

As with all interval functions in the library, these will need to use correct rounding, which you should be able to do with the `@round` macro.

You will also need to add some tests. 

Thanks!

dpsanders:
Had a quick look at the paper. They don't seem to make any comments about correct rounding, so I doubt that their results are rigorous.

If this is the case, then it is not suitable for IntervalArithmetic.jl, and I would suggest that you could make a separate package that eg. extends this one instead.

lbenet:
I took also a quick look in the paper. 

From what I read, it seems to me that what they called there as the H difference corresponds to `cancelminus`, and what they refer as the gH difference is a generalization when certain condition on the widths of the intervals (needed to define `cancelminus`) does not hold. In that case, they seem to use something which looks like the definition of `cancelplus`.

I think the issue of rounding can be properly handled from the definitions (at the end, it is a question of additions or substractions), but that should be thoroughly tested, in the spirit of the ITF1788 tests. I certainly agree that tests must be present.

AKS1996:
Sorry. Got the IEEE standard just now. The paper referred doesn't conform with the standard and will (probably) fail all of them

AKS1996:
By the way, thanks for the response. I'll keep a check that the code complies with the standard next time.

PR #97, opened 2018-02-26
Revert dotdot

dpsanders:
- Change definition of `..` back to give tight intervals.
- Add `convert(Interval, s::String)`

### DIFF
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -122,7 +122,7 @@ include("hyperbolic.jl")
 
 # Syntax for intervals
 
-a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
+a..b = interval(convert(Interval, a).lo, convert(Interval, b).hi)
 
 # ..(a::Integer, b::Integer) = interval(a, b)
 # ..(a::Integer, b::Real) = interval(a, nextfloat(float(b)))
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -122,13 +122,13 @@ include("hyperbolic.jl")
 
 # Syntax for intervals
 
-# a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
+a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
 
-..(a::Integer, b::Integer) = interval(a, b)
-..(a::Integer, b::Real) = interval(a, nextfloat(float(b)))
-..(a::Real, b::Integer) = interval(prevfloat(float(a)), b)
-
-..(a::Real, b::Real) = interval(prevfloat(float(a)), nextfloat(float(b)))
+# ..(a::Integer, b::Integer) = interval(a, b)
+# ..(a::Integer, b::Real) = interval(a, nextfloat(float(b)))
+# ..(a::Real, b::Integer) = interval(prevfloat(float(a)), b)
+#
+# ..(a::Real, b::Real) = interval(prevfloat(float(a)), nextfloat(float(b)))
 
 macro I_str(ex)  # I"[3,4]"
     @interval(ex)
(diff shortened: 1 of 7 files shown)


src/intervals/intervals.jl
```diff
@@ -122,13 +122,13 @@ include("hyperbolic.jl")
 
 # Syntax for intervals
 
-# a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
+a..b = Interval(convert(Interval, a).lo, convert(Interval, b).hi)
```
lbenet: Perhaps, we need `interval` here.

lbenet:
Some tests need revision...

lbenet:
Three tests are failing, related to the mid-point radius notation (using `±`). The first two are easily solved by:
```julia
    @test parse(Interval{Float64}, "3 ±  4") == Interval(-1, 7)
    @test parse(Interval{Float64}, "0.2 ± 0.1") == @interval(0.1,0.3)
```
which is intuitively what they should do.

The [last one](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/1b59cc606aec4223393609e2ed7afc7e78932e97/test/interval_tests/parsing.jl#L21) is trickier and I think some changes in the code itself are needed, perhaps [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/1b59cc606aec4223393609e2ed7afc7e78932e97/src/parsing.jl#L46).

In my opinion, `±` should behave (for `BigFloat`s) just as `@biginterval` does, so that test would correspond to 
```julia
@test parse(Interval{BigFloat}, "0.2 ± 0.1") == @biginterval(0.1,0.3)
```
Yet, this test also fails, because  `a` (and `b`) is parsed as a `BigFloat`. In this case, `(a-b)..(a+b)` will not contain the true lower bound due to rounding issues.

dpsanders:
I agree with you that it would be nice to have that, but it seems difficult.

Note that we have some other problems:
```
julia> @biginterval(0.1, 0.3)
Interval(1.000000000000000055511151231257827021181583404541015625000000000000000000000000e-01, 2.999999999999999888977697537484345957636833190917968750000000000000000000000000e-01)

julia> @biginterval("0.1", "0.3")
Interval(1.000000000000000055511151231257827021181583404541015625000000000000000000000000e-01, 2.999999999999999888977697537484345957636833190917968750000000000000000000000000e-01)
```

dpsanders:
I think the latest changes fix this.

lbenet:
Now, other tests are failing: [this one](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/1b59cc606aec4223393609e2ed7afc7e78932e97/test/interval_tests/construction.jl#L68) and [this one](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/1b59cc606aec4223393609e2ed7afc7e78932e97/test/interval_tests/construction.jl#L77), at least...

lbenet:
Some tests still fail. The failure, as far as I have checked locally, is only related to the display of intervals (i.e., `display_tests/display.jl`). Hopefully, some lefts-over only...

lbenet:
Locally, everything seems to work fine! 

I propose to wait for travis, merge and tag a new version! The question is if it should be a patch or a minor version?

dpsanders:
Since we changed the meaning of `..` I think it should be a minor version.

lbenet:
Excellent!

dpsanders:
The string parsing code in general needs an overhaul.
This should be done by checking what the basic standard and the full standard require and implementing exactly that.

E.g. they require being able to parse `2/3` inside an interval string. Currently this is done by using Julia's `parse`, but we should avoid that in the future and do it ourselves.

I have a candidate in mind to do all of this work...

lbenet:
I agree with you. But, let's leave it for another PR! The first require to make https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/24 work is tag a new release here!

dpsanders:
Tests are passing on Travis! Though the doc generation seems to be broken... :/

lbenet:
Thanks a lot!

PR #106, opened 2018-03-12
Use broadcasting consistently for multidim operations

dpsanders:
This gives significant speedups for several of the functions.


lbenet:
Thanks! Very nice that you noticed this spot for improvement! Tests pass (code coverage is reduced by a really small amount), so I am merging this!

PR #111, opened 2018-04-05
Add extended_div()

eeshan9815:
Add function `extended_div()` for extended division of two intervals


src/intervals/arithmetic.jl
```diff
@@ -352,7 +407,7 @@ Return the diameter (length) of the `Interval` `a`.
 function diam(a::Interval{T}) where T<:Real
     isempty(a) && return convert(T, NaN)
 
-    @round_up(a.hi - a.lo) # cf page 64 of IEEE1788
+    Interval_up(a.hi - a.lo) # cf page 64 of IEEE1788
```
dpsanders: What is this change for?

src/intervals/arithmetic.jl
```diff
@@ -153,6 +153,61 @@ function /(a::Interval{T}, b::Interval{T}) where T<:Real
     end
 end
 
+function extended_div(a::Interval{T}, b::Interval{T}) where T<:Real
```
dpsanders: This function seems far too complicated -- I think it should just be able to use the `/` function that is already defined.

dpsanders: Although maybe your version is slightly more efficient -- would need benchmarks to check.

eeshan9815: Should I make changes to the `/` function instead?

src/intervals/arithmetic.jl
eeshan9815: That was by mistake. Fixed.

dpsanders:
As far as I can tell, you are violating a key "rule" of Julia -- rule in the sense of for having good performance: a function should always return the same output type if it has the same input type.

Here you seem to be returning a vector of `Interval`s or a single `Interval` from the same function.

Also, you should return a tuple, rather than an array.

One possibility would be to e.g. return the second interval as the empty set, and/or return an integer or`Symbol` that says how many intervals are being returned.

eeshan9815:
@dpsanders 
Is it fine now?
I am returning the second interval as an empty set and returning a tuple instead of an array. Also, I implemented the function using `/`.

test/interval_tests/numeric.jl
```diff
@@ -35,17 +35,17 @@ setprecision(Interval, Float64)
...
+    @test extended_div(@interval(-30,-15) , @interval(-5,-3)) == (Interval(3.0, 10.0), emptyinterval(c))
+    @test extended_div(1.0..2.0, 0.1..1.0) == (Interval(1, 20.000000000000004), emptyinterval(c))
+    @test extended_div(a, c) == (Interval(2.4999999999999998e-02, 4.4e+00), emptyinterval(c))
+    @test extended_div(c, Interval(4.0)) == (Interval(6.25e-02, 1e+00), emptyinterval(c))
+    @test extended_div(c, zero(c)) == (emptyinterval(c), emptyinterval(c))
+    @test extended_div(Interval( 0.0, 1.0), Interval(0.0,1.0)) == (Interval(0.0, Inf), emptyinterval(c))
+    @test extended_div(Interval(-1.0, 1.0), Interval(0.0,1.0)) == (entireinterval(c), emptyinterval(c))
+    @test extended_div(Interval(-1.0, 1.0), Interval(-1.0,1.0)) == (entireinterval(c), emptyinterval(c))
+    @test extended_div(Interval(1.0, 2.0), Interval(-4.0, 4.0)) == ((-∞.. -0.25), (0.25..∞))
+    @test extended_div(Interval(-1.0, -2.0), Interval(-2.0, 4.0)) == ((-∞.. -0.5), (1.0..∞))
```
dpsanders: You should not use `Interval` to construct intervals, but rather `interval`. 
Here, for example, `Interval(-1.0, -2.0)` gives an *invalid* interval since -1 > -2.

```
julia> interval(-1, -2)
ERROR: ArgumentError: `[-1, -2]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
```

dpsanders: This test seems to be wrong:
```
julia> (-2.. -1) / (-2..0)
[0.5, ∞]

julia> (-2.. -1) / (0..4)
[-∞, -0.25]
```
So maybe your code is wrong?

dpsanders: Or maybe it is just coming from using the invalid interval.

dpsanders: You should add a test with an interval that contains 0 divided by another interval that contains 0.

dpsanders:
This is looking good @eeshan9815, thanks for the contribution!
Please fix the remaining details, in particular the test, and it'll be ready.

test/interval_tests/numeric.jl
eeshan9815: Sorry, I accidentally put `Interval(-1.0, -2.0)` instead of `Interval(-2.0, -1.0)`. Fixing that should fix this test case. I will also replace `Interval` with `interval`.

lbenet:
@dpsanders I think this is ready to be merged. Shall I go ahead?

dpsanders:
Sure, thanks!

lbenet:
Thanks a lot @eeshan9815 !

Issue #104, opened 2018-03-12
Horrible performance for operations combining floats and intervals


dpsanders:
```jl
julia> x = 0.1..0.2
[0.0999999, 0.200001]

julia> @btime 3.1 + $x
  332.254 ns (3 allocations: 128 bytes)
[3.19999, 3.30001]
```
This is due to the fact that the promotion machinery calls `convert(Interval{Float64}, 3.1)`, which is now very slow again. 

Maybe we should just write explicit methods for operations of the form `op(x::Number, y::Interval)`.

dpsanders:
```jl
+(x::Real, y::Interval) = Interval(x) + y
```
etc.

lbenet:
While i think it is a good idea to avoid promotion, I don't see how will we avoid rounding, which is not considered above. An example would simply be `0.1 + Interval(0.0)`.

dpsanders:
I think we can just do `Interval(prevfloat(x), nextfloat(x))`.

dpsanders:
The problem seems to be in `find_quadrants` in
`temp = x / half_pi(x)`

dpsanders:
But `sin` is still very slow.

dpsanders:
There is a huge discrepancy due to finding the quadrants:

```
julia> x = 3..4

julia> @btime IntervalArithmetic.@round(sin($x.lo), sin($x.hi))
  44.328 ns (0 allocations: 0 bytes)

julia> @btime sin($x)
  700.951 ns (14 allocations: 448 bytes)
[-0.756803, 0.141121]
```

dpsanders:
```
julia> @btime Interval($y) / IntervalArithmetic.half_pi($y)
  36.213 ns (0 allocations: 0 bytes)
```
So the total time should be something like 80ns (which is still slow compared to other libraries :/ )

tkoolen:
Just my two cents (partly from https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/108): I'm probably missing something, but why not:

1. define
    ```julia
     convert(::Type{Interval{T}}, x::T) where {T<:Real} = Interval{T}(x)
     convert(::Type{Interval{T}}, x::T) where {T<:AbstractFloat} = Interval{T}(x)
     ```
    so that these conversions are fast and the resulting interval is as tight as possible (namely a trivial interval).
[...]

PR #114, opened 2018-04-06
Introduce `closure` function, extracted out of `convert`

tkoolen:
This implements https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/104#issuecomment-379126777.

* `closure` essentially does what `convert` did previously.
* `convert` will now avoid calling `closure` when doing so makes sense
* all explicit uses of `convert` in the package are replaced with `closure`

Result:

```julia
julia> x = 0.1..0.2
[0.0999999, 0.200001]

julia> @btime 3.1 + $x
  19.124 ns (0 allocations: 0 bytes)
[3.19999, 3.30001]
```

Fixes https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/104.

I'm not married to the name `closure`, but it kind of works.

~~Tests almost pass, but there's a couple of failures for `tan`; this is the first:~~

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/e6375b0bedd097e4efdc590b2ebdc52b9714811f/test/ITF1788_tests/libieeep1788_tests_elem.jl#L3629

~~`lo_quadrant` appears to be wrong. Any idea why that would be the case?~~

Never mind, already fixed it; all tests pass now.

### DIFF
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -13,16 +13,16 @@ promote_rule(::Type{BigFloat}, ::Type{Interval{T}}) where T<:Real =
 
 
 # convert methods:
-convert(::Type{Interval{T}}, x) where {T} = closure(Interval{T}, x)
+convert(::Type{Interval{T}}, x) where {T} = atomic(Interval{T}, x)
 convert(::Type{Interval{T}}, x::T) where {T} = Interval{T}(x)
 convert(::Type{Interval{T}}, x::Interval{T}) where {T} = x
-convert(::Type{Interval{T}}, x::Interval) where {T} = closure(Interval{T}, x)
+convert(::Type{Interval{T}}, x::Interval) where {T} = atomic(Interval{T}, x)
 
 convert(::Type{Interval}, x::Real) = (T = typeof(float(x)); convert(Interval{T}, x))
 convert(::Type{Interval}, x::Interval) = x
 
 """
-    closure(::Type{<:Interval}, x)
+    atomic(::Type{<:Interval}, x)
 
 Construct the tightest interval of a given type that contains the value `x`.
 
@@ -36,7 +36,7 @@ Construct an `Interval{Float64}` containing a given `BigFloat`:
 julia> x = big"0.1"
 1.000000000000000000000000000000000000000000000000000000000000000000000000000002e-01
 
-julia> i = IntervalArithmetic.closure(Interval{Float64}, x)
+julia> i = IntervalArithmetic.atomic(Interval{Float64}, x)
 [0.0999999, 0.100001]
 
 julia> i isa Interval{Float64}
@@ -52,10 +52,10 @@ true
 Construct an `Interval{Float32}` containing a the real number 0.1 in two ways:
 
 ```julia
-julia> i1 = IntervalArithmetic.closure(Interval{Float32}, "0.1")
+julia> i1 = IntervalArithmetic.atomic(Interval{Float32}, "0.1")
 [0.0999999, 0.100001]
 
-julia> i2 = IntervalArithmetic.closure(Interval{Float32}, 1//10)
+julia> i2 = IntervalArithmetic.atomic(Interval{Float32}, 1//10)
 [0.0999999, 0.100001]
 
 julia> i1 === i2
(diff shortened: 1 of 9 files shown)


src/intervals/conversion.jl
```diff
@@ -12,23 +12,32 @@ promote_rule(::Type{BigFloat}, ::Type{Interval{T}}) where T<:Real =
...
     isinf(x) && return wideinterval(T(x))
 
     Interval{T}( T(x, RoundDown), T(x, RoundUp) )
     # the rounding up could be done as nextfloat of the rounded down one?
     # use @round_up and @round_down here?
 end
 
-function convert(::Type{Interval{T}}, x::S) where {T<:AbstractFloat, S<:AbstractFloat}
+function closure(::Type{Interval{T}}, x::S) where {T<:AbstractFloat, S<:AbstractFloat}
     isinf(x) && return wideinterval(x)#Interval{T}(prevfloat(T(x)), nextfloat(T(x)))
```
lbenet: A small remark, which is not due to the changes you propose (perhaps it is me the one to be blamed!), but this is a good chance to correct it: This method is not type stable. To be type stable, I think it should return `wideinterval(T(x))`.

An example that shows the problem is:
```julia
julia> typeof( IntervalArithmetic.closure(Interval{BigFloat}, Inf) )
IntervalArithmetic.Interval{Float64}

julia> typeof( IntervalArithmetic.closure(Interval{BigFloat}, 0.1) )
IntervalArithmetic.Interval{BigFloat}
```

lbenet: As you prefer...

tkoolen:
Re: test coverage for `closure(::Type{Interval{Rational{Int}}}, x::Irrational)`
and `closure(::Type{Interval{Rational{BigInt}}}, x::Irrational)`: these weren't covered on master either and are in fact broken:

```julia
julia> convert(Interval{Rational{Int}}, π)
ERROR: MethodError: Cannot `convert` an object of type IntervalArithmetic.Interval{BigFloat} to an object of type IntervalArithmetic.Interval{Rational{Int64}}
This may have arisen from a call to the constructor IntervalArithmetic.Interval{Rational{Int64}}(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] convert(::Type{IntervalArithmetic.Interval{Rational{Int64}}}, ::Irrational{:π}) at /home/twan/code/RigidBodyDynamics/v0.6/IntervalArithmetic/src/intervals/conversion.jl:62
```

How should this be fixed?

lbenet:
The following is an idea, so feel free to change it as you want. 

`Irrational`s would yield a very narrow interval, but not a tight one. It may happen that rationalizing the bounds yields one distinct rational (instead of two), which cannot be used to build an interval that contains the irrational. So then, the idea is to shift to the left and right the rationalized value, the minimum possible amount in each case, and then construct the interval.

The following function does that in a very non-optimized way:
```julia
function clausure(::Type{Interval{Rational{T}}}, x::S) where {T<:Integer, S<:Irrational}
       x_rat = rationalize(T, x)
       p_rat = numerator(x_rat)
       q_rat = denominator(x_rat)
       lo = max( p_rat//(q_rat+1), (p_rat-1)//q_rat )
       hi = min( p_rat//(q_rat-1), (p_rat+1)//q_rat )
       return Interval(lo,hi)
end
```

And it seems to work:
```julia
julia> π ∈ clausure(Interval{Rational{Int}}, π)
true

julia> big(π) ∈ clausure(Interval{Rational{Int}}, π)
true

julia> eu ∈ clausure(Interval{Rational{Int}}, eu)
true

julia> big(eu) ∈ clausure(Interval{Rational{Int}}, eu)
true

julia> float( diam(clausure(Interval{Rational{Int}}, π)) )
2.373975753932138e-18

julia> float( diam(clausure(Interval{Rational{Int}}, eu)) )
9.472290677529826e-19

julia> diam(@interval(π))
4.440892098500626e-16
```

I hope this is useful!

dpsanders:
Thanks very much for the nice contribution, @tkoolen! 

I've now had a chance to discuss this with @lbenet and we are very happy to merge this as is, or would you like to add the methods to fix the bugs you found? (Or should we just add it to your PR?)

tkoolen:
Yeah, might be a good idea to merge now if you're sufficiently happy with it. It'd be another couple of days (maybe even a week) before I would have time to make and test the changes proposed by @lbenet.

dpsanders:
I think we will definitely merge then. 

My only doubt would be about the name closure. In true julia style, we should bikeshed it... 

I suggest enclosure (standard in the interval community) or atomic (as in atomic - indivisible).

lbenet:
I think I prefer `atomic`...

tkoolen:
`atomic` it is for now (further bikeshedding can be done in a later PR).

dpsanders:
Thanks again for this contribution, @tkoolen!

PR #116, opened 2018-04-09
Add docs on interval construction

dpsanders:
Detail the different ways of constructing intervals.


docs/src/construction.md
```diff
@@ -0,0 +1,84 @@
...
+
+Constructing an interval is the most basic operation in the library. There are several methods to construct intervals listed below.
+
+Note that a valid interval `[a, b]` must have `a ≤ b`.
+
+- `interval(x)`
+
+  `interval(x, y)`
+
+  This is the most fundamental way to build an interval for a user. It accepts one or two floating-point values and constructs the resulting interval with lower and upper endpoints exactly equal to those floating-point values, checking that the resulting interval is valid:
```
lbenet: I think it is worth emphasizing that no correct (directed) rounding is done by `interval`, and therefore `interval(0.1)` actually does not contain the real number 0.1.

PR #117, opened 2018-04-09
Speed up sin and cos by a factor 6.5

dpsanders:
Current master:
```jl
julia> x = 3..4
[3, 4]

julia> @btime sin($x)
  648.567 ns (14 allocations: 448 bytes)
[-0.756803, 0.141121]
```

This PR:
```jl
julia> @btime sin($x)
  94.037 ns (0 allocations: 0 bytes)
[-0.756803, 0.141121]
```

### DIFF
--- a/src/intervals/trigonometric.jl
+++ b/src/intervals/trigonometric.jl
@@ -1,9 +1,20 @@
 # This file is part of the IntervalArithmetic.jl package; MIT licensed
 
-half_pi(::Type{T}) where T = pi_interval(T) / 2
-half_pi(x::T) where T<:AbstractFloat = half_pi(T)
 
-two_pi(::Type{T})  where T = pi_interval(T) * 2
+# hard code this for efficiency:
+const one_over_half_pi_interval = inv(0.5 * pi_interval(Float64))
+
+"""
+Multiply an interval by a positive constant.
+For efficiency, does not check that the constant is positive.
+"""
+multiply_by_positive_constant(α, x::Interval) = @round(α*x.lo, α*x.hi)
+
+half_pi(::Type{Float64}) where {T} = multiply_by_positive_constant(0.5, pi_interval(T))
+half_pi(::Type{T}) where {T} = 0.5 * pi_interval(T)
+half_pi(x::T) where {T<:AbstractFloat} = half_pi(T)
+
+two_pi(::Type{T})  where {T} = multiply_by_positive_constant(2.0, pi_interval(T))
 
 range_atan2(::Type{T}) where {T<:Real} = Interval(-(pi_interval(T).hi), pi_interval(T).hi)
 half_range_atan2(::Type{T}) where {T} = (temp = half_pi(T); Interval(-(temp.hi), temp.hi) )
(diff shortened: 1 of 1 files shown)


dpsanders:
I have not found a sensible way to do tan correctly. 
We will be able to use #119 to calculate the quadrant on Julia 0.7.

Hence leaving tan aside for now.

lbenet:
Is this ready for review? Ping me when it is ready. (Regarding `tan`, we can leave it for another PR...)

dpsanders:
Yes, ready for review. 
tan will have to wait for Julia 0.7 / 1.0 (due any time now...)

src/intervals/trigonometric.jl
```diff
@@ -1,9 +1,21 @@
...
 
-two_pi(::Type{T})  where T = pi_interval(T) * 2
+# hard code this for efficiency:
+const one_over_half_pi_interval = inv(0.5 * pi_interval(Float64))
+
+"""
+Multiply an interval by a positive constant.
+For efficiency, does not check that the constant is positive.
+"""
+multiply_by_positive_constant(α, x::Interval) = @round(α*x.lo, α*x.hi)
```
lbenet: Perhaps instead of defining `multiply_by_positive_constant` we should rather define the methods `*(a::T, x::Interval)` and  `*(x::Interval, a::T)` (in `arithmetic.jl`), where `T<:Real` (or maybe `<:Number`), or even unspecified, as it is here. Currently we use promotion to an interval and rounding, which may slow down things.

Question: Below you use this function for to define `half_pi(::Type{Float64})`, but not for other types; can you explain why?

src/intervals/trigonometric.jl
```diff
@@ -19,9 +31,17 @@ This is a rather indirect way to determine if π/2 and 3π/2 are contained
 in the interval; cf. the formula for sine of an interval in
 Tucker, *Validated Numerics*."""
 
-function find_quadrants(x::T) where {T<:AbstractFloat}
+function find_quadrants(x::T) where {T}
     temp = atomic(Interval{T}, x) / half_pi(x)
-    (floor(temp.lo), floor(temp.hi))
+
+    return SVector(floor(temp.lo), floor(temp.hi))
```
lbenet: What about returning a tuple here, instead of a `SVector`? (Same comment applies in line 44.)

src/intervals/trigonometric.jl
dpsanders: I use it for `two_pi` as well.

Actually the `half_pi` function is no longer used, I believe.

We may well want to define explicit methods for `*(constant, Interval)`. I've opened an issue: https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/120

src/intervals/trigonometric.jl
dpsanders: It used to be a tuple; changing it to an `SVector` had a ridiculously big impact on the speed due to the `minimum` and `maximum` below that are not specialised for tuples!

In any case all of this code can be simplified a lot when #119 is possible, I believe.

src/intervals/trigonometric.jl
lbenet: That was actually the suggestion.

src/intervals/trigonometric.jl
lbenet: Fair enough.

src/intervals/trigonometric.jl
dpsanders: Right, but I'd prefer to leave it for a separate PR.
Also, if you actually know that the constant is positive then this specialised version is even slightly quicker.

src/intervals/trigonometric.jl
dpsanders: OK, I found an alternative solution: return a tuple and do

`max(tuple...)`

Too late...

Issue #113, opened 2018-04-06
StackOverflowError with erfc


SebastianM-C:
Code to reproduce:
```julia
using IntervalArithmetic
using SpecialFunctions
a=0..1
b=0..2
erfc(a/(2*sqrt(b)))
```
This leads to
```julia
ERROR: StackOverflowError:
Stacktrace:
 [1] erfc(::IntervalArithmetic.Interval{Float64}) at /mnt/storage/.julia/v0.6/SpecialFunctions/src/erf.jl:10 (repeats 80000 times)
```

lbenet:
Calling `erfc(a/(2*sqrt(b)))` executes [this line](https://github.com/JuliaMath/SpecialFunctions.jl/blob/032eb893cbae36674f0c16ddf7b17c5089883844/src/erf.jl#L10) in `SpecialFunctions.jl` (since the argument is an interval and Interval{T} <: Real). The problem occurs because `float(a::Interval)` returns an interval, which is used as the argument of `erfc`, again and again... So you get the StackOverflowError.

Since `erfc` is monotonously decreasing, it is not difficult to write a specialized method for intervals. Proper rounding to get a rigorous *tight* interval may be a headache, but the following may be good-enough for your:

```julia
julia> import SpecialFunctions: erfc

julia> erfc(a::Interval{T}) where {T} = interval(prevfloat(erfc(a.hi)), nextfloat(erfc(a.lo)))
erfc (generic function with 12 methods)

julia> erfc(a/(2*sqrt(b)))
[-4.94066e-324, 1.00001]

```

lbenet:
Was this helpful?

SebastianM-C:
Yes. Thank you!
Should I close the issue?

P.s. For compatibility with IntervalConstraintProgramming.jl is there a need for a special treatment for this function?

lbenet:
> P.s. For compatibility with IntervalConstraintProgramming.jl is there a need for a special treatment for this function?

I think there isn't; @dpsanders can you confirm this?

If that solved the issue, you may close it.

SebastianM-C:
I think that it would be helpful to add your answer on defining specialized implementations for intervals for monotonous functions somewhere in the docs. (If this is already documented, could you point me in the right direction?)

For IntervalConstraintProgramming.jl I tried to use a constraint containing `erfc`, but I got `ArgumentError: Function erfc not available. Use @function to define it.`

dpsanders:
[...]

PR #118, opened 2018-04-11
WIP: Add erf and erfc

SebastianM-C:
Implement `erfc` as discussed in #113
- In IntervalArithmetic.jl:
- [ ] Implement `erf(x::Interval)` and `erfc(x::Interval)` by using `erf(y::BigFloat)` and `erfc(y::BigFloat)` with monotonicity and directed rounding.
- [ ] Implement the inverse functions `erfinv` and `erfcinv` using interval Newton method.
- In IntervalContractors.jl:
- [ ] Follow the pattern that is defined there to add the reverse function `erf_rev`.


dpsanders:
Thanks for the contribution!

At the moment, you're missing `erf(x, r)` when `x` is a `Float64`.

dpsanders:
I was also discussing with @lbenet that since this introduces a dependency on SpecialFunctions.jl, we probably should split this out into a new IntervalSpecialFunctions.jl package.

SebastianM-C:
Should I continue with the PR here or it would be better to see what can be done regarding the split?

dpsanders:
Feel free to continue here, then we can just move the branch to a new repo when it's ready.

src/intervals/functions.jl
```diff
@@ -266,3 +266,22 @@ for f in (:log, :log2, :log10, :log1p)
...
+end
+
+function erf(a::Interval{T}) where T
+    isempty(a) && return a
+    @round( erf(a.lo), erf(a.hi) )
+end
+
+function erfc(a::Interval{T}) where T
+    isempty(a) && return a
+    @round( erf(a.hi), erfc(a.lo) )
```
dpsanders: `erfc(a.hi)`

dpsanders:
Will need some tests. May as well put them in a new test file.

src/intervals/functions.jl
```diff
@@ -266,3 +266,22 @@ for f in (:log, :log2, :log10, :log1p)
...
         end
 end
+
+for f in (:erf, :erfc)
+    @eval function($f)(x::BigFloat, r::RoundingMode)
+        setrounding(BigFloat, r) do
+            ($f)(x)
+        end
+    end
+
```
SebastianM-C: For the `Float64` case is `@eval ($f)(a::Interval{Float64}) = atomic(Interval{Float64}, $f(big53(a)))` ok? (I saw this for the `exp2` function, but it is not clear to me what `big53` is)

SebastianM-C:
Should I add the `SpecialFunctions` package to REQUIRE so that the CI can work?

src/intervals/functions.jl
dpsanders: `big53` is defined in the package. It converts the interval to a BigFloat interval with 53 bits of precision (i.e. the same as Interval{Float64}). You don't need `atomic`, just `convert`.

dpsanders:
Yes, go ahead and add `SpecialFunctions` to REQUIRE for CI.

src/intervals/functions.jl
```diff
@@ -266,3 +266,24 @@ for f in (:log, :log2, :log10, :log1p)
...
+    @eval ($f)(a::Interval{Float64}) = ($f)(big53(a))
```
dpsanders: You need to `convert` the result back to an `Interval{Float64}`.

SebastianM-C:
I think that the first part is almost complete. I will add some tests, but I am not sure how to build them (or what should be tested).

dpsanders:
You can just choose some random intervals and write literal tests like
```
@test erf(1..2) == Interval(0.8427007929497148, 0.9953222650189528)
```
or test that certain numbers are in the result of `erf(1..2)` etc.

dpsanders:
@SebastianM-C Once you add a couple more tests, I think we can merge this and do the rest in separate PRs.

SebastianM-C:
Should the checklist at the top be moved in #113 ?

Also: do you think that the tests are ok? I don't understand why the absolute coverage decreased.

dpsanders:
The tests look good, thanks!
Not sure about the coverage, wouldn't worry about it.

I would close that issue since it's now solved and move the checklist to a new issue.

dpsanders:
Thanks a lot!

dpsanders:
Now that I think about it, if we use the interval Newton method for the inverse functions, that will need IntervalRootFinding (or a custom implementation). In that case we should definitely split this out into a separate package.

dpsanders:
Wow, it really is slow:
```
x = 1..2
julia> @btime erf($x)
  61.184 μs (63 allocations: 2.41 KiB)
Interval(0.8427007929497148, 0.9953222650189528)

julia> @btime sin($x)
  119.898 ns (0 allocations: 0 bytes)
Interval(0.8414709848078965, 1.0)
```

SebastianM-C:
Glad I could help! Thanks for guiding me 😄

Issue #120, opened 2018-04-11
Define specialised methods for *(constant, Interval) and other arithmetic operators


dpsanders:
They seem to be about twice as fast as the current method involving `convert`-sion to an `Interval`.

lbenet:
Implemented in #125

PR #122, opened 2018-04-12
Add NEWS for v0.13

dpsanders:


NEWS.md
```diff
@@ -1,5 +1,32 @@
 # What's new in `IntervalArithmetic.jl`
 
+## v0.13
+
+### Breaking changes
+- `convert(Interval{T}, x::T)` no longer does any rounding: it creates
+the thin (zero-width) interval `[x, x]`. [#114](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/114)
+
+  Note that this implies that numbers are no longer rounded in operations combining them with intervals, such as `0.1 + interval(0.2, 0.3)`, since `0.1` is `convert`ed to an `Interval` by Julia's promotion machinery.
```
lbenet: Auch! I didn't realized this. It seems dangerous, isn't it?
How can this be circumvented?

NEWS.md
```diff
@@ -1,5 +1,32 @@
...
+
+### New functions
+- `atomic(Interval{T}, x)` creates the smallest atomic interval obtained by directed rounding, treating `x` as representing a real number (the old behaviour
```
lbenet: Since `atomic` is not exported, perhaps it is a good idea to indicate this, or write `IntervalArithmetics.atomic`...

NEWS.md
dpsanders:
```
julia> (1//10) + (1..2)
Interval(1.0999999999999999, 2.1)
```

dpsanders:
```
julia> @btime (1//10) + (1..2)
  149.455 ns (4 allocations: 64 bytes)

julia> @btime 0.1 + (1..2)
  25.432 ns (0 allocations: 0 bytes)
```

dpsanders: But 
```

julia> 0.1 * (1..2)
Interval(0.1, 0.2)

julia> big(ans)
Interval(1.000000000000000055511151231257827021181583404541015625000000000000000000000000e-01, 2.000000000000000111022302462515654042363166809082031250000000000000000000000000e-01)
```

lbenet: My concern is define a function `f(x) = 0.1 + x` and find out that 0.1 is not in `f(interval(0.0))`...
I guess we should extend solving #120 to other binary operations involving a `Real`.

tkoolen: > Basically it's treating 0.1 as the floating-point number 0.1 that it is, and not trying to be "clever" and guessing what the user wants.

These were my thoughts exactly.

PR #125, opened 2018-04-12
Add specialized arithmetic methods for Interval{T} and T

lbenet:
Fix to #120, including discussion found here: https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/122#pullrequestreview-111472074

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -65,6 +65,21 @@ typemax(::Type{Interval{T}}) where T<:Integer = Interval(typemax(T))
 +(a::Interval) = a
 -(a::Interval) = Interval(-a.hi, -a.lo)
 
+function +(a::Interval{T}, b::T) where {T<:Real}
+    isempty(a) && return emptyinterval(T)
+    @round(a.lo + b, a.hi + b)
+end
++(b::T, a::Interval{T}) where {T<:Real} = a+b
+
+function -(a::Interval{T}, b::T) where {T<:Real}
+    isempty(a) && return emptyinterval(T)
+    @round(a.lo - b, a.hi - b)
+end
+function -(b::T, a::Interval{T}) where {T<:Real}
+    isempty(a) && return emptyinterval(T)
+    @round(b - a.lo, b - a.hi)
+end
+
 function +(a::Interval{T}, b::Interval{T}) where T<:Real
     (isempty(a) || isempty(b)) && return emptyinterval(T)
     @round(a.lo + b.lo, a.hi + b.hi)
@@ -77,6 +92,17 @@ end
 
 
 ## Multiplication
+function *(x::T, a::Interval{T}) where {T<:Real}
+    isempty(a) && return emptyinterval(T)
+    (iszero(a) || iszero(x)) && return zero(Interval{T})
+
+    if x ≥ 0.0
+        return @round(a.lo*x, a.hi*x)
+    else
+        return @round(a.hi*x, a.lo*x)
+    end
+end
+*(a::Interval{T}, x::T) where {T<:Real} = x*a
 
 function *(a::Interval{T}, b::Interval{T}) where T<:Real
     (isempty(a) || isempty(b)) && return emptyinterval(T)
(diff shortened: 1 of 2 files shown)


dpsanders:
LGTM, thanks! How are the benchmarks?

lbenet:
Benchmarks seem ok, or actually improved:
In master:
```julia
julia> @btime 0.1 + interval(1)
  58.709 ns (4 allocations: 64 bytes)
Interval(1.0999999999999999, 1.1)
```
This PR:
```julia
julia> @btime 0.1 + interval(1)
  19.152 ns (0 allocations: 0 bytes)
Interval(1.0999999999999999, 1.1)
```

I haven't test yet other cases, because I just found the following:
```julia
julia> 0.1 + 1..1
ERROR: ArgumentError: `[1.0999999999999999, 1.0]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
```
This happens both in master and this PR. Any ideas?

dpsanders:
It's doing `(0.1 + 1)..1`. You need `0.1 + (1..1)`.

dpsanders:
Are you comparing with latest master?

lbenet:
> It's doing (0.1 + 1)..1. You need 0.1 + (1..1).

You are right... sorry.

Yes, I'm comparing with master.

dpsanders:
Also you are benchmarking the construction of the interval. 

```
x = 1..1
@btime 0.1 + $x
```

master:
```
16.100 ns (0 allocations: 0 bytes)
```

This PR:
```
9.555 ns (0 allocations: 0 bytes)
```

lbenet:
However, this is not solving the following problem, which is what I had in mind to get into this:
```julia
julia> 0.1 * interval(1)
Interval(0.1, 0.1)
```
I would have like to get a proper interval containing the *real* 0.1...

lbenet:
The following (obtained in this PR as in current master) also bothers me:
```julia
julia> @interval(0.3) + 0.1
Interval(0.39999999999999997, 0.4000000000000001)

julia> 0.3 + @interval(0.1)
Interval(0.39999999999999997, 0.4)

julia> @interval(0.3) + @interval(0.1)
Interval(0.39999999999999997, 0.4000000000000001)
```

lbenet:
Do you know if it is written something on the standard about this type of operations (float + interval); in the IT1788 there is no test like that.

dpsanders:
One solution to this would be to expand the number `x` to `Interval(prevfloat(x), nextfloat(x))`.
How would you feel about this?

But as @tkoolen emphasized, maybe this is all too magical: `0.1` is not `1//10`, and so we should not treat it as such?

lbenet:
In master, we get
```julia> 1//10 + interval(3//10)
ERROR: MethodError: no method matching atomic(::Type{IntervalArithmetic.Interval{Rational{Int64}}},
::IntervalArithmetic.Interval{Rational{Int64}})
...
```
This PR yields:
```julia
julia> 1//10 + interval(3//10)
Interval(2//5, 2//5)
```

lbenet:
> But as @tkoolen emphasized, maybe this is all too magical: 0.1 is not 1//10, and so we should not treat it as such?

I agree with both; when I write `f(x) = 0.1 + x`, 0.1 will be treated as a floating point, and not as the real which happens to be 1//10. Yet, I think the idea of all this is to return an interval that contains the true answer. Your suggestion using `prevfloat` and `nextfloat` is perhaps a way to go; the problem is that intervals always get wider, as we know...

dpsanders:
You can do that by writing
```
@interval 0.1 + x
```
Here you are telling the computer: "I know that there is some issue with 0.1. Ignore it and treat 0.1 as meaning the real number."

If you write instead `0.1 + x`, it is the sum of a particular floating point number with an interval, and we identify floating-point numbers with the corresponding thin interval.

dpsanders:
Indeed, `@interval` is the solution, since it does exactly the right thing:
```
julia> @macroexpand @floatinterval 0.1+x
:((IntervalArithmetic.Interval)((IntervalArithmetic.atomic)((IntervalArithmetic.Interval){Float64}, 0.1) + (IntervalArithmetic.atomic)((IntervalArithmetic.Interval){Float64}, x)))
```
!

lbenet:
So the take away is use `@interval` inside the function. That is, prefer the following function `g` over `f`:
```julia
julia> f(x) = 0.1 + x

julia> g(x) = @interval 0.1 + x

julia> f(interval(0.0))
Interval(0.1, 0.1)

julia> g(interval(0.0))
Interval(0.09999999999999999, 0.1)
```

Where whall we document this?

dpsanders:
Yes, that's the takeaway I would say!

I think we need to rewrite the Basic usage page, and we can document it there.

In any case I think we should merge this PR and tag the new release!

lbenet:
I agree.

lbenet:
Thanks a lot!

PR #131, opened 2018-04-16
Fix a - x and add tests

dpsanders:


lbenet:
Thanks a lot! I'm merging it right away!

Issue #140, opened 2018-04-26
Tests fail on Windows


Kolaru:
When I use Windows 10 with Julia version 0.6.2 (x86_64-w64-mingw32) and the master branch of the package, the tests fail with the following message:

```julia
julia> Pkg.test("IntervalArithmetic")
INFO: Computing test dependencies for IntervalArithmetic...
INFO: No packages to install, update or remove
INFO: Testing IntervalArithmetic
Constructing intervals: Test Failed
  Expression: @interval(1.0e300) == Interval(9.999999999999999e299, 1.0e300)
   Evaluated: Interval(1.0e300, 1.0e300) == Interval(9.999999999999999e299, 1.0e300)
Stacktrace:
 [1] macro expansion at C:\Users\Kolaru\.julia\v0.6\IntervalArithmetic\test\interval_tests\construction.jl:39 [inlined]
 [2] macro expansion at .\test.jl:860 [inlined]
 [3] anonymous at .\<missing>:?
Constructing intervals: Test Failed
  Expression: @interval(-1.0e307) == Interval(-1.0000000000000001e307, -1.0e307)
   Evaluated: Interval(-1.0e307, -1.0e307) == Interval(-1.0000000000000001e307, -1.0e307) Stacktrace:
 [1] macro expansion at C:\Users\Kolaru\.julia\v0.6\IntervalArithmetic\test\interval_tests\construction.jl:40 [inlined]
 [2] macro expansion at .\test.jl:860 [inlined]
 [3] anonymous at .\<missing>:?
Constructing intervals: Test Failed
  Expression: a == @floatinterval("0.1")
   Evaluated: Interval(0.09999999999999999, 0.1) == Interval(0.1, 0.1)
Stacktrace:
 [1] macro expansion at C:\Users\Kolaru\.julia\v0.6\IntervalArithmetic\test\interval_tests\construction.jl:103 [inlined]
 [2] macro expansion at .\test.jl:860 [inlined]
 [3] anonymous at .\<missing>:?
Test Summary:          | Pass  Fail  Total
Constructing intervals |   82     3     85
ERROR: LoadError: LoadError: LoadError: Some tests did not pass: 82 passed, 3 failed, 0 errored, 0 broken.
while loading C:\Users\Kolaru\.julia\v0.6\IntervalArithmetic\test\interval_tests\construction.jl, in expression starting on line 7
[...]

PR #143, opened 2018-04-30
Use BigFloat parsing for Float64 to fix Windows bug.

dpsanders:
Remove unnecessary `@eval`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6
-CRlibm 0.5
+CRlibm 0.6
 StaticArrays 0.5
 FastRounding 0.0.4
 AdjacentFloats 0.0.5
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -74,12 +74,22 @@ atomic(::Type{Interval{T}}, x::T) where {T<:Integer} = Interval{T}(x)
 atomic(::Type{Interval{T}}, x::AbstractString) where T<:AbstractFloat =
     parse(Interval{T}, x)
 
-function atomic(::Type{Interval{T}}, x::S) where {T<:AbstractFloat, S<:Real}
-    isinf(x) && return wideinterval(T(x))
+@static if is_windows()  # Windows cannot round properly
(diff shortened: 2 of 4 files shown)


dpsanders:
Fixes #140.

cc @kolaru

dpsanders:
Appveyor has found another annoying rounding issue...

dpsanders:
Probably the best way to fix this is to use 
`@static if is_windows()`
and make a version which uses BigFloat parsing as well.

lbenet:
Regarding the changes on `atomic`, have you seen #144?

Kolaru:
It looks like on Windowsthe rounding mode is never taken in account, so `exp(a, RoundDown)..exp(b, RoundUp)` is equivalent to `exp(a)..exp(b)`, which causes the problem.

It makes fail the `round_expr` function which relies on creating the `exp(a, RoundDown)..exp(b, RoundUp)` expression (which has then no effect on Windows). Since the `@round` uses `round_expr`, and all mathematical functions use `@round` this may affect any calculations on Windows. I do not see how the problem can be solved.

dpsanders:
Functions like `exp` are supposed to use CRlibm, but I don't know how to get that working on Windows (would require providing a binary download), so instead they wrap the BigFloat functions.

I get this on a Windows virtual machine (finally got one working), which seems to show that it's working correctly:

```jl
julia> exp(a, RoundDown)
2.718281828459045

julia> exp(a, RoundUp)
2.7182818284590455
```

dpsanders:
Ah, but the problem is here:
```jl
 julia> exp(709.7827128933841, RoundDown)
Inf
```

dpsanders:
I have got tests to pass locally on 64-bit Windows using the following PR:

https://github.com/JuliaIntervals/CRlibm.jl/pull/20

!!!

dpsanders:
Tests pass on Windows 64 on Appveyor. Windows 32 still has some weird stuff going on.

dpsanders:
Tests pass on Appveyor on Win32 and 64!

dpsanders:
Merging!

lbenet:
Thanks a lot!

PR #144, opened 2018-05-03
Add methods to use .. notation with Irrationals

### LINKED ISSUE #134: -π..π no longer works
```jl
julia> -π..π
ERROR: MethodError: no method matching atomic(::Type{IntervalArithmetic.Interval{Irrational{:π}}}, ::Irrational{:π})
```

lbenet:
Fix #134

### DIFF
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -126,6 +126,21 @@ function ..(a::T, b::S) where {T, S}
     interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
 end
 
+function ..(a::T, b::Irrational{S}) where {T, S}
+    R = promote_type(T, Irrational{S})
+    interval(atomic(Interval{R}, a).lo, atomic(Interval{R}, b).hi)
+end
+
+function ..(a::Irrational{T}, b::S) where {T, S}
+    R = promote_type(Irrational{T}, S)
+    interval(atomic(Interval{R}, a).lo, atomic(Interval{R}, b).hi)
+end
+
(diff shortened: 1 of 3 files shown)


lbenet:
Tests pass in linux and mac.

src/intervals/intervals.jl
```diff
@@ -126,6 +126,21 @@ function ..(a::T, b::S) where {T, S}
     interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
 end
 
+function ..(a::T, b::Irrational{S}) where {T, S}
+    R = promote_type(T, Irrational{S})
+    interval(atomic(Interval{R}, a).lo, atomic(Interval{R}, b).hi)
```
dpsanders: I think we should add a method
`atomic(T, a) = atomic(Interval{T}, a)`.

Also, since this line occurs in each of the three versions, could it be refactored out somehow?

lbenet: The reason to have it like this is because all `atomic` methods are defined with small variations of `atomic(::Type{Interval{T}}, x::S)`; see #114. I can certainly introduce the method you suggest, but maybe it will end up to be too generic.

lbenet: I just checked your suggestion and tests pass (locally) in Julia 0.6.3; if you'd like I add a commit with the suggestion.

dpsanders: Yes please go ahead! Maybe the type T should be restricted to <: AbstractFloat for example.

dpsanders:
Thanks!

lbenet:
Thanks for merging... and sorry I didn't do it before.

Issue #129, opened 2018-04-16
What to do with mod?


dpsanders:
How should we deal with functions like `mod` acting on intervals?

E.g. 
- `mod(0.8..1.2, 1)` should give `(0..0.2) ∪ (0.8..1)`;
`mod(0.8..2.2, 1)` should give `0..1`?

One solution is to define an interval union type; see e.g.
https://link.springer.com/article/10.1007/s10543-016-0632-y

Really the correct solution is probably to use contractors.

lbenet:
Speaking about `Interval`, I think it should behave as `tan`. For the specific example you give, I think it should return `0..1`, which corresponds to the `hull`.

A complementary possibility is to have `mod(a::Interval)` returning the `hull`, and also implement `extended_mod(a::Interval)`, among other examples (e.g. `extended_inv(a::Interval)`). 

`extended_mod(a)` would return a `Vector{Interval)` (or perhaps `NTuple{N,Interval}`) containing the list of (connected) intervals, as a way of representing the union in the sense you use above. If the interval `a` does not contain the discontinuity, the result would correspond to `[mod(a)]`.

Perhaps, this is what you mean by using contractors.

lbenet:
In #145 I have implemented something that is equivalent to `mod(0.8..2.2, 1)` --> `0..1`.

We could also implement some `extended_mod` which would return a tuple of intervals (two or three, this we should discuss) which somehow mimics `extended_div`.

dpsanders:
Great, thanks.
Why would there be three intervals?

Is this version of mod (which just returns 0..1) useful? I tried it with
root finding and it didn't seem to be too useful.

lbenet:
> Why would there be three intervals? 

Considering your example, `mod(0.8..2.2, 1)`, I would say that `extended_mod` would return `(0.8..1.0) ∪ (0.0..1.0) ∪(0.0..0.2)`, if we consider each segment.

> Is this version of mod (which just returns 0..1) useful?

This may be useful for things like `find_quadrants`; for root-finding methods, we would need `extended_mod`. I'll push a new commit dealing with `extended_mod` later.

lbenet:
Just pushed another commit to #145 with `extended_mod` implemented. Can you retest your example with root finding, using now `extended_mod`? Maybe you could post it so I can play with it.

dpsanders:
The problem is that all of the root finding functions assume that a single
[...]

PR #145, opened 2018-05-03
Add mod

lbenet:
This PR addresses #129.

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -62,7 +62,7 @@ export
     RoundTiesToEven, RoundTiesToAway,
     cancelminus, cancelplus, isunbounded,
     .., @I_str, ±,
-    pow, extended_div,
+    pow, extended_div, extended_mod,
     setformat, @format
 
 export
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -27,6 +27,7 @@ import Base:
     in, zero, one, eps, typemin, typemax, abs, abs2, real, min, max,
     sqrt, exp, log, sin, cos, tan, inv,
     exp2, exp10, log2, log10,
+    mod,
     asin, acos, atan, atan2,
     sinh, cosh, tanh, asinh, acosh, atanh,
     union, intersect, isempty,
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -286,3 +286,33 @@ function mod(a::Interval, y::T) where {T<:Real}
         return interval(mod(a.lo, y), mod(a.hi, y))
     end
 end
+
+
+function extended_mod(a::Interval, y::T) where {T<:Real}
+    yy = abs(y)
+    fld_lo = fld(a.lo, yy)
+    fld_hi = fld(a.hi, yy)
+    z = zero(fld_lo)
+    S = typeof( z )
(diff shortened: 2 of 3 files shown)


dpsanders:
Does this work with e.g. `mod(X, pi_interval(Float64))`?
Maybe we could add some tests for cases like this.

lbenet:
The problem seems to be that `mod(x::Interval, y::Real)` assumes `y` different from an `Interval`. When `y` is an interval, `fld` (used internally) makes a promotion which ends calling (recursively) `mod` which yields the segmentation fault.

Any idea about how to restrict `y` to avoid being an Interval?

dpsanders:
You can do
```
mod(x::Interval, y::AbstractFloat)
```
Or use `/` instead of `fld`?

lbenet:
Using `/` instead of `fld` would not guarantee rounding it down. In any case, I'll take your suggestion into account.

dpsanders:
Maybe the following?

```jl
julia> x = 7.3..7.3
[7.29999, 7.30001]

julia> y = 5.5..5.5
[5.5, 5.5]

julia> floor(x / y)
[1, 1]
```

dpsanders:
```
julia> y = pi..pi
Interval(3.141592653589793, 3.1415926535897936)

julia> x / y
Interval(2.3236621691416715, 2.3236621691416723)

julia> floor(x / y)
Interval(2.0, 2.0)
```

lbenet:
I'll certainly test it, since it is simply faster:
```julia
julia> @btime floor(7.3/5.5)
  1.895 ns (0 allocations: 0 bytes)
1.0

julia> @btime fld(7.3, 5.5)
  10.660 ns (0 allocations: 0 bytes)
1.0
```

lbenet:
Let me rebase to current master and check if everything runs smoothly.

lbenet:
Rebased to current master and implemented [this suggestion](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/145#issuecomment-395522571).

lbenet:
Locally, tests pass.

dpsanders (commented):
Lgtm, thanks! Could you add some tests with mod respect to pi, please?

lbenet:
Thanks for the suggestion. I'll try to add more tests.

dpsanders:
But any `mod` with `pi` *should* use the interval version of `pi`, no?
We certainly want to allow `mod` with (narrow -- probably atomic?) intervals, for this reason, and the answer should be an interval.

dpsanders:
Isn't this basically the same as the issue with quadrants for `sin` and `cos`? And the solution should be similar, I guess.

dpsanders:
I think it should be something like
```
julia> X = 10..11
[10, 11]

julia> p = pi_interval(Float64)
[3.14159, 3.1416]

julia> division = X / p
[3.18309, 3.50141]

julia> division_floor = floor(division)
[3, 3]

julia> result = division - division_floor
[0.183098, 0.501409]
```
But sometimes we will have:
julia> X = 2p
[6.28318, 6.28319]

julia> X / p
[1.99999, 2.00001]
```
If the `lo` and `hi` of the floor are different then the interval spans two quadrants, so the `mod` should be `0..(p.hi)`. 

For `extended_mod`, we can then do whatever the correct thing is.

PR #147, opened 2018-05-08
Fix mid for large floats

Kolaru:
- Allows to shift the middle point with `mid(-Inf..Inf, α)` with `α != 0.5` (`α` was previously ignored for infinite intervals).
- Change the expression which evaluate the midpoint to avoid that it returns `Inf` or `-Inf`.

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -374,33 +374,31 @@ the interval `a`. The default is the true midpoint at α=0.5.
 
 Assumes 0 ≤ α ≤ 1.
 
-Unbounded intervals are special cases which respect, for `x` arbitrary
-    - `mid(-∞..x, α) == nextfloat(-∞)` (`α` arbitrary)
-    - `mid(x..+∞, α) == prevfloat(+∞)` (`α` arbitrary)
-    - `mid(-∞..+∞, 0.5) == 0`
-    - `mid(-∞..+∞, α) == nextfloat(-∞)` (`α < 0.5`)
-    - `mid(-∞..+∞, α) == prevfloat(+∞)` (`α > 0.5`)
+The infinite values `-∞` and `+∞` are replaced by respectively `nextfloat(-∞)`
+and `prevfloat(+∞)` for the sake of finding the shifted midpoint of the interval.
 """
 function mid(a::Interval{T}, α) where T
 
     isempty(a) && return convert(T, NaN)
 
-    if isentire(a)
-        α == 0.5 && return zero(a.lo)
-        # Shift center if α != 0.5
-        α < 0.5 && return nextfloat(-∞)
-        α > 0.5 && return prevfloat(+∞)
-    end
+    lo = (a.lo == -∞ ? nextfloat(-∞) : a.lo)
+    hi = (a.hi == +∞ ? prevfloat(+∞) : a.hi)
 
-    a.lo == -∞ && return nextfloat(-∞)
-    a.hi == +∞ && return prevfloat(+∞)
-
-    midpoint = α * (a.hi - a.lo) + a.lo
+    midpoint = α * (hi - lo) + lo
     isfinite(midpoint) && return midpoint
-    # Fallback in case of overflow: a.hi - a.lo == +∞
-    return (1-α) * a.lo + α * a.hi
+    # Fallback in case of overflow: hi - lo == +∞
+    return (1-α) * lo + α * hi
 end
 
+"""
+    mid(a::Interval)
+
+Find the midpoint of interval `a`.
+
+For intervals of the form `[-∞, x]` or `[x, +∞]` where `x` is finite, return
+respectively `nextfloat(-∞)` and `prevfloat(+∞)`. Note that it differs from the
+behavior of `mid(a, α=0.5)`.
+"""
 function mid(a::Interval{T}) where T
 
     isempty(a) && return convert(T, NaN)
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -376,15 +376,21 @@ Assumes 0 ≤ α ≤ 1.
 function mid(a::Interval{T}, α) where T
 
     isempty(a) && return convert(T, NaN)
-    isentire(a) && return zero(a.lo)
+
+    if isentire(a)
+        α == 0.5 && return zero(a.lo)
+        # Shift center if α != 0.5
+        lo = nextfloat(-∞)
+        hi = prevfloat(+∞)
+        return (1-α) * lo + α * hi
+    end
 
     a.lo == -∞ && return nextfloat(-∞)
     a.hi == +∞ && return prevfloat(+∞)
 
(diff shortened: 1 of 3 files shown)


src/intervals/arithmetic.jl
```diff
@@ -375,15 +375,21 @@ Assumes 0 ≤ α ≤ 1.
...
 
     isempty(a) && return convert(T, NaN)
-    isentire(a) && return zero(a.lo)
+
+    if isentire(a)
+        α == 0.5 && return zero(a.lo)
+        # Shift center if α != 0.5
+        lo = nextfloat(-∞)
+        hi = prevfloat(+∞)
+        return (1-α) * lo + α * hi
```
dpsanders: I wonder if this should be something like a geometric mean?

src/intervals/arithmetic.jl
```diff
@@ -375,15 +375,21 @@ Assumes 0 ≤ α ≤ 1.
...
+    end
 
     a.lo == -∞ && return nextfloat(-∞)
```
dpsanders: This (semi-infinite intervals) should probably be changed in a similar way.

dpsanders:
Great idea, thanks!

dpsanders:
Could you please rebase this against current master?

src/intervals/arithmetic.jl
Kolaru: The use case I have in mind (shifted midpoint for `IntervalRootFinding`) only requires something like

```julia
mid(-Inf..Inf, 0.4) < 0
```

so the average used does not really matter. Clearly using a geometric mean makes more sense, but I thing that expecting anything really meaningful when calling `mid(-Inf..Inf, α)` would be very optimistic from the user.

Also using different types of average for different inputs may be confusing.

Kolaru:
There is a test specifically for

```julia
mid(-Inf..Inf, 0.75) == 0
```

is there a particular reason for that ? Or can I change it to match the new result ?

dpsanders:
Feel free to change it, thanks.

Kolaru:
I had to separate two cases in `mid to comply with the ITF 1788 test suite.

Also, since the standard forces `mid(-Inf..x) == nextfloat(-Inf)`, it is not possible for `α` to be taken in account meaningfully and have `mid(-Inf..x) == mid(-Inf..x, 0.5)`. Therefore I uniformized everything by making`mid(a, α)` more or less ignore `α` if `a` is unbounded, simply going to `nextfloat(-Inf)` or `prevfloat(+Inf)` depending on the value of `α` for `a = -Inf..Inf`.

dpsanders:
> and have mid(-Inf..x) == mid(-Inf..x, 0.5)

Do we really need this? I think it was nicer that you could use `alpha` to change the behaviour.

Kolaru:
It would a really unique and strange feature to have a function with an optional parameter where calling the function with the parameter set to the default value gives a result different from calling the function without the parameter specified.

I would be in favor of either defining another function like `extended_mid` (which could implement logarithmic-like definition of the midpoint) or adding a keyword parameter to opt-in for the scaling of infinite intervals (or logarithmic-like midpoint), something like

```julia
mid(X, alpha=0.5 ; scale_infinite=false, log_scale=false)
```

Kolaru:
> Well these would really be two different methods of the same function. One of them (without the additional parameter) gives the behaviour mandated by the standard; the other gives behaviour that is different and possibly more useful in some contexts. I don't see any problem with that, as long as it's documented.

My point was more about getting a easier to use  documentation, since even if two methods are in principle totally distinct some relations may be expected by the user.

But anyway, I may be overthinking it (which will be confirmed if nobody else complains about it) and fundamentally it doesn't really bother me. So let's go with it.

src/intervals/arithmetic.jl
```diff
@@ -368,24 +368,36 @@ end
...
+    hi = (a.hi == +∞ ? prevfloat(+∞) : a.hi)
 
-    # @assert 0 ≤ α ≤ 1
-
-    # return (1-α) * a.lo + α * a.hi  # rounds to nearest
-    return α*(a.hi - a.lo) + a.lo  # rounds to nearest
+    midpoint = α * (hi - lo) + lo
+    isfinite(midpoint) && return midpoint
+    # Fallback in case of overflow: hi - lo == +∞
+    return (1-α) * lo + α * hi
```
dpsanders: Doesn't this line work for all cases?

src/intervals/arithmetic.jl
```diff
@@ -394,9 +406,10 @@ function mid(a::Interval{T}) where T
     a.lo == -∞ && return nextfloat(a.lo)
     a.hi == +∞ && return prevfloat(a.hi)
 
-    # @assert 0 ≤ α ≤ 1
-
-    return 0.5 * (a.lo + a.hi)
+    midpoint = 0.5 * (a.lo + a.hi)
+    isfinite(midpoint) && return midpoint
+    # Fallback in case of overflow: a.hi + a.lo == +∞ or a.hi + a.lo == -∞
+    return 0.5 * a.lo + 0.5 * a.hi
```
dpsanders: Doesn't this line work for all cases? Is this for efficiency?

Kolaru: Yes it works for all cases, it is simply a bit less efficient (on floating point operation more). I simply tried to preserve the previous implementation.

However it looks like the added structure add a prohibitive cost so I should only keep the last line

```julia
julia> function f(a::Interval)
           midpoint = 0.5 * (a.lo + a.hi)
           isfinite(midpoint) && return midpoint
       end

julia> g(a::Interval) = 0.5 * a.lo + 0.5 * a.hi

julia> @btime f.(rr) ;
  248.527 μs (20032 allocations: 391.97 KiB)

julia> @btime g.(rr) ;
  20.160 μs (26 allocations: 79.23 KiB)
```

(should have run this tests before committing)

Kolaru: **Erratum:** Actually some tests (in `minimal_mid_tests` in the `ITF1788_tests\libieeep1788_tests_num.jl` test suite) fail with the `0.5 * a.lo + 0.5 * a.hi` line. The reason for that is the fact that
```julia
0.5 * nextfloat(0.0) == 0.5 * prevfloat(0) == 0
```

Therefore (with the longer version) we would have
```julia
julia> X = (2*prevfloat(0.0))..nextfloat(0.0)
[-1.4822e-323, 9.88132e-324]

julia> mid(X)
-5.0e-324
```

but the value `0.0` is expected (this is phrased in term of hexadecimal number in the test suite but I find that version more readable). So I think I did the right choice (I should definitively document it in the code though).

dpsanders: OK great, thanks.

dpsanders:
Maybe this needs rebasing against current master to deal with the failing tests?

Kolaru:
The previous tests for bisect tested behaviors specific of the old `mid`, the new commit lower their expectation to better reflect the expectations for the cases of infinite intervals.

dpsanders:
Thanks!

Issue #149, opened 2018-05-15
Returned type of arithmetic operations and functions for `IntervalBox`es


lbenet:
I just noticed the following behavior:

```julia
julia> using IntervalArithmetic

julia> b = (0..1) × (0..1)
IntervalBox(Interval(0.0, 1.0), Interval(0.0, 1.0))

julia> b+b
2-element StaticArrays.SArray{Tuple{2},IntervalArithmetic.Interval{Float64},1,2}:
 Interval(0.0, 2.0)
 Interval(0.0, 2.0)

julia> typeof(ans)
StaticArrays.SArray{Tuple{2},IntervalArithmetic.Interval{Float64},1,2}
```

Similar behavior is obtained for other arithmetic operations and functions:
```julia
julia> 2*b
2-element StaticArrays.SArray{Tuple{2},IntervalArithmetic.Interval{Float64},1,2}:
 Interval(0.0, 2.0)
 Interval(0.0, 2.0)

julia> exp(b)
2-element StaticArrays.SArray{Tuple{2},IntervalArithmetic.Interval{Float64},1,2}:
 Interval(1.0, 2.7182818284590455)
 Interval(1.0, 2.7182818284590455)
```

In order to use interval boxes in functions or types which require interval boxes, the returned type of such arithmetic operations should be an `IntervalBox`.

lbenet:
Is there a reason *not* to have `IntervalBox`es be the returned type in the above cases? 

Tests actually check things like `A + B == IntervalBox( A .+ B)`, which `A` and `B` `IntervalBox`es. My suggestion is to that `A + B` returns directly `IntervalBox( A .+ B)`.

dpsanders:
Yes I think these opérations should return IntervalBoxes as you suggest.

lbenet:
I just noticed i actually don't need this in TaylorModels, but still it would be nice to have it like that. I'll send a PR tomorrow; I haven't completed the tests...

Issue #150, opened 2018-05-16
Rewrite IntervalBox as a wrapper around SVector


dpsanders:
Instead of a subtype of `StaticVector`.

This is because we do not respect the semantics of an `AbstractArray`. See https://github.com/JuliaArrays/StaticArrays.jl/issues/406

PR #151, opened 2018-05-16
Return IntervalBoxes for operations and functions on IntervalBoxes

lbenet:
Addresses #149

### DIFF
--- a/src/multidim/arithmetic_functions.jl
+++ b/src/multidim/arithmetic_functions.jl
@@ -1,0 +1,29 @@
+# This file is part of the IntervalArithmetic.jl package; MIT licensed
+
++(a::IntervalBox, b::IntervalBox) = IntervalBox( a .+ b )
++(a::IntervalBox, b::Real) = IntervalBox( a .+ b )
++(a::Real, b::IntervalBox) = IntervalBox( a .+ b )
+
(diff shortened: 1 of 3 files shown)


lbenet:
Tests passed locally; but no idea how they may behave in windows.

dpsanders:
I'm not sure we want the following:
```
julia> X = IntervalBox(3..3, 4..4)
[3, 3] × [4, 4]

julia> sin(X)
[0.14112, 0.141121] × [-0.756803, -0.756802]
```
This should really be `sin.(X)`, which still returns
```
julia> sin.(X)
2-element StaticArrays.SArray{Tuple{2},IntervalArithmetic.Interval{Float64},1,2}:
  [0.14112, 0.141121]
 [-0.756803, -0.756802]
```
This is equivalent to 
```
broadcast(sin, X)
```
so it's probably that that we should be changing.

dpsanders:
Not sure how to do that correctly, though.
See also the comment in https://github.com/JuliaArrays/StaticArrays.jl/issues/406 about `similar_type`.

dpsanders:
I think this will actually be easier once we do #150 .

lbenet:
Closing in favor of #151

PR #152, opened 2018-05-17
Rewrite IntervalBox to contain an SVector of Intervals

### LINKED ISSUE #150: Rewrite IntervalBox as a wrapper around SVector

dpsanders:
Fixes #150.

Rewrites `IntervalBox` to contain an `SVector`. 
(Previously `IntervalBox` was a subtype of `StaticVector`.
Incorporates #151.

cc @lbenet


src/multidim/arithmetic.jl
```diff
@@ -0,0 +1,18 @@
...
+-(a::Real, b::IntervalBox) = IntervalBox( a .- b.v )
+-(a::IntervalBox) = IntervalBox( .- a.v )
+
+*(a::IntervalBox, b::Real) = IntervalBox( a.v .* b )
+*(a::Real, b::IntervalBox) = IntervalBox( a .* b.v )
+
+/(a::IntervalBox, b::Real) = IntervalBox( a.v ./ b )
+
+Base.broadcast(f, X::IntervalBox) = IntervalBox(f.(X.v))
+Base.broadcast(f, X::IntervalBox, Y::IntervalBox) = IntervalBox( f.(X.v, Y.v) )
```
lbenet: Neat!

src/multidim/intervalbox.jl
```diff
@@ -3,53 +3,56 @@
...
-# StaticArrays.Size{N,T}(::Type{IntervalBox{N,T}}) = StaticArrays.Size(N) # @pure not needed, I think...
-Base.@propagate_inbounds Base.getindex(a::IntervalBox, i::Int) = a.data[i]
+# IntervalBox(x::Interval) = IntervalBox( SVector(x) )  # single interval treated as tuple with one element
 
+IntervalBox(x::Interval...) = IntervalBox(SVector(x))
+IntervalBox(x::Tuple{T}) where {T<:Interval} = IntervalBox(SVector(x))
 
-IntervalBox(x::Interval) = IntervalBox( (x,) )  # single interval treated as tuple with one element
-
+Base.getindex(X::IntervalBox, i) = X.v[i]
```
lbenet: This replaces what it was in line 11, before; do we need `Base.@propagate_inbounds` here, or not?

dpsanders: Good question. To be honest I'm not sure, but it probably doesn't hurt.

lbenet: I'm not sure either... I've never understood why it was needed. Proceed as you prefer.

dpsanders:
Just added it.

dpsanders:
The tests from #151 are included here I believe?

src/multidim/arithmetic.jl
lbenet: While I truly think that this broadcasting change is very nice, I just noticed that, in order to use a function (say `exp`) on an `IntervalBox`, you *do* require the dot-notation. The following uses the current commit (93bb8e0f4) of this PR:
```julia
julia> A = IntervalBox(1..2, 3..4)
[1, 2] × [3, 4]

julia> exp.(A) # this is implemented here
[2.71828, 7.38906] × [20.0855, 54.5982]

julia> exp(A) # throwing an error here is uncomfortable!
ERROR: MethodError: no method matching exp(::IntervalArithmetic.IntervalBox{2,Float64})
Closest candidates are:
  exp(::Float16) at math.jl:950
  exp(::Complex{Float16}) at math.jl:951
  exp(::BigFloat) at mpfr.jl:517
  ...
```

This is sort of uncomfortable, because you force the user to write whatever functions using the dot notation instead of the usual functions. This is actually not a problem, a priori, but then it should be explicitly written somewhere. My idea with #151 was to avoid this, i.e., use directly `exp(A)`.

Should we allow `exp(A)` to simply return `exp.(A)`, or shall we leave it as it is now?

lbenet:
> The tests from #151 are included here I believe?

Yes, but using broadcasting, i.e., requiring using the dot notation. See [this comment](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/152/files#r189391089).

lbenet:
The kind of subtleties seems important in what I'm doing in TaylorModels. There, from time to time I need to evaluate expressions of the form `f( x0 + v )`, where `x0` is an `IntervalBox`, and `v` is an array with `TaylorN`s (actually, the independent variables). Using this branch, I get the following errors:
```julia
julia> x0
IntervalBox(Interval(0.0, 0.0), Interval(1.0, 1.0))

julia> v
2-element Array{TaylorSeries.TaylorN{IntervalArithmetic.Interval{Float64}},1}:
  Interval(1.0, 1.0) x + 𝒪(‖x‖³)
  Interval(1.0, 1.0) y + 𝒪(‖x‖³)

julia> x0 + v
ERROR: MethodError: no method matching +(::IntervalArithmetic.IntervalBox{2,Float64}, ::Array{TaylorSeries.TaylorN{IntervalArithmetic.Interval{Float64}},1})
...

julia> x0 .+ v
ERROR: MethodError: no method matching +(::IntervalArithmetic.IntervalBox{2,Float64}, ::TaylorSeries.TaylorN{IntervalArithmetic.Interval{Float64}})
...
```

Can we add methods that at least one of these works?

src/multidim/arithmetic.jl
dpsanders: Julia has moved to requiring `exp.(v)` to make explicit that you want to apply `exp` element-wise, since `exp` of a vector does not make sense.

I don't understand the context where `exp` of an `IntervalBox` makes sense. I don't actually understand your `Taylor` example.

dpsanders:
I think we would want `x0 .+ v` to work, and even `x0 + v`. I also came across similar issues.
But when would `exp(x0)` be needed?

src/multidim/arithmetic.jl
lbenet: > Julia has moved to requiring exp.(v) to make explicit that you want to apply exp element-wise, since exp of a vector does not make sense.

Fair enough.

lbenet:
I'm not sure if we actually need to have `f(::IntervalBox)`, so let's leave that for the time being.

In any case, it would be nice to have, at least, `f( x0 .+ v )` working, with `x0::IntervalBox` and `v::AbstractArray`. It certainly can be done manually, but I think it is worth having it as a function by itself.

dpsanders:
I still do not understand broadcasting properly, so for now the easiest thing will be to define + for an IntervalBox and a Taylor.

Kolaru:
I think that defining

```julia
Base.broadcast(X::IntervalBox, v::AbstractVector) = IntervalBox( f.(X, v)... )
```

does the trick for all number-interval operations, since it is more specific than something like

```julia
Base.broadcast(A::Any, v::AbstractVector)
```

which is probably currently called when using this PR with `X .+ v`, `v` being a vector.

lbenet:
Good point. Certainly it is worth checking.

dpsanders:
Currently things like
`mid.(X)` are failing, i.e. things which are *not* supposed to return an `IntervalBox`.

dpsanders:
The solution seems to be the following: explicitly say how to broadcast those functions:
```
Base.broadcast(::typeof(mid), X::IntervalBox) = mid.(X.v)
```

dpsanders:
Other functions:
`mig`, `mag`, `radius`

Also variants of `mid` with an extra argument.

Kolaru:
Another solution is to define a function which turn vector of `Interval` into `IntervalBox` but leaves others unchanged.

```julia
choose_type(v::SVector{Interval}) = IntervalBox(v)
choose_type(v) = v
```

Then broadcasting becomes

```julia
Base.broadcast(f, X::IntervalBox) = choose_type(f.(X.v))
```
and it should work as expected.

dpsanders:
@Kolaru That sounds like a great solution, thanks!

dpsanders:
The failure on 32-bit Windows seems to be a Julia bug: https://github.com/JuliaLang/julia/issues/27402

So I think this is ready to merge.

lbenet:
Merging! Thanks a lot!

PR #154, opened 2018-05-17
Add some set operations

eeshan9815:


lbenet:
Thanks for the contribution! 

I think the definition of `⊂` corresponds to `⪽` (`isinterior`); can you confirm? (If I recall correctly, the reason we used that name for the function and the symbol is because they are used like that in the standard.) In any case, the "other-way-around" operators are certainly a good addition.

dpsanders:
Thanks!

PR #157, opened 2018-06-03
Fix set minus in docs

dpsanders:
Use Unicode `\setminus` so that the package is loadable on Julia 0.7-alpha.


dpsanders:
The tests on Julia 0.7 don't pass, but we should still merge this right away since it at least allows doing `using IntervalArithmetic` on 0.7-alpha! cc @lbenet

lbenet:
Fine with me. Merging. Thanks!

PR #159, opened 2018-06-07
Add another broadcast method for IntervalBox

dpsanders:


lbenet:
LGTM; and travis seems to agree. Can you add some tests?

lbenet:
I'll merge this, and then #160.

PR #160, opened 2018-06-19
Add methods for IntervalBox previously in IntervalRootFinding

### LINKED ISSUE #158: Add methods for IntervalBox that are used in IntervalRootFinding

lbenet:
Fixes #158


lbenet:
See also https://github.com/JuliaIntervals/IntervalRootFinding.jl/issues/78, which requires a new patch version.

lbenet:
@dpsanders Question: Currently, `mid(X)` is equivalent to `mid(X, 0.5)`; on the other hand, `bisect(X)` corresponds to `bisect(X, 127/256)`, which was the way it was defined in `IntervalRootFinding`. Shall I leave it like that, or have the default value equal to 0.5, and change things in `IntervalRootFinding`?

Incidentally, tests are passing (except maybe for appveyor).

dpsanders:
The default for `mid` corresponds to that of the IEEE standard, so I guess we should leave it.
The default for `bisect` is more useful as being not equal to 0.5, since that avoids problems with zeros at 0 when using symmetric intervals, and generally seems more useful.

src/multidim/intervalbox.jl
```diff
@@ -45,7 +45,7 @@ Return a vector of the `mid` of each interval composing the `IntervalBox`.
 See `mid(X::Interval, α=0.5)` for more informations.
 """
 mid(X::IntervalBox) = mid.(X)
-mid(X::IntervalBox, α) = mid.(X, α)
+mid(X::IntervalBox, α) = mid.(X[:], α)
```
dpsanders: This can be left as `mid.(X, alpha)` after incorporating the change in #159

src/multidim/intervalbox.jl
```diff
@@ -73,6 +73,12 @@ diam(X::IntervalBox) = maximum(diam.(X.v))
 
 emptyinterval(X::IntervalBox{N,T}) where {N,T} = IntervalBox(emptyinterval.(X.v))
 
+isinf(X::IntervalBox) = any(isinf.(X))
```
dpsanders: This can be written `any(isinf, X)`, which I believe should be slightly faster.

src/multidim/intervalbox.jl
```diff
@@ -73,6 +73,12 @@ diam(X::IntervalBox) = maximum(diam.(X.v))
...
+
+isinterior(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(isinterior.(X, Y))
+
+contains_zero(X::SVector) = all(contains_zero.(X))
```
dpsanders: `all(contains_zero, X)`

src/multidim/intervalbox.jl
```diff
@@ -73,6 +73,12 @@ diam(X::IntervalBox) = maximum(diam.(X.v))
...
+contains_zero(X::IntervalBox) = all(contains_zero.(X))
```
dpsanders: Same

test/interval_tests/complex.jl
```diff
@@ -14,7 +14,6 @@ end
 
 @testset "Complex functions" begin
     Z = (3 ± 1e-7) + (4 ± 1e-7)*im
-    @test sin(Z) == Interval(3.853734949309744, 3.8537411265295543) - Interval(27.016810169394066, 27.016816346613904)*im
-
+    @test sin(Z) == complex(sin(real(Z))*cosh(imag(Z)),sinh(imag(Z))*cos(real(Z)))
```
dpsanders: Nice!

dpsanders:
LGTM, thanks! Feel free to incorporate #159 here.

src/multidim/intervalbox.jl
lbenet: Shall I wait that #159 to change it, or should I leave it like this for the time being? (#161 has some difficulties related to this.)

test/interval_tests/complex.jl
lbenet: Thanks! I though this way is more stable :-)

src/bisect.jl
```diff
@@ -0,0 +1,42 @@
+
+const where_bisect = 0.49609375
```
cvanaret: This looks like a magic constant :) 0.5 - 1/256, if I'm not mistaken. Is there a particular reason for that?

dpsanders:
No particular reason, just something close to 0.5 but not exactly equal to 0.5.

lbenet:
I think we needed something close to zero and to be exactly-representable floating point. Also, I think there were cases were rounding down had some sort of preference.

PR #161, opened 2018-06-21
Add hashing for Intervals

eeshan9815:


eeshan9815:
The hashing tests are passing, but the `mid, diam, × for IntervalBox` test is failing

`mid(IntervalBox(0..1, 3), 0.75)` 
This seems to be the culprit and is causing a `StackOverflowError` as it is recursively calling itself.

test/interval_tests/construction.jl
```diff
@@ -310,3 +310,10 @@ end
 @testset "a..b with a > b" begin
     @test_throws ArgumentError 3..2
 end
+
+@testset "Hashing of Intervals" begin
+    x = Interval{Float64}(1, 2)
+    y = Interval{BigFloat}(1, 2)
+    @test isequal(x, y)
+    @test isequal(hash(x), hash(y))
```
lbenet: Using your definitions of `x` and `y`, `x == y` is `true` in current master, but `x === y` is `false`. Is it ok in that case that `isequal(hash(x), hash(y))` yields `true`?

eeshan9815: According to the definition of the `hash` function, I think this should be fine.
```
help?> hash
  hash(x[, h::UInt])

  Compute an integer hash code such that isequal(x,y) implies
  hash(x)==hash(y). The optional second argument h is a hash code to be mixed
  with the result.
```

eeshan9815:
So tests are failing until #160 is not merged?

test/interval_tests/construction.jl
lbenet: I've tried your implementation with other less trivial cases, which I think you should include as tests. For instance:
`x = @interval(0.1)` and `y = IntervalArithmetic.big53(x)`. In master `x == y` yields `true`, and with your PR we get that `hash(x) == hash(y)` is `true` too. Probably including also a case resulting `false` in an obvious way should also be there.

lbenet:
> So tests are failing until #160 is not merged?

I only know that tests are passing in #160...

test/interval_tests/construction.jl
eeshan9815: Alright, I will. Thanks!

lbenet: And add some docstrings or some explanation in the documentation :-)

lbenet:
#159 and #160 are merged; I've restarted travis tests. Let's see if tests pass now...

@dpsanders Do you have any comments on this PR? LGTM, so I'm in favor of merging it.

dpsanders:
Tests are failing on appveyor.

dpsanders:
@eeshan9815 I think you need to rebase on master to get rid of the failing tests.

eeshan9815:
@dpsanders Rebased on the current master, tests pass now

dpsanders:
Thanks!

PR #169, opened 2018-06-29
Bump minimum Julia version to 0.7-alpha

dpsanders:
FemtoCleaner should run automatically and report output in the `Checks` tab.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ os:
   - osx
 
 julia:
-  - 0.6
+  - 0.7
   - nightly
 
 matrix:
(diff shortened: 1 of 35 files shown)


dpsanders:
@lbenet It seems that you had already hit almost all the points found by FemtoCleaner, except `indmax` -> `argmax`. See the `Checks` tab at the top of the PR page.

lbenet:
> @lbenet It seems that you had already hit almost all the points found by FemtoCleaner, except indmax -> argmax. See the Checks tab at the top of the PR page.

Nice! Can you correct that in this PR?

I am not too sure how shall we proceed with respect to the annotations related to support Julia 0.6 as well (`@compat` usage, or checking `VERSION`). We should continue to support 0.6 for a while (among others, TaylorModels so far works only in 0.6, and depends on current master).

dpsanders:
People are just dropping 0.6 support completely.
We should just tag a new version that supports 0.6, put an upper bound in TaylorModels for the moment, and drop all 0.6 stuff, IMO.

I can just let FemtoCleaner do its magic and it will fix all of this automatically.

lbenet:
I agree with you. Let's get things done for v0.13.1, which still supports Julia 0.6, and then bump drop it in v0.14.0. (Don't take too literally the proposed version numbers.)

lbenet:
You should probably update travis, appveyor and friends...

dpsanders:
The new version should be 0.14, since at least the change to IntervalBox is in principle breaking.

lbenet:
Fully agree with the proposed version numbering.

dpsanders:
Out of date.

PR #176, opened 2018-07-09
Fix broadcasting on 0.7

dpsanders:
This seems to make broadcasting work, but it's very slow, so it's probably not the correct solution.


dpsanders:
Included in #194.

PR #178, opened 2018-07-09
Mod by an interval

dpsanders:
This builds on #145 but provides mod by an interval.

It seems to me that `X mod a` makes sense only if `a` is a positive, atomic (or thin) interval. What do you think, @lbenet?


petvana:
I just came across `mod` as well and implemented a simplified version. Would it be possible to merge at least the not extended version? Is there anything I can help with?

PR #185, opened 2018-07-20
Fix Stackoverflow when creating interval of intervals

### LINKED ISSUE #182: Interval(1..2, 3..4) crashes Julia
[Mistake for IntervalBox.]

Kolaru:
Fix #182. I have define a new type alias of all supported type (basically all built-in number type). Therefore this make the implementation the less generic possible, which is also the safest.

The change also somehow confuse Julia when it comes to finding the most specific method in some cases, for unknown reason.

Finally a small change in the display function to be sure to avoid an infinite recursion by using specifically the `BigFloat` constructor.


lbenet:
Thanks @Kolaru for looking into this. I just checked in Julia 1.0 and the problem reported in #182 persists. 

I like your idea of introducing `IASupportedType` [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/185/files#diff-5fe3ca1a972a5bf936c0f370e3366d0fR14). However, your approach may be too restrictive, in the sense that a user may introduce another struct which is subtype of `Real`, for which it makes sense to define intervals; an example could be `DoubleDouble` or `ArbFloat`.

I think the following would allow for such an enlargement, but I haven't being able to test it in Julia 1.0.
```jl
using InteractiveUtils: subtypes
const IASupportedType = Union{setdiff(subtypes(Real), [AbstractInterval])...}
```
This would result on `IASupportedType` being `Union{AbstractIrrational, AbstractFloat, Integer, Rational}`, which is essentially what you listed.

Another option, as @dpsanders [suggested](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/182#issuecomment-406157353), is that `Interval(::Interval)` throws an error.

Kolaru:
What about a trait-like solution ? Something like

```jl
struct Interval{T} where {T <: Real}
  a::T
  b::T
  Interval{T}(::Type{Val{true}}, a::T, b::T) = new(a, b)
  Interval{T}(::Type{Val{false}}, ::T, ::T) = error("Type $T not good for interval arithmetic.")
end

Interval(a::T, b::T) where {T <: Real} = Interval(Val{isGoodForIA(T)}, a, b)
```
So to make a custom type `A` usable in interval, the user would simply have to write
```jl
struct A
  val
end
isGoodForIA(A) = true
```

Like that absurd intervals are never passed silently, but it can still easily be extended with arbitrary type. Also I don't think this would justifiy adding `SimpleTraits.jl` as dependency, since it is quite simple still.

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
+
 abstract type AbstractInterval{T} <: Real end
 
 struct Interval{T<:Real} <: AbstractInterval{T}
     lo :: T
     hi :: T
 
-    function Interval{T}(a::Real, b::Real) where T<:Real
-
+    function Interval{T}(::Type{Val{true}}, a::Real, b::Real) where {T <: Real}
```
dpsanders: Change to `::Val{true}`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
-
         new(a, b)
-
     end
 end
 
-
+## Traits functions
+canBeIntervalBound(::Type{T}) where {T <: IASupportedType} = true
+canBeIntervalBound(T) = false  # Default case
```
dpsanders: Maybe `canBeIntervalBound(::Type{T}) where {T} = false`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
 
 ## Outer constructors
+function Interval{T}(::Type{Val{false}}, ::Real, ::Real) where {T <: Real}
```
dpsanders: `::Val{false}`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
+    throw(ArgumentError("Type $T is not suitable to construct an interval. If it is a custom type and should be accepted, make sure to define `canBeIntervalBound($T) = true`"))
```
dpsanders: Nice! Change to
`canBeIntervalBound(::Type{$T}) = true`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
+end
 
-Interval(a::T, b::T) where T<:Real = Interval{T}(a, b)
-Interval(a::T) where T<:Real = Interval(a, a)
+Interval{T}(a::Real, b::Real) where {T <: Real} = Interval{T}(Val{canBeIntervalBound(T)}, a, b)
```
dpsanders: I guess you can now remove the `<: Real`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
         end
-
         new(a, b)
-
     end
 end
 
-
+## Traits functions
+canBeIntervalBound(::Type{T}) where {T <: IASupportedType} = true
```
dpsanders: rename to `can_bound_interval`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
+end
 
-Interval(a::T, b::T) where T<:Real = Interval{T}(a, b)
-Interval(a::T) where T<:Real = Interval(a, a)
+Interval{T}(a::Real, b::Real) where {T <: Real} = Interval{T}(Val{canBeIntervalBound(T)}, a, b)
+Interval(a::T, b::T) where {T <: Real} = Interval{T}(Val{canBeIntervalBound(T)}, a, b)
+Interval(a::T) where {T <: Real} = Interval(a, a)
 Interval(a::Tuple) = Interval(a...)
-Interval(a::T, b::S) where {T<:Real, S<:Real} = Interval(promote(a,b)...)
+Interval(a::T, b::S) where {T <: Real, S <: Real} = Interval(promote(a,b)...)
```
dpsanders: Remove `<: Real`

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
 
 ## Concrete constructors for Interval, to effectively deal only with Float64,
 # BigFloat or Rational{Integer} intervals.
-Interval(a::T, b::T) where T<:Integer = Interval(float(a), float(b))
-Interval(a::T, b::T) where T<:Irrational = Interval(float(a), float(b))
+Interval(a::T, b::T) where {T <: Integer} = Interval(float(a), float(b))
```
dpsanders: I think these `<:` can now be removed.

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
...
+Interval(a::T, b::T) where {T <: Irrational} = Interval(float(a), float(b))
 
-eltype(x::Interval{T}) where T<:Real = T
+eltype(x::Interval{T}) where {T <: Real} = T
```
dpsanders: Remove `<: Real`

src/intervals/intervals.jl
dpsanders: Replace by `Val(can_bound_interval(T))`

Kolaru:
Superseeded by #271.

Kolaru:
Wrong PR, sorry for this.

dpsanders:
What's the status here? Should we merge this?

Kolaru:
I need to go back to it to address your last review. I don't remember why I didn't do it earlier, but I'll do it shortly.

src/intervals/intervals.jl
```diff
@@ -11,52 +11,56 @@ else
     const validity_check = false
 end
 
+const IASupportedType = Union{Integer, Rational, Irrational, Float16, Float32,
```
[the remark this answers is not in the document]
Kolaru: I kept it in the latest update, for the definition of the default cases for `can_bound_interval`.

Kolaru:
The coverage decrease comes from the fact `can_be_interval_bound` is not found to be hit by Coveralls, probably because the method is always optimized out by the compiler. I don't think that makes much sense to test it directly, so this PR should be ready.

Kolaru:
This is ready (and may also help avoid some amiguity error I get in #271 when trying to have generic constructors).

PR #187, opened 2018-08-14
Added complex sqrt and some complex set functions

wormell:
This adds a `sqrt` function for `Complex{Interval{T}}`, and some functions such as `in`, `intersect`, `union`, `hull`. 

(The latter two output a `Complex{Interval{T}}`, i.e., a complex square covering the complex squares being unioned. Thus, for what it's worth `hull` isn't really a convex hull per se.)


dpsanders:
Thanks! Some of your tests seem to be failing.

Issue #188, opened 2018-08-14
v0.7 errors


mohamed82008:
So I opened this issue to discuss weird errors and causes in v0.7. The first one I find weird is:
```julia
julia> Interval(1) * (im)
ERROR: ArgumentError: cannot parse "false" as BigFloat
Stacktrace:
 [1] _parse_failure(::Type, ::String, ::Int64, ::Int64) at .\parse.jl:349 (repeats 2 times)
 [2] tryparse_internal at .\parse.jl:345 [inlined]
 [3] parse at .\parse.jl:355 [inlined]
 [4] #6 at C:\Users\user\.julia\dev\IntervalArithmetic\src\intervals\rounding.jl:58 [inlined]
 [5] setrounding(::getfield(IntervalArithmetic, Symbol("##6#8")){String}, ::Type{BigFloat}, ::RoundingMode{:Down}) at .\rounding.jl:158
 [6] #5 at C:\Users\user\.julia\dev\IntervalArithmetic\src\intervals\rounding.jl:57 [inlined]
 [7] setprecision(::getfield(IntervalArithmetic, Symbol("##5#7")){String,RoundingMode{:Down}}, ::Type{BigFloat}, ::Int64) at .\mpfr.jl:904
 [8] atomic at C:\Users\user\.julia\dev\IntervalArithmetic\src\intervals\rounding.jl:56 [inlined]
 [9] macro expansion at .\show.jl:556 [inlined]
 [10] convert(::Type{Interval{Float64}}, ::Bool) at C:\Users\user\.julia\dev\IntervalArithmetic\src\intervals\conversion.jl:16
 [11] _promote at .\promotion.jl:261 [inlined]
 [12] promote at .\promotion.jl:284 [inlined]
 [13] * at .\promotion.jl:314 [inlined]
 [14] *(::Interval{Float64}, ::Complex{Bool}) at .\complex.jl:299
 [15] top-level scope at none:0
```
This seems to be related to some promotion changes in Julia?

Using `1im` instead solves the problem.

mohamed82008:
I think something is happening with the `convert` function. In v0.6.4, calling `Interval(1.)*im)` triggers the convert in the line 
https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/cab2433d6719899c38c0c3514fc935208614bdd0/src/intervals/conversion.jl#L16
[...]

PR #189, opened 2018-08-15
Fixes Interval-Boolean multiplication on Windows

mohamed82008:
This PR fixes the error noticed in https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/188#issue-350628965.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ os:
 
 julia:
   - 0.6
+  - 0.7
   - nightly
 
 matrix:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,8 @@ environment:
   matrix:
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7-latest-win32.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
 
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -13,6 +13,7 @@ promote_rule(::Type{BigFloat}, ::Type{Interval{T}}) where T<:Real =
 
 
 # convert methods:
+convert(::Type{Interval{T}}, x::Bool) where {T} = convert(Interval{T}, Int(x))
 convert(::Type{Interval{T}}, x::Real) where {T} = atomic(Interval{T}, x)
 convert(::Type{Interval{T}}, x::T) where {T<:Real} = Interval{T}(x)
 convert(::Type{Interval{T}}, x::Interval{T}) where {T} = x


lbenet:
Thanks for the PR. 

I suggest to include some tests, precisely to catch those situations. A good start is the one that lead to #188

lbenet:
Incidentally, I just noticed that current checks of travis and appveyor do not check Julia 0.7; they do it for Julia 1.0.

mohamed82008:
But the error is only on Windows and Travis doesn't check Windows.

lbenet:
See #190

mohamed82008:
Hmm the error is already there https://ci.appveyor.com/project/dpsanders/intervalarithmetic-jl/build/1.0.117/job/qufll6qvubwxghqp#L176. I will setup appveyor for Windows like that PR then.

lbenet:
I am no expert in appveyor or travis. However, I think there are still some problems to deal with. Can you check that *locally* tests pass?

mohamed82008:
Rebased, let's if the error goes away or not.

mohamed82008:
https://ci.appveyor.com/project/dpsanders/intervalarithmetic-jl/build/1.0.118/job/d70odm4dem5m28ar#L175

These tests now pass on Windows.

lbenet:
So do you understand the `ayan2` problem?

mohamed82008:
This is the offending line:
```julia
julia> atan2(@interval(0.0, 1.0), @biginterval(0.0))
┌ Warning: `atan2(y, x)` is deprecated, use `atan(y, x)` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
ERROR: StackOverflowError:
```

mohamed82008:
Julia is not finding https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/cab2433d6719899c38c0c3514fc935208614bdd0/src/intervals/trigonometric.jl#L207 so it keeps infinitely calling https://github.com/JuliaLang/julia/blob/72af6204a7fa448dd1bdc38305e78533b4a83040/base/math.jl#L559. Calling the deprecated `atan2` makes a warning then calls the replacement `atan` which is not overloaded by `IntervalArithmetic` so it causes the infinite recursion as `atan(y::Real, x::Real)` is the most specific method when `y` and `x` are `Interval{BigFloat}`. Changing the name of the method to `atan` fixes the error.

dpsanders:
Thanks! Could you send a PR for that change, please?

dpsanders:
Thanks. I guess these should be too separate PRs, though.

dpsanders:
@lbenet Do you know what is going on with the `bisect` tests?

lbenet:
Actually, I don't know. 

I've been trying to get tests pass, but somehow there are too many things which are breaking or broken in Julia 0.7.0.

mohamed82008:
Separate PRs made. Both rebased on https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/190 so either close #190 or merge it first.

lbenet:
I think we should tag a version (0.14.1) for whatever is in master now, as the last version which runs in Julia 0.6, and then bump into 0.7.0. 

I think this imposes (at least) some changes around: 
- ValidatedNumerics, should be tagged to run in Julia 0.6 with ValidatedNumerics.jl up to (but probably not including 0.14.0). 
- We should also tag a version (v0.7.0?) of CRlibm.jl which again will be the last to work in Julia 0.6.x

lbenet:
So how do we proceed?

dpsanders:
Is there any breaking change on master? If so, then the new tagged version for Julia 0.6 should be 0.15.

The current version of CRlibm is the one that works for Julia 0.6 (except for the weird compilation problem reported recently), so I don't  think it needs another tag right now.

Maybe it would be useful to make an issue with the list of all the problems that need to be fixed for 0.7, so that we can fix them one by one.

dpsanders:
As far as I can tell, master === v0.14, so there is no need for another release.

dpsanders:
Included in #194.

dpsanders:
Thanks!

PR #190, opened 2018-08-15
Tests for Julia 0.7

lbenet:


lbenet:
@dpsanders There is a broken test related to `atan2` in Julia 0.7.0. The same occurs in #189

dpsanders:
Included in #194.

Issue #103, opened 2018-03-11
Power ^ allocates


saschatimme:
It seems that `x^2` currently allocates, which I think should not happen since Base has a special handling for these cases.
```jl
julia> using IntervalArithmetic, BenchmarkTools
julia> x = interval(rand())
[0.375296, 0.375297]
julia> pow(a) = a^2
pow (generic function with 1 method)
julia> @benchmark pow($x)
BenchmarkTools.Trial: 
  memory estimate:  2.15 KiB
  allocs estimate:  55
  --------------
  minimum time:     4.419 μs (0.00% GC)
  median time:      4.777 μs (0.00% GC)
  mean time:        6.551 μs (20.78% GC)
  maximum time:     2.424 ms (99.35% GC)
  --------------
  samples:          10000
  evals/sample:     7
```

I found this comment
```julia
# overwrite new behaviour for small integer powers:
# ^{p}(x::IntervalArithmetic.Interval, ::Type{Val{p}}) = x^p
```
So is this just not implemented or is there a specific reason to not include these?

dpsanders:
This is an unfortunate corner of the package at the moment:

The allocation is due to the fact that the only way to get *correct rounding* for power is using MPFR, so this is in fact converting the interval to a `BigFloat` interval, doing the power, and converting back :/

However, if you are happy to have a slightly wider interval, there is actually an exported function `pow` that uses a fast path (power by squaring for integer powers):

```
julia> @btime ($x)^2
@btime pow($x, 2)
  3.266 μs (51 allocations: 2.02 KiB)
Interval(0.3183098861837905, 0.5000000000000003)

julia> @btime pow($x, 2)
  69.124 ns (0 allocations: 0 bytes)
Interval(0.3183098861837905, 0.5000000000000003)
```

We had a discussion in https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/14 about whether to make `pow` the default behaviour.

In any case, there should certainly be an easy way to choose to use fast powers: 
https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/56

saschatimme:
The problem with a special function like `pow` is that I cannot use this library as a drop-in replacement. I would like to provide a fast (default) implementation but also a verified implementation using interval arithmetic without any changes to the code base.
At least I write `x^2` only instead of `x * x` since I know that the former is lowered to the latter. So maybe one could argue that here you should also use the small integer power behaviour from base.

[...]

Issue #14, opened 2017-04-23
Make fast powers (pow) the default


dpsanders:
Make `^` use the fast version and `pow` the slow but precise version using `BigFloat` (i.e., invert what currently happens).

dpsanders:
Make ^ use the fast version and pow the slow but precise version using bigfloat.

dpsanders:
What do you think about this @lbenet?

lbenet:
I think the default should be to get the tightest intervals, and it that sense, it shouldn't be the default. I wouldn't mind to have it as default for the `:wide` rounding mode, though. 

Do you think that the current `^` could be speed-up using `power_by_squaring` somewhere?

dpsanders:
In this particular case, I don't agree. The reason is simply that using the tight `^` requires the use of `BigFloat` (rounding powers is difficult), and hence is reaallly slooow. 

For a real use case I just tried, just changing the two instances of `^` to `pow` in a function that is evaluated many times gave a **10x** speedup.

As soon as `power_by_squaring` is used, tightness is lost -- but (a lot of) speed is gained.
(I believe it is still possible to speed up, and possibly make more accurate, `pow` in `:errorfree` mode.) 

I also don't think that we should change the behaviour based on the rounding type (more than is necessary).

lbenet:
Just to make my point clear: I am in favor of the slow version (`^`), since it assures tighter intervals.

My comment about the `:wide` mode is, simply,  that in this case we already loose the tightness of the intervals, and therefore it may be appropriate to have `^` equivalent `to `pow`. But I guess this may complicate things a lot.

You are probably right; using `Base.power_by_squaring` may not be useful as such to get tighter intervals. What I had in mind was to use the same algorithm but with tighter versions (e.g., `a^2` instead of `a*a`). I'll think about it...

dpsanders:
My proposal is not to remove the tight version, just to not make it the default. 
In my opinion, it's more of a "gotcha" (unpleasant, unexpected outcome) to surprise the user by making the standard functions too slow, rather than to document that a *very slightly* tighter option is the `pow` function (in the new version).

lbenet:
I understand your point for the "gotcha" about the speed, and agree that the problem lies in the conversion to BigFloat's, performing there the calculations, and convert back to Float64. 

[...]

PR #210, opened 2018-09-16
Speed up pow by 4x or more

dpsanders:


dpsanders:
Tests will fail until `pow` and `^` are swapped over.

lbenet:
I've taken a look on this PR, and must begin saying that I understand the motivation, but I am not yet convinced that we should sacrifice tightness for speed; see #14.

As it was discussed in #103, current `^` allocates since we must play with `big53`. The question is if intercepting small powers can help us here?

Regarding the infimum and supremum of an empty interval, see [this comment](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/232#issuecomment-424948383); we are currently following the standard on this, so this shouldn't be changed. Reverting the changes ([here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/210/files#diff-d9d17fed755178cc4f7291ba34870545R285) and [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/210/files#diff-d9d17fed755178cc4f7291ba34870545R291)) in arithmetic.jl allows some tests to pass.

Very nice tricks for `fast_sqrt`. The question is how does the diameter of `Interval(fast_sqrt(a.lo , RoundDown)), fast_sqrt(a.hi , RoundUp))` compare with the results for `^(a, 0.5)` and `pow(a, 0.5)`?

dpsanders:
The fast version of `sqrt` is now in `FastRounding` and will be used automatically once https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/230 is merged. The fast version is (supposed to) give exactly the same result as the slow version.

The fast powers in this PR give a result that is "almost" as tight as the terribly slow correctly-rounded version - for higher powers they will be less tight, since more squarings are required. In particular, squares (`x^2`) will give exactly the same result as now (but repeated squares may start losing a few ulps of accuracy).

I really think it's urgent that we make this change. Tight powers will still be available via `pow`. We could provide a way for automatically using tight powers instead, but I'm not sure what the API should look like (there are early open issues for this, e.g. #14 that you cited).

dpsanders:
I have been using fast powers for a few months now. 
The difference in speed and allocations is like night and day. There is no question in my mind that we *must* make fast powers the default (but provide a mechanism to change to tight-but-slow powers).

lbenet:
Tests are failing; I guess those that fail are related to `^`, and they should be using `pow`, right?

dpsanders:
Yes, exactly.

dpsanders:
That will require changes to itf1788. I think I have those in a branch somewhere.

lbenet:
I guess the same holds; instead of using `^` they should use `pow`...

mforets:
Is there a chance of getting this function in `master` but with another name, eg. `fast_pow(a, n)`?

dpsanders:
It is already there with the name `pow`,  I believe?

mforets:
I thought this branch gives a faster `pow` than master but it hasn't been resolved yet because it sacrifices tightness for speed. On `master` i measure ~1us for `Interval(0, 1)^2`, so i was assuming that it's using the "slow" version (?).

mforets:
.. so if that's the case i was just proposing to export the "fast" version, possibly with a different name, keeping the slow one as default.

dpsanders:
I get this on master:
```jl
julia> using BenchmarkTools, IntervalArithmetic
[ Info: Precompiling IntervalArithmetic [d1acc4aa-44c8-5952-acd4-ba5d80a2a253]

julia> @btime $(Ref(X))[]^2^C

julia> X = 0..1
[0, 1]

julia> @btime $(Ref(X))[]^2
  673.264 ns (29 allocations: 1.25 KiB)
[0, 1]

julia> @btime pow($(Ref(X))[], 2)
  30.811 ns (0 allocations: 0 bytes)
[0, 1]
```

But yes you're right, the `nonrecursive_powers` branch speeds this up by 3x.

mforets:
Oh, i see what you mean know. I was missing that there's already a difference between `pow(a, n)` and `a^n`, and it's in the [docs](https://juliaintervals.github.io/IntervalArithmetic.jl/latest/api/#IntervalArithmetic.pow-Tuple{Interval,Integer})... 

> But yes you're right, the nonrecursive_powers branch speeds this up by 3x.

Alright, i will keep it in mind. Thanks!

dpsanders:
We should definitely update `pow` to use the fast version, and then provide a way to use that instead of `^`.

lbenet:
I agree with @mforets that this should be in master with another name (`fast_pow`), without touching the current `pow` for compatibility with the standard.

@dpsanders Do you have time to address the conflicts with the files currently in master?

dpsanders:
Yes we should definitely incorporate the latest version into base in the `pow` function.

OlivierHnt:
Solved in PR #593.

PR #218, opened 2018-09-18
Fix @format and format display

dpsanders:


lbenet:
Thanks a lot for this. I think it is worth to update the related documentation on this.

dpsanders:
@lbenet I have updated the docs and this should be ready to merge.

lbenet:
Merging! Thanks a lot!

PR #220, opened 2018-09-19
==(DecoratedInterval,Interval) == false, hotfix for #219

### LINKED ISSUE #219: ∞🔁🐞 : ==(DecoratedInterval, Interval) gives an infinite loop.
```
using IntervalArithmetic
DecoratedInterval(1,2) == Interval(1, 2)
```

Versioninfo()
``` [d1acc4aa] IntervalArithmetic v0.15.0+ [`~/.julia/dev/IntervalArithmetic`]```

```
julia> versioninfo()
Julia Version 1.0.0
Commit 5d4eaca0c9 (2018-08-08 20:58 UTC)
Platform Info:
  OS: Linux (i686-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i3-7100 CPU @ 3.90GHz
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
```

### PR DESCRIPTION
**Pull request by miguelraz**
Fixes #219.
Added tests so as to not regress coverage, guarantee desired case passes.

### DIFF
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -33,6 +33,9 @@ end
 
 in(x::T, a::DecoratedInterval) where T<:Real = in(x, interval_part(a))
 
+@inline Base.:(==)(xx::DecoratedInterval, yy::Interval) = false
+@inline Base.:(==)(y::Interval, xx::DecoratedInterval) = false
+
 
 ## scalar functions: mig, mag and friends
 scalar_functions = (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/decorations/functions.jl [hunk 1]
```diff
@@ -33,6 +33,9 @@ end
 
 in(x::T, a::DecoratedInterval) where T<:Real = in(x, interval_part(a))
 
+@inline Base.:(==)(xx::DecoratedInterval, yy::Interval) = false
```
dpsanders: `==` should already be imported, so you can just write `==`. Inlining should be automatic for a simple function like this I guess.

dpsanders:
I don't think we want to go in this direction. I would suggest throwing an `ArgumentError`.

PR #226, opened 2018-09-21
Speed up trivials.

miguelraz:
Consider
```
julia> @btime nai()
  46.999 ns (1 allocation: 32 bytes)
[NaN, NaN]
julia> @btime emptyinterval()
  47.715 ns (1 allocation: 32 bytes)
∅
julia> @btime entireinterval()
  45.143 ns (1 allocation: 32 bytes)
[-∞, ∞]
```

versus
```
julia> const zentire() = Interval(-Inf,Inf)
zentire (generic function with 1 method)

julia> @btime zentire()
  1.315 ns (0 allocations: 0 bytes)
[-∞, ∞]

julia> entireinterval() == zentire()
true

julia> const zempty() = Interval(Inf,-Inf)
zempty (generic function with 1 method)

julia> zempty() == ∅

julia> const znai() = Interval(NaN,NaN)
znai (generic function with 1 method)

julia> @btime znai()
  1.352 ns (0 allocations: 0 bytes)
[NaN, NaN]

julia> znai() === nai()
true

julia> @btime zempty()
  1.354 ns (0 allocations: 0 bytes)
∅
```
All tests pass locally and should help speed up testing a bit.

### DIFF
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -33,6 +33,9 @@ end
 
 in(x::T, a::DecoratedInterval) where T<:Real = in(x, interval_part(a))
 
+@inline Base.:(==)(xx::DecoratedInterval, yy::Interval) = false
+@inline Base.:(==)(y::Interval, xx::DecoratedInterval) = false
+
 
 ## scalar functions: mig, mag and friends
 scalar_functions = (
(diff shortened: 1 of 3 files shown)


miguelraz:
Polluted branch. brb.

PR #227, opened 2018-09-21
🏎️ Speed up trivials

miguelraz:
Consider

```
julia> @btime nai()
  46.999 ns (1 allocation: 32 bytes)
[NaN, NaN]
julia> @btime emptyinterval()
  47.715 ns (1 allocation: 32 bytes)
∅
julia> @btime entireinterval()
  45.143 ns (1 allocation: 32 bytes)
[-∞, ∞]
```
versus
```
julia> const zentire() = Interval(-Inf,Inf)
zentire (generic function with 1 method)

julia> @btime zentire()
  1.315 ns (0 allocations: 0 bytes)
[-∞, ∞]

julia> entireinterval() == zentire()
true

julia> const zempty() = Interval(Inf,-Inf)
zempty (generic function with 1 method)

julia> zempty() == ∅
true

julia> @btime zempty()
  1.354 ns (0 allocations: 0 bytes)
∅

julia> const znai() = Interval(NaN,NaN)
znai (generic function with 1 method)

julia> @btime znai()
  1.352 ns (0 allocations: 0 bytes)
[NaN, NaN]

julia> znai() === nai()
true
```

### DIFF
--- a/src/intervals/special.jl
+++ b/src/intervals/special.jl
@@ -8,7 +8,7 @@ that this interval is an exception to the fact that the lower bound is
 larger than the upper one."""
 emptyinterval(::Type{T}) where T<:Real = Interval{T}(Inf, -Inf)
 emptyinterval(x::Interval{T}) where T<:Real = emptyinterval(T)
-emptyinterval() = emptyinterval(precision(Interval)[1])
+const emptyinterval() = Interval(Inf,-Inf)
 const ∅ = emptyinterval(Float64)
 
 isempty(x::Interval) = x.lo == Inf && x.hi == -Inf
@@ -19,7 +19,7 @@ const ∞ = Inf
 """`entireinterval`s represent the whole Real line: [-∞, ∞]."""
 entireinterval(::Type{T}) where T<:Real = Interval{T}(-Inf, Inf)
 entireinterval(x::Interval{T}) where T<:Real = entireinterval(T)
-entireinterval() = entireinterval(precision(Interval)[1])
+const entireinterval() = Interval(-Inf,Inf)
 
 isentire(x::Interval) = x.lo == -Inf && x.hi == Inf
 isinf(x::Interval) = isentire(x)
@@ -30,7 +30,7 @@ isunbounded(x::Interval) = x.lo == -Inf || x.hi == Inf
 """`NaI` not-an-interval: [NaN, NaN]."""
 nai(::Type{T}) where T<:Real = Interval{T}(convert(T, NaN), convert(T, NaN))
 nai(x::Interval{T}) where T<:Real = nai(T)
-nai() = nai(precision(Interval)[1])
+const nai() = Interval(NaN,NaN)
 
 isnai(x::Interval) = isnan(x.lo) || isnan(x.hi)
 


lbenet:
Thanks @miguelraz for addressing this. 

I like your approach. The question is if we can generalize it to work for other parameterizations of `Interval`, e.g., `Interval{BigFloat}`. If this can be done, I would suggest to exchange the current definition of all those functions by the one you are introducing.

miguelraz:
Besides all tests passing in `IntervalArithmetic.jl`, you don't lose information when operating Intervals/DecoratedIntervals with these trivial cases.
See:
```
julia> Interval{Float32}(NaN,NaN) + Interval{BigFloat}(1,2)
Interval(NaN, NaN)

julia> typeof(ans)
Interval{BigFloat}
```
julia> Interval{Float32}(∅) + Interval{BigFloat}(1,2)
∅

julia> typeof(ans)
Interval{BigFloat}

julia> Interval{Float32}(-Inf,Inf) + Interval{BigFloat}(1,2)

Interval(-Inf, Inf)

julia> typeof(ans)
Interval{BigFloat}

miguelraz:
Yes, @lbenet , we can lose information if we don't default to the `Float32` case.
```
julia> nai() + Interval{Float32}(1,2)
Interval(NaN, NaN)

julia> typeof(ans)
Interval{Float64}
```

miguelraz:
Huh, it gets worse.
```
julia> @btime Interval{BigFloat}(Inf,-Inf)
  90.825 ns (5 allocations: 172 bytes)
∅

julia> @btime Interval{Float32}(Inf,-Inf)
  1.319 ns (0 allocations: 0 bytes)
```
I should amend the PR to call all these trivial functions on the `Float32` case and trust that promotion will take care of it.
Spoiler: it doesn't because of some consistency tests.
```
(v1.0) pkg> Total
Consistency tests                          |  198     3     31    232
  Interval types and constructors          |   27            5     32
  inv                                      |    9     1      2     12  fma consistency                          |    7                   7
  ∈ tests                                  |    5                   5
  Inclusion tests                          |   11           12     23  Comparison tests                         |   15            5     20
  Intersection tests                       |    6            2      8  Hull and union tests                     |    4                   4  Special interval tests                   |   12     1      5     18
  mid                                      |    7                   7
  mid with parameter                       |    4                   4  mid with large floats                    |    4                   4
  diam                                     |    5                   5  mig and mag                              |    6                   6
  cancelplus tests                         |   22                  22  mid and radius                           |    6     1             7
  abs, min, max, sign                      |   22                  22  Precision tests                          |    4                   4  Interval power of an interval            |    4                   4
  isatomic                                 |    6                   6  iszero                                   |    7                   7  Difference between Interval and interval |    5                   5
```

dpsanders:
There is already `nai(T)`, e.g. `nai(Float64)`. 

I am actually in favour of getting rid of this global precision mode completely in any case.

dpsanders:
Any default should be `Float64`.

dpsanders:
I'm worried about type stability: if we return `nai()` from a function that we usually return `Interval{BigFloat}`, we get type instability. In those cases it should be `nai(T)`.

lbenet:
I fully agree with David's comment. That's why I was thinking on returning intervals with different parameterizations.

OlivierHnt:
The performance regression should be fixed on master.

PR #230, opened 2018-09-26
Use fast path for sqrt and inv

dpsanders:
- [x] Add / define missing methods for big intervals


dpsanders:
Tests are failing since e.g. `sqrt(x::Interval{BigFloat})` is no longer correctly defined.

dpsanders:
This is ready @lbenet.

dpsanders:
If there's no objections, I'll merge this soon and tag a patch release.

lbenet:
There is no objection to merge this, so please, go ahead. 

(I am sorry for the delay I may have caused, but the last few days have been too busy so I have only come here as a random walker.)

PR #238, opened 2018-10-25
Add IntervalBox constructor from vectors giving opposite corners

dpsanders:
Construct an `IntervalBox` from opposite corners, even if those intervals are the wrong way round.

The function `force_interval` that creates an interval even when the endpoints are the wrong way round could be generally useful. It probably needs a better name.

### DIFF
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -27,6 +27,9 @@ end
 
 IntervalBox(lo::AbstractVector, hi::AbstractVector) where {N,T} = IntervalBox(force_interval.(lo, hi))
 
+IntervalBox(lo::SVector{N,T}, hi::SVector{N,T}) where {N,T} = IntervalBox(force_interval.(lo, hi))
+
+
 Base.@propagate_inbounds Base.getindex(X::IntervalBox, i) = X.v[i]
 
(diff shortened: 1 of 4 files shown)


src/multidim/intervalbox.jl
```diff
@@ -17,6 +17,16 @@ IntervalBox(x...) = IntervalBox(x)
 IntervalBox(x) = IntervalBox(x...)
 IntervalBox(X::IntervalBox, n) = foldl(×, Iterators.repeated(X, n))
 
+# construct from two vectors giving bottom and top corners
```
lbenet: It is more appropriate to have the function `force_interval` in `intervals/interval.jl`. 

I also think the returned intervals should use `Interval` instead of `interval`, since the latter checks (again) if the actual interval is valid.

lbenet:
By the way, there are conflicts to merge this in master...

src/multidim/intervalbox.jl
dpsanders: I moved `force_interval` as you suggested, and added tests.

It has to check validity for cases like `force_interval(3, NaN)`.

lbenet: Thanks for the clarification!

lbenet:
Thanks a lot! Merging.

PR #243, opened 2018-11-28
Adjust multiplication of intervals to take account of 0 * Inf == 0

### LINKED ISSUE #242: Nasty bug in multiplication
```jl
julia> (0..∞) * (-1..∞)
[NaN, ∞]
```

Reported by Charlie Vanaret.

dpsanders:
Fixes #242.

This is surprisingly ~painful~ annoying since basically the case `0 * Inf` has to be explicitly caught.
I decided to do it by using a separate path whenever one or the other of the intervals is infinite, and just following the same logic that was already implemented. There may be a better way by just identifying the (few?) cases where the problem actually might arise.

cc @lbenet, @cvanaret


dpsanders:
This is ready (the failure on Appveyor is due to an old version of Julia).

lbenet:
Thanks a lot for this! Merging.

PR #244, opened 2018-11-28
Add "in" for IntervalBox

### LINKED ISSUE #240: "in" operator (∈) works differently for Interval and IntervalBox
```jl
julia> 2 ∈ 1..3
true

julia> [2, 2] ∈ IntervalBox(1..3, 1..3)
false

julia> 1..3 ∈ IntervalBox(1..3, 1..3)
true
```

It is expected since this behavior corresponds to the one expected when thinking about iterating over the components of an `IntervalBox`, but at the same time the discrepency is kind of confusing. The fix is probably to document all function that works only on `Interval`.

dpsanders:
Fixes #240.

Defines ` ∈` for `IntervalBox` correctly.

cc @Kolaru

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -28,7 +28,7 @@ import Base:
     union, intersect, isempty,
     convert, promote_rule, eltype, size,
     BigFloat, float, widen, big,
-    ∩, ∪, ⊆, ⊇, eps,
+    ∩, ∪, ⊆, ⊇, ∈, eps,
     floor, ceil, trunc, sign, round,
     expm1, log1p,
     precision,
--- a/src/intervals/complex.jl
+++ b/src/intervals/complex.jl
@@ -1,3 +1,6 @@
+
+⊆(x::Complex{Interval{T}}, y::Complex{Interval{T}}) where T = real(x) ⊆ real(y) && imag(x) ⊆ imag(y)
+
 function ^(x::Complex{Interval{T}}, n::Integer) where {T}
     if n < 0
         return inv(x)^n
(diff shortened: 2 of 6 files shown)


Kolaru:
Looks globally good to me, I only have some minor improvement propositions:
- Defining `in` for `Vector` as well, so that `[0, 0] in IntervalBox(-1..1, -1..1)` does not error.
- Testing a case where `v in X` is `false`.

dpsanders:
OK, will define for `AbstractVector{T}`. 

I realised that the problem is that the type T inside the vector should *not* be allowed to be an interval though? Or maybe that's not important.

dpsanders:
The reason for restricting to `SVector{N}` was that then you guarantee that the vector is of the correct length.

Also, is there any particular reason why you are using standard Julia vectors instead of `SVector`s?

Kolaru:
> I realised that the problem is that the type T inside the vector should _not_ be allowed to be an interval though? Or maybe that's not important.

Technically you want `T` to be a type that can be a bound for an interval. Disallowing `Interval` is good but does not cover all cases. Doing so probably requires something like what is proposed in #185.

> Also, is there any particular reason why you are using standard Julia vectors instead of `SVector`s?

Only when testing stuff in REPL as standard vectors are quicker to type.

dpsanders:
Yes I agree that something more is required, but I think this is as far as this PR needs to go. Do you agree?

Kolaru:
Yes, the "more" should be dealt with in a separate PR (I'll look in renaming/polishing my proposal soon).

The current PR looks good to me, by the way.

lbenet:
Is this ready for reviewing/merging? I actually agree with @Kolaru's view that this is already ok.

dpsanders:
Yes, in my opinion it's ready to merge.

lbenet:
Thanks!

PR #246, opened 2018-12-05
Fix printing of intervals

dpsanders:
Fixes printing of intervals on master.
The non-public function `to_mpfr` that was previously used was removed in https://github.com/JuliaLang/julia/pull/29157.


lbenet:
Fine with me. merging. 

Thanks a lot!

PR #250, opened 2018-12-10
Define specialised abs2 and abs for complex intervals

dpsanders:


test/interval_tests/complex.jl
```diff
@@ -44,3 +46,10 @@ end
     a = -3.1
     @test x ⊆ (x^a)^(1/a)
 end
+
+@testset "abs2 and abs" begin
```
Kolaru: I suggest to use either a Pythagoras triplet or something similar so to have exact result for `abs2(x)` and `abs(x)`, so the result looks a bit less magic, e.g.

```jl
x = (0..3) + (0..4)*im
@test abs2(x) = Interval(0.0, 25.0)
@test abs(x) = norm(x) = Interval(0.0, 5.0)
```

Also in the current master we have `abs2( (-1..1) + (-2..2)*im == (-5..5)`, so I suggest to add the following tests:

```jl
y = (-1..1) + (-2..2)*im
@test abs(y).lo == 0.0
@test abs2(y).lo = 0.0
```

dpsanders: Good idea, thanks -- I've implemented both.
I believe you could have actually suggested those directly as code changes in the PR.

dpsanders: Hmm, maybe that's not yet enabled.

test/interval_tests/complex.jl
```diff
@@ -44,3 +46,13 @@ end
...
+    x = (0..3) + (0..4)*im
+    @test abs2(x) == 0..25
+    @test abs(x) == norm(x) == 0..5
+
+    y = (-1..1) + (-2..2)*im
+    @test abs(y).lo == 0.0
+    @test abs2(y).lo = 0.0
```
Kolaru:
```suggestion
    @test abs2(y).lo == 0.0
```

The tests fails because of this line, I should have tested it before suggesting.

It's the occasion to test the "suggest change feature".

dpsanders: Thanks! I should have checked more carefully (and tested locally).

Issue #237, opened 2018-10-20
Remove Interval <: Real ?


dpsanders:
Since ForwardDiff now no longer requires types to be `<: Real` (https://github.com/JuliaDiff/ForwardDiff.jl/pull/369, although not yet in a released version), maybe it is time to make `Interval`s no longer be subtypes of `Real`.

cc @gwater

Kolaru:
I am currently looking into that issue and complex numbers appear to be problematic. `Complex{T}` requires `T <: Real`, which means our current definition of complex intervals (`Complex{Interval{Float64}}`) would no longer be valid.

Even if the requirement to have `Real` components was removed from Base, it would kind of be awkward to have `(0.1..0.5)` not be a `Real` while `(0.1..0.5) * im` is a `Complex` (which is inavoidable to reuse the definitions from Base).

So I think we will have to keep `Interval` as `Real`.

gwater:
> it would kind of be awkward to have `(0.1..0.5)` not be a `Real` while `(0.1..0.5) * im` is a `Complex` (which is inavoidable to reuse the definitions from Base).

I'm not exactly sure how `Complex{Interval}` currently behaves. In mathematical terms it should be equivalent to a 2-dimensional `IntervalBox`, no? Or is it more like a straight line on the complex plane? In either case, it raises the same questions as the real case: Are complex intervals sets of numbers or do they each represent one specific (but uncertain) number? In the latter case subtyping makes sense in the former it's problematic.

Kolaru:
> I'm not exactly sure how `Complex{Interval}` currently behaves.

A `Complex{Interval}` is simply an object `A + B*im` with `A` and `B` intervals, so it is like an 2-dimensional `IntervalBox`.


> Are complex intervals sets of numbers or do they each represent one specific (but uncertain) number? In the latter case subtyping makes sense; in the former it's problematic.

While I agree with this analysis (and prefer the former interpretation), it lefts out the following practical concern: all `Base` complex functions for a complex number `a + b*im` are defined in term of operation on `a` and `b` and as a consequence, the current definition of complex interval can reuse all the definitions from `Base` (with the exception of some functions doing branching as in #245 ). Not subtyping complex interval to `Complex` implies to either redefine all these functions, or to remove the requirement for `Complex` to have `Real` components and have `Complex{T}` not always be a subtype of `Complex`. Both solution require a lot of work that I am not sure is worth for the sake of mathematical purity.

[...]

PR #254, opened 2018-12-12
Implement interval types that are not Real

Kolaru:
Here is a very restricted implementation of a solution to #237.

This PR defines 3 new types of intervals, all wrapping an internal `BaseInterval`:
- `SetInterval` is not a subtype of `Real` and error on ambiguous cases.
- `RealSetInterval` is the same as the previous, but is a subtype of `Real`.
- `PseudoNumberInterval` is a subtype of `Real` and return `false` on ambiguous cases.

`BaseInterval` is equivalent to the current `Interval` except for function returning boolean, for which it uses three valued logic, returning `[true, false]` on ambiguous cases as both `true` and `false` are possible. The wrappers then implement different response to this.

A macro `@interval_function` is also implemented such that it automatically extends functions defined for `BaseInterval` to the three wrappers types with the correct behavior.

In the current state of this PR only the functions `lo`, `hi`, `in` and `iszero` are available to illustrate the intended behavior. These functions are defined at the end of the `wrapper_intervals.jl` file.

Most of the package has been commented out to have it testable without having to do all the work before getting some review.

Here is an example of how it works:

```jl
julia> a = SetInterval(-1, 1)
SetInterval{Float64}(BaseInterval{Float64}(-1.0, 1.0))

julia> 0 in a
true

julia> iszero(a)
ERROR: AmbiguousError: Result to "# TODO" is ambiguous.
Stacktrace:
 [1] adapt_result at C:\Users\Kolaru\.julia\dev\IntervalArithmetic\src\intervals\wrapper_intervals.jl:46 [inlined]
 [2] iszero(::SetInterval{Float64}) at .\none:0
 [3] top-level scope at none:0

julia> b = PseudoNumberInterval(-1, 1)
PseudoNumberInterval{Float64}(BaseInterval{Float64}(-1.0, 1.0))

julia> 0 in b
true

julia> iszero(b)
false
```

(The error message is not yet finished, hence the "# TODO" tag)

If the community judges it is the way to go, I'll come up with a todo list for the (massive) overhaul this would imply for the package.

cc @gwater @lbenet @dpsanders

### DIFF
--- a/src/intervals/special.jl
+++ b/src/intervals/special.jl
@@ -1,4 +1,4 @@
-# This file is part of the BaseIntervalArithmetic.jl package; MIT licensed
+# This file is part of the IntervalArithmetic.jl package; MIT licensed
 
 ## Definitions of special intervals and associated functions
 
--- a/src/intervals/special.jl
+++ b/src/intervals/special.jl
@@ -1,41 +1,47 @@
-# This file is part of the IntervalArithmetic.jl package; MIT licensed
+# This file is part of the BaseIntervalArithmetic.jl package; MIT licensed
 
 ## Definitions of special intervals and associated functions
 
+const ∞ = Inf
+
 ## Empty interval:
-"""`emptyinterval`s are represented as the interval [∞, -∞]; note
+"""
+    emptyinterval(::Type{T})
+
+Return an `emptyinterval` with interval bounds of type `T`.
+
+`emptyinterval`s are represented as the interval [∞, -∞]; note
 that this interval is an exception to the fact that the lower bound is
 larger than the upper one."""
-emptyinterval(::Type{T}) where T<:Real = Interval{T}(Inf, -Inf)
-emptyinterval(x::Interval{T}) where T<:Real = emptyinterval(T)
-emptyinterval() = emptyinterval(precision(Interval)[1])
+emptyinterval(::Type{T}) where {T<:Real} = Interval{T}(Inf, -Inf)
+emptyinterval() = emptyinterval(precision(BaseInterval)[1])
 const ∅ = emptyinterval(Float64)
 
-isempty(x::Interval) = x.lo == Inf && x.hi == -Inf
+emptyinterval(x::IT) where {IT<:AnyInterval} = AnyInterval(Inf, -Inf)
 
-const ∞ = Inf
+isempty(x::BaseInterval) = x.lo == Inf && x.hi == -Inf
 
 ## Entire interval:
 """`entireinterval`s represent the whole Real line: [-∞, ∞]."""
-entireinterval(::Type{T}) where T<:Real = Interval{T}(-Inf, Inf)
-entireinterval(x::Interval{T}) where T<:Real = entireinterval(T)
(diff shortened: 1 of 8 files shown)


src/intervals/special.jl
```diff
@@ -1,41 +1,47 @@
-# This file is part of the IntervalArithmetic.jl package; MIT licensed
+# This file is part of the BaseIntervalArithmetic.jl package; MIT licensed
```
gwater: small typo:
```suggestion
# This file is part of the IntervalArithmetic.jl package; MIT licensed
```

src/intervals/wrapper_intervals.jl
```diff
@@ -0,0 +1,76 @@
...
+                          RealSetInterval{T},
+                          PseudoNumberInterval{T}}
+
+# Alias for convenience.
+# Default is to the most restrictive version.
+Interval{T} = SetInterval{T}
+
+disambiguation_mode(val) = NoDisambiguation
+disambiguation_mode(::Union{SetInterval, RealSetInterval}) = ErrorOnAmbiguous
+disambiguation_mode(::PseudoNumberInterval) = FalseOnAmbiguous
```
gwater: I still think the intervals should behave more like:
```suggestion
disambiguation_mode(::Union{SetInterval, RealSetInterval}) = IEEEConform
disambiguation_mode(::PseudoNumberInterval) = ErrorOnAmbiguous
```
(Here `IEEEConform` is the current behavior.) It's not quite clear to me if `FalseOnAmbiguous` would add anything since one could always wrap an `ErrorOnAmbiguous` type in a `try-catch` statement.

Kolaru: The idea behind `FalseOnAmbiguous` was that it return `true` if and only if the tested predicate is guaranteed to be true, otherwise returning `false`. It turns out that for comparison, the IEEE standard (section 10.5.10) call the corresponding function `precedes` (and denotes it with a funky operator), while the `less` operator (denoted as `<` in the standard) represents something else.

Your are correct that this should simply refers to the standard and in the documentation it must be clear what we call `<`. It will be hard to be fully compliant with the notation of the standard in a meaningful way, because I don't know if it takes in account languages supporting unicode operators, nor the very Julian expectation of being able to throw new types into old algorithms.

Then I totally messed up the naming, the standard calling the described default interval as "set based", using the same convention we end up to your proposition (that I do not commit yet as `IEEEConform` is not defined).

gwater: > Your are correct that this should simply refers to the standard and in the documentation it must be clear what we call `<`. It will be hard to be fully compliant with the notation of the standard in a meaningful way, because I don't know if it takes in account languages supporting unicode operators, nor the very Julian expectation of being able to throw new types into old algorithms.

I don't think that is an issue considering IEEE section 12.2 "Naming conventions for operations" states that implementations can use 'generic text names' like `add(a, b)` instead of `a+b`. The standard only specifies functionality. (In fact I find it rather unfortunate that they specified set operations using operators normally used for number operations, such as `<`.)

Kolaru:
Superseeded by #271.

PR #258, opened 2019-02-23
Type stability for some functions

dpsanders:
Some functions were not type stable for `Interval{Float32}`. 

~This needs tests.~ Done

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -394,7 +394,7 @@ The default is the true midpoint at `α = 0.5`.
 
 Assumes 0 ≤ α ≤ 1.
 
-Warning: if the parameter `α = 0.5` is explicitely set, the behavior differs
+Warning: if the parameter `α = 0.5` is explicitly set, the behavior differs
 from the default case if the provided `Interval` is not finite, since when
 `α` is provided `mid` simply replaces `+∞` (respectively `-∞`) by `prevfloat(+∞)`
 (respecively `nextfloat(-∞)`) for the computation of the intermediate point.
@@ -406,13 +406,15 @@ function mid(a::Interval{T}, α) where T
     lo = (a.lo == -∞ ? nextfloat(-∞) : a.lo)
     hi = (a.hi == +∞ ? prevfloat(+∞) : a.hi)
 
-    midpoint = α * (hi - lo) + lo
+    β = convert(T, α)
+
+    midpoint = β * (hi - lo) + lo
     isfinite(midpoint) && return midpoint
     #= Fallback in case of overflow: hi - lo == +∞.
        This case can not be the default one as it does not pass several
        IEEE1788-2015 tests for small floats.
     =#
-    return (1-α) * lo + α * hi
+    return (1 - β) * lo + β * hi
 end
 
 """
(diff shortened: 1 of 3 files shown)


dpsanders:
@lbenet This is ready.

test/interval_tests/consistency.jl
```diff
@@ -388,4 +388,20 @@ setprecision(Interval, Float64)
...
     end
 
+    @testset "Type stability" begin
+        for T in (Float32, Float64, BigFloat)
+
+            x = Interval{T}(3, 4)
+            y = Interval{T}(5, 6)
+
+            for op in (+, -, *, /, atan)
+                @inferred op(a, b)
```
Kolaru:
```suggestion
                @inferred op(x, y)
```

test/interval_tests/consistency.jl
```diff
@@ -388,4 +388,20 @@ setprecision(Interval, Float64)
...
+            end
+
+            for op in (sin, cos, exp, log, tan, abs, mid, diam)
+                @inferred op(a)
```
Kolaru:
```suggestion
                @inferred op(x)
```

You probably want to use the intervals you define above.

src/intervals/arithmetic.jl
```diff
@@ -162,8 +162,8 @@ function inv(a::Interval{T}) where T<:Real
     isempty(a) && return emptyinterval(a)
 
     if zero(T) ∈ a
-        a.lo < zero(T) == a.hi && return @round(-Inf, inv(a.lo))
-        a.lo == zero(T) < a.hi && return @round(inv(a.hi), Inf)
+        a.lo < zero(T) == a.hi && return @round(T(-Inf), inv(a.lo))
+        a.lo == zero(T) < a.hi && return @round(inv(a.hi), T(Inf))
```
Kolaru: Since `T(Inf)` is used quite a bit, it may be worth to define `infinity(T)` to work the same way `zero(T)` does.

test/interval_tests/consistency.jl
```diff
@@ -388,4 +388,20 @@ setprecision(Interval, Float64)
 
     end
 
+    @testset "Type stability" begin
```
lbenet: Some observations on the tests:
- Line: 398: `@inferred op(a, b)`; I think it should read `@inferred op(x, y)`.
- I think there should be tests involving `/` with intervals having `zero(T)` in one of the edges, since this PR changes precisely that function. The same holds for `inv` and `extended_div`.
- Similarly, I think there should be some tests for `mid`, which is also affected by this PR

src/intervals/arithmetic.jl
lbenet: I fully agree (though maybe in another PR).

dpsanders:
Thanks for the comments and catching the stupid typos! Will change accordingly.

dpsanders:
There used to be a function `inf(T)` in Base that gave exactly the infinity of type T, but it got removed at some point. I agree that it probably makes sense for us to use one here.

lbenet:
@Kolaru Do you agree to merge this PR as it is now?

Kolaru:
Yes, and I can even merge myself now =D

PR #260, opened 2019-03-05
Add rand(X) for Intervals and IntervalBoxes

dpsanders:


dpsanders:
@lbenet This is ready.

dpsanders:
Suppose you want to generate several random points inside a box and evaluate a function there, e.g. for optimization. To get an idea of how the function behaves, you want to evaluate the function as quickly as possible on each point. Once you have selected the best candidate point, you only then enclose it in a box and do it rigorously, since evaluating complicated functions on a box seems to be at least an order of magnitude slower.

lbenet:
Thanks for the explanation! Merging

PR #262, opened 2019-03-05
Add missing set operations involving `IntervalBox`es, with tests

### LINKED ISSUE #261: Add missing set operations to IntervalBox
```julia

julia> A = IntervalBox(1..2, 3..4)
[1, 2] × [3, 4]

julia> B = IntervalBox(0..2, 3..6)
[0, 2] × [3, 6]

julia> A ⊂ B
ERROR: MethodError: no method matching ⊂(::IntervalBox{2,Float64}, ::IntervalBox{2,Float64})
Stacktrace:
 [1] top-level scope at none:0

julia> B ⊃ A
ERROR: MethodError: no method matching ⊃(::IntervalBox{2,Float64}, ::IntervalBox{2,Float64})
Stacktrace:
 [1] top-level scope at none:0
```

lbenet:
Close #261

### DIFF
--- a/src/multidim/arithmetic.jl
+++ b/src/multidim/arithmetic.jl
@@ -31,11 +31,8 @@ Base.size(X::IntervalBox{2,Float64}) = (2,)
 @inline broadcasted(f, x, y, Z::IntervalBox) = wrap(f.(x, y, Z.v))
 @inline broadcasted(f, x, Y::IntervalBox, z) = wrap(f.(x, Y.v, z))
 
-for op in (:+, :-, :∩, :∪, :⊆, :isinterior, :dot, :setdiff, :×)
+for op in (:+, :-, :∩, :∪, :⊆, :⊂, :⊃, :isinterior, :dot, :setdiff, :×)
     @eval $(op)(a::SVector, b::IntervalBox) = $(op)(IntervalBox(a), b)
-end
-
-for op in (:+, :-, :∩, :∪, :⊆, :isinterior, :dot, :setdiff, :×)
     @eval $(op)(a::IntervalBox, b::SVector) = $(op)(a, IntervalBox(b))
 end
 
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -65,8 +65,9 @@ big(X::IntervalBox) = big.(X)
 ## set operations
 
 # TODO: Update to use generator
-⊆(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} =
-    all(X.v .⊆ Y.v)
+⊆(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊆ Y)
+⊂(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊂ Y)
+⊃(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊃ Y)
 
 ∩(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} =
     IntervalBox(X.v .∩ Y.v)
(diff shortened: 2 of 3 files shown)


lbenet:
Restarted travis, since there was some problem with one test...

lbenet:
Tests pass. I think this is ready for review.

src/multidim/intervalbox.jl
```diff
@@ -65,8 +65,9 @@ big(X::IntervalBox) = big.(X)
 ## set operations
 
 # TODO: Update to use generator
-⊆(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} =
-    all(X.v .⊆ Y.v)
+⊆(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊆ Y)
+⊂(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊂ Y)
+⊃(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .⊃ Y)
```
Kolaru: Since all definitions are the same you could use something like (untested):

```jl
for op in (:⊆, :⊂, :⊃)
    @eval $(op)(X::IntervalBox{N,T}, Y::IntervalBox{N,T}) where {N,T} = all(X .$op Y)
end
```

test/multidim_tests/multidim.jl
```diff
@@ -37,7 +37,16 @@ let X, A  # avoid problems with global variables
...
+    @test A ⊂ B.v
+    @test A ⊂ B
+    @test A.v ⊂ B
     @test A ⊆ B.v
+    @test B ⊇ A
+    @test B.v ⊇ A
+    @test B ⊇ A.v
+    @test B ⊃ A
+    @test B.v ⊃ A
+    @test B ⊃ A.v
```
Kolaru: Adding negative tests like

```jl
@test !(B ⊂ A)
```

would be good since currently the function 

```jl
⊂(A, B) = true
```
would pass the tests.

Also there may be corner cases worth testing, like when both `IntervalBox` are the same.

lbenet: Thanks for the suggestions! I have implemented new tests as suggested, including cases where both the function acts on the same `IntervalBox`.

src/multidim/intervalbox.jl
lbenet: Done!

lbenet:
Thanks @Kolaru for the review; I've followed all your recommendations, though the actual way I implemented the meta programming part is slightly different than suggested.

dpsanders:
@lbenet Is this ready?

src/multidim/arithmetic.jl
```diff
@@ -31,11 +31,8 @@ Base.size(X::IntervalBox{2,Float64}) = (2,)
 @inline broadcasted(f, x, y, Z::IntervalBox) = wrap(f.(x, y, Z.v))
 @inline broadcasted(f, x, Y::IntervalBox, z) = wrap(f.(x, Y.v, z))
 
-for op in (:+, :-, :∩, :∪, :⊆, :isinterior, :dot, :setdiff, :×)
+for op in (:+, :-, :∩, :∪, :⊆, :⊂, :⊃, :isinterior, :dot, :setdiff, :×)
```
dpsanders: What about `⊇`?

lbenet: It just works (Julia magic); see the tests. ;-)

lbenet:
Sorry for the delay to respond; yes, I think this is ready.

lbenet:
@dpsanders @Kolaru Any comments?

Kolaru:
Looks good to me.

dpsanders:
LGTM, thanks a lot!

lbenet:
Thanks a lot!

PR #273, opened 2019-03-27
Fix copysign

dlfivefifty:
`copysign` was not behaving correctly, so this overloads it.

Should I also have `signbit` return an interval? Right now it returns:
```julia
julia> signbit(@interval(-1,1))
false
```
which caused the error in `copysign` without warning. Arguably it could return `@interval(false,true)` and then an error would be thrown by the default `copysign`:
```julia
copysign(x::Real, y::Real) = ifelse(signbit(x)!=signbit(y), -x, +x)
```

Unfortunately, it's impossible to overload `ifelse`:
```julia
julia> function Base.ifelse(a::Interval, x, y)
       false in a && true in a && return x ∪ y
       false in a && return x
       true in a && return y
       throw(TypeError("non-boolean used in boolean contex"))
       end
ERROR: cannot add methods to a builtin function
Stacktrace:
 [1] top-level scope at none:0
```


dlfivefifty:
This is ready for review.

dpsanders:
Thanks and sorry for the delay in getting to this. I'm getting
```
julia> copysign(-1..2, -1..2)
[-2, 2]
```
which seems wrong.

dpsanders:
This is (presumably) because `sign` of an `Interval` is an interval function that gives
```
julia> sign(-1..2)
[-1, 1]
```

dpsanders:
I'm not sure if it really makes sense to define `signbit` on an `Interval`. Maybe it should just give an error?

dlfivefifty:
> Thanks and sorry for the delay in getting to this. I'm getting
> 
> ```
> julia> copysign(-1..2, -1..2)
> [-2, 2]
> ```
> which seems wrong.

I don't follow, `[-2,2]` seems right to me:
```julia
julia> copysign(2,-1), copysign(2,2)
(-2, 2)
```

> I'm not sure if it really makes sense to define `signbit`

"bit" isn't literal here:
```julia
help?> signbit
search: signbit

  signbit(x)

  Returns true if the value of the sign of x is negative, otherwise false.

julia> signbit(BigFloat(2)) # It works for BigFloats even
false
```
So the right thing is for `signbit(1..2)` to return `false..false`, `signbit(-2..(-1))` to return `true..true`, and `signbit(-1..1)` to return `false..true`.

dpsanders:
I guess this depends how the docs are interpreted.
I interpret `copysign` as returning an object that is identical to `x` except that its sign might have changed.

The following difference in particular seems wrong:
```
julia> copysign(-1..2, 2)
[-1, 2]

julia> copysign(-1..2, 2..2)
[0, 2]
```

dlfivefifty:
Interpreting it that way would make essentially all usages of `copysign` do the wrong thing. In particular, I need it for Householder reflections in QR. Any other behaviour would just lead to inelegant code.

```julia
julia> copysign(-1..2, 2)
[-1, 2]
```
is a bug caused by using the default definition of `copysign`, which would throw an error if `signbit` was overloaded.

dlfivefifty:
I pushed the proposed behaviour of `signbit`. I view `signbit` as another variant of `sign` and `angle`, and so this change makes this consistent. I added docs for the `Interval` case to emphasise that it does not return a `Bool`.

I found one cases where Base will throw an arguably unexpected error:
```julia
signequal(x, y) = signbit(x)::Bool == signbit(y)::Bool
signless(x, y) = signbit(x)::Bool & !signbit(y)::Bool
```
But an error would be thrown anyways if `signbit` always throws an error.

dpsanders:
Tests are failing on appveyor, apparently due to using `@interval(true, true) `. Note that boolean intervals are not defined in the package.

dpsanders:
On windows a different path is used that passes through a `parse`.

To solve this just replace true by 1 and false by 0.

dpsanders:
LGTM, thanks a lot!

PR #279, opened 2019-04-27
Update docs deployment

lbenet:


lbenet:
Not important at this point because it is related with the nightly version, but worth keeping in mind since it signals a possible type unstability: https://travis-ci.org/JuliaIntervals/IntervalArithmetic.jl/jobs/525408976#L547

lbenet:
It seems that the docs generation went ok... Let's wait for all tests to complete.

lbenet:
@dpsanders I think this is ready.

dpsanders:
Thanks!

PR #299, opened 2019-07-04
Add missing dependencies to Project.toml and remove REQUIRE

lbenet:
Specifically, `LinearAlgebra` and `Markdown` were missing. Note that the version of the package was also missing so I fixed it to 0.15.2 (the current one), until we want a new patch/minor release.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,23 +1,30 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
-
-[compat]
-CRlibm = "≥ 0.7.0"
-StaticArrays = "≥ 0.8.0"
-FastRounding = "≥ 0.1.2"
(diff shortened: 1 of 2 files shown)


lbenet:
@dpsanders This is ready to be merged. Shall I go ahead?

lbenet:
In case we want to make a new release, it is probably a good idea to change the version number to 0.15.3 or 0.16.0, whatever is better.

dpsanders:
Yes we should probably make it 0.16. Go ahead and merge, thanks a lot!

lbenet:
Done. I'll wait for the tests to complete and then I'll merge.

lbenet:
Do you know if we need to open an issue for Registry, or can it be done after merging from this PR?

dpsanders:
It can be done from the commit.

lbenet:
Merging...

dpsanders:
Thanks a lot!

Issue #306, opened 2019-07-23
hypot() on intervals


bocc:
Apparently some of the magic that hypot() does to avoid overflows (and underflows?) interferes when you use intervals. For example,

hypot(-3..4, 1..4) results in [0..\infty] in Julia 1.1, on Windows.

It is not a problem per se (as the sqrt version works well), I just thought you might find it interesting. Also, the package is excellent, thank you!

dpsanders:
Confirmed:
```
julia> hypot(-3..4, 1..4)
[0, ∞]
```

This is pretty bad... Would you like to submit a PR with a trivial implementation of `hypot` for `Interval`s?

bocc:
Yes, I will try to do so, hopefully today, tomorrow the latest.

bocc:
So, I've created a PR, but didn't add a test case. I would suggest something along the lines of:
~~~
√2..5 ⊆ hypot(-3..(-1),1..4)
hypot(-3..(-1),1..4) ⊆ (√2-eps())..(5+eps())
~~~
as this returns `[1.05409, 12.3694]` with the default `hypot` implementation. Also, I wanted to mention that I didn't know how to approach the `Complex{Interval{T}}` case. In the code below:
~~~
using ValidatedNumerics

a = (@interval 0 2) + (@interval 1 5)*im
b = (@interval 1 3) + (@interval 2 4)*im

# a real interval
hypot(a,b)

# a complex interval
sqrt(a^2 + b^2)

# with and without rotation
a^2 + b^2
(sqrt(a^2 + b^2))^2
~~~
`hypot` (from `Base`) returns a Real interval, while `sqrt` returns a Complex one. Also, the two consecutive scaling and rotation in the last part broadens the resulting interval a bit.

lbenet:
Thanks for the contribution; for reference, it is #308 

I do not know exactly where the tests should be coded. Yet, I'd like to comment in your observation that `sqrt` and `hypot` behave different for complex numbers. I think the reason is relaated to `^2` (that you use in `sqrt` above) instead of `abs2`; if you redo your example as `sqrt(abs2(a)+abs2(b))` you get the same as `hypot(a,b)`.

gwater:
Note that the issue comes from a binary comparison (`ax < ay`) in `Base.hypot()` which also explains why some Intervals work fine:
```julia
julia> a = Interval(0.2, 0.5);
julia> b = Interval(1, 4);
[...]

PR #308, opened 2019-07-26
hypot() defined for intervals (#306)

bocc:
Hello,

I have added the trivial implementation and the necessary import and export for hypot to be usable. I didn't add a doc comment or a test case (as I didn't know where you would want to put that).


dpsanders:
Thanks for the PR! 

I agree that the tests are not very well structured right now. Maybe `numeric.jl` is a good place for the tests.

PR #320, opened 2019-08-13
Fix convert bug

dpsanders:


src/intervals/conversion.jl
```diff
@@ -17,7 +17,7 @@ convert(::Type{Interval{T}}, x::Bool) where {T} = convert(Interval{T}, Int(x))
 convert(::Type{Interval{T}}, x::Real) where {T} = atomic(Interval{T}, x)
 convert(::Type{Interval{T}}, x::T) where {T<:Real} = Interval{T}(x)
 convert(::Type{Interval{T}}, x::Interval{T}) where {T} = x
-convert(::Type{Interval{T}}, x::Interval) where {T} = atomic(Interval{T}, x)
+convert(::Type{Interval{T}}, x::Interval{S}) where {T,S} = atomic(Interval{T}, x)
 
 convert(::Type{Interval}, x::Real) = (T = typeof(float(x)); convert(Interval{T}, x))
 convert(::Type{Interval}, x::Interval) = x
```
Kolaru: Not directly related to the issue, but this definition is the same as the one on line 20 and should probably be deleted (especially since the result would not be correct if for some weird reason it ever gets called).

dpsanders: Sorry, I don't understand which definition you're talking about that is the same as the one on line 20?

src/intervals/conversion.jl
```diff
@@ -17,7 +17,7 @@ convert(::Type{Interval{T}}, x::Bool) where {T} = convert(Interval{T}, Int(x))
 convert(::Type{Interval{T}}, x::Real) where {T} = atomic(Interval{T}, x)
 convert(::Type{Interval{T}}, x::T) where {T<:Real} = Interval{T}(x)
 convert(::Type{Interval{T}}, x::Interval{T}) where {T} = x
```
schillic:
```suggestion
convert(::Type{Interval{T}}, x::Interval{T}) where {T<:Real} = x
```

src/intervals/conversion.jl
```diff
@@ -17,7 +17,7 @@ convert(::Type{Interval{T}}, x::Bool) where {T} = convert(Interval{T}, Int(x))
...
-convert(::Type{Interval{T}}, x::Interval) where {T} = atomic(Interval{T}, x)
+convert(::Type{Interval{T}}, x::Interval{S}) where {T,S} = atomic(Interval{T}, x)
```
schillic:
```suggestion
convert(::Type{Interval{T}}, x::Interval) where {T<:Real} = atomic(Interval{T}, x)
```

test/interval_tests/construction.jl
```diff
@@ -340,3 +342,16 @@ import IntervalArithmetic: force_interval
...
+@testset "#320 fixing convert bug" begin
+    x = 3..4
+    T = typeof(x)
+
+    m = @which convert(T, x)
+
+    @test m.sig == Tuple{typeof(convert),Type{Interval{T}},Interval{T}} where T
+
+    x = interval(1//1, 5//3)
+    @test convert(typeof(x), x) == x
```
schillic:
```suggestion
    @test convert(typeof(x), x) === x
```
This would check that it does not create a new interval.

src/intervals/conversion.jl
Kolaru: On line 23:

```jl
convert(::Type{Interval}, x::Interval) = x
```

Since `Interval` is the same as `Interval{T} where {T}` and nothing there requests both parameter to be the same.

src/intervals/conversion.jl
Kolaru: While I see your point, this would make the code harder to maintain if we change the parameter type constrain.

schillic: You can add a constant and use that instead of `Real`.

```julia
const IT = Real  # type constraint of intervals
```

lbenet:
Tests are not passing (in Julia >= 1.2) due to `InteractiveUtils.jl`; it seems to be somehow required but we don't have in Project.toml.

Aside from this, I also interpreted Jeff's remark in JuliaLang/julia#32884 as restricting the parameter `T`, as @schillic suggests.

Kolaru:
I see your point and wouldn't mind if it wasn't for #185: rewriting the type constrain here prevent the use of traits as a custom type allowed by a user would fail here.

Issue #336, opened 2019-11-28
Problems with trigonometric functions for Float32 intervals


mfherbst:
I'm interested in using IntervalArithmetic with Float32 data types. At the moment I'm not that much interested in speed, more to get it to work.

There are currently some issues with some functions for Float32 intervals. For example 
```julia
cos(Interval{Float32}(0.1))
ERROR: MethodError: no method matching pi_interval(::Type{Float32})
Closest candidates are:
  pi_interval(::Type{BigFloat}) at /home/herbstm/.julia/packages/IntervalArithmetic/1T5or/src/intervals/precision.jl:74
  pi_interval(::Type{Float64}) at /home/herbstm/.julia/packages/IntervalArithmetic/1T5or/src/intervals/precision.jl:75
Stacktrace:
 [1] two_pi(::Type{Float32}) at /home/herbstm/.julia/packages/IntervalArithmetic/1T5or/src/intervals/trigonometric.jl:18
 [2] cos(::Interval{Float32}) at /home/herbstm/.julia/packages/IntervalArithmetic/1T5or/src/intervals/trigonometric.jl:151
 [3] top-level scope at REPL[10]:1
```

Similarly
```
exp(Interval{Float32}(0.1))
ERROR: StackOverflowError:
Stacktrace:
 [1] exp(::Float32, ::RoundingMode{:Down}) at /home/herbstm/.julia/packages/CRlibm/2S9DB/src/CRlibm.jl:137 (repeats 80000 times)
```
which I have also reported here: https://github.com/JuliaIntervals/CRlibm.jl/issues/32. There might be more things, but this is what I found so far.

dpsanders:
This is now working with version 0.8 of `CRlibm`.
No changes were required to `IntervalArithmetic`!

```
julia> exp(Interval{Float32}(0.1))
[1.10517, 1.10518]

julia> cos(Interval{Float32}(0.1))
[0.995004, 0.995005]

julia> sin(Interval{Float32}(0.1))
[0.0998334, 0.0998335]

julia> log(Interval{Float32}(0.1))
[-2.30259, -2.30258]
```

dpsanders:
We may want to change `show` to show that these are intervals of Float32.

PR #271, opened 2019-03-26
Implement systems for flavor, pointwise extensions, default bound and rounding mode

Kolaru:
This PR is my latest take on solving #237 and is a replacement #254. The main reason for a new PR and the closing of the previous one is that #254 relied heavily on macro and was thus hard to maintain and even harder to extend. Moreover the discussion around #254 arrived at the conclusion that what we want is really what the standard call a "flavor", and thus trying to come up with a flavor system seems the way to go.

In summary my first try in #254 was both misguided and overcomplicated.

The new trick is simpler and is essentially contained in the following lines:

```jl
abstract type AbstractRealFlavor{T} <: Real end
abstract type AbstractNonRealFlavor{T} end

const AbstractFlavor{T} = Union{AbstractRealFlavor{T}, AbstractNonRealFlavor{T}}
```

Concrete interval types must subtype either `AbstractRealFlavor` or `AbstractNonRealFlavor` and will benefit from all methods defined for `AbstractFlavor`. Thus this system should be extensible and relatively easy to maintain as for the most part we can treat `AbstractFlavor` as the common supertype of all intervals (it is however not as it can not be directly subtyped).

Currently this is still a work in progress, but the package compile, thanks to the definition of a default interval type named `Interval`. Thus most old definitions work but are only defined for the default flavor (note that some functionnality like the `..` syntax will always only be defined for the default interval type).

What remain to be done is

- [ ] Separate the functionnalities common for all flavors from the flavor dependant ones and redefined them accordingly.
- [ ] Implement all required functions for the IEEE standard "set based flavor".
- [ ] For each function in the IEEE standard, add ref. to where it is defined in the IEEE standard.
- [ ] Add all relevant flavors.
- [ ] Fix all the bugs.

I do not forget the problem of boolean functions that were at the core of #254 and solutions to these problems to them will be implemented in the corresponding flavors.

PS: Since this requires to review more or less the whole codebase I'm also trying to clean up the code along the way.

List of issues that should be fixed by this PR or that this PR will help solve: 
- [ ] #237 Core issue tackled here.
- [ ] #264 Division by 0 will be handled differently depending on the flavor used.
- [ ] #272 Kaucher arithmetic could be one of the supported flavors.
- [ ] #168 Behavior of `isinf()` will be flavor dependant.
- [ ] #165 Boolean operations will be flavor dependant.
- [ ] #2 Oldest issue about interval being subtype of `Real`.
[...]

PR #337, opened 2019-11-28
Add pi_interval for Float32 and Float16

mfherbst:
Partially deals with #336

### DIFF
--- a/src/intervals/precision.jl
+++ b/src/intervals/precision.jl
@@ -69,10 +69,14 @@ setprecision(::Type{Interval}, t::Tuple) = setprecision(Interval, t...)
 precision(::Type{Interval}) = (parameters.precision_type, parameters.precision)
 
 
-const float_interval_pi = atomic(Interval{Float64}, pi)  # does not change
+const float64_interval_pi = atomic(Interval{Float64}, pi)  # does not change
+const float32_interval_pi = atomic(Interval{Float32}, pi)
+const float16_interval_pi = atomic(Interval{Float16}, pi)
 
 pi_interval(::Type{BigFloat}) = parameters.pi
-pi_interval(::Type{Float64})  = float_interval_pi
+pi_interval(::Type{Float64})  = float64_interval_pi
+pi_interval(::Type{Float32})  = float32_interval_pi
+pi_interval(::Type{Float16})  = float16_interval_pi
 
 
 function Base.setrounding(f::Function, ::Type{Rational{T}},


Kolaru:
I haven't tested it for performance much yet, but in #271 I planned to achieve that through

```julia
const Interval{T}(::Irrational{:π}) where T = atomic(Interval{T}, π)
const Interval(::Irrational{:π}) = atomic(Interval{Float64}, π)
```
With that you are able to get `pi_interval` for any type using regular constructor, for example `Interval{Float32}(π)`. Also this would allow to get rid of all `pi_interval_*` constant, but also requires more changes than your proposition.

If you think it fits your needs, I can look into having this as a separate PR.

dpsanders:
Do the const actually do anything there? The idea is definitely not to be recomputing this each time.

Kolaru:
I don't remember if the `const` is really needed (I'll test that), but the result should be computed only once when the function is compiled anyway, since the constructor do not depends on the value of the argument (and it works since the irrational `pi` is the only object that has type `Irrational{:π}`).

mfherbst:
@Kolaru My main point for this PR was achieving a fix for #336, so if you propose a PR, which introduces a working set of trigonometric functions for `Float32` intervals, then I'm happy, too :smile:.

Actually using `Irrational` seems more like idiomatic Julia to me and less like the quick fix I proposed, so I'd surely appreciate if you could have a more polished go at it.

dpsanders:
@Kolaru That would be great if it works, but it's not clear to me that it will. (And no time to check right now.)

Kolaru:
```julia
julia> @btime Interval{Float16}(π)
  2.299 ns (0 allocations: 0 bytes)
[3.14062, 3.14258]

julia> @btime Interval{Float32}(π)
  0.001 ns (0 allocations: 0 bytes)
[3.14159, 3.1416]

julia> @btime Interval{Float64}(π)
  0.001 ns (0 allocations: 0 bytes)
[3.14159, 3.1416]
```

Somehow it looks like it works fine for `Float32` and `Float64`, but not `Float16`, as confirmed by the following:


```julia
julia> @code_warntype Interval{Float16}(π)
Body::Interval{Float16}
1 ─ %1 = invoke IntervalArithmetic.atomic(Interval{Float16}::Type{Interval{Float16}}, IntervalArithmetic.π::Irrational{:π})::Interval{Float16}
└──      return %1

julia> @code_warntype Interval{Float32}(π)
Body::Interval{Float32}
1 ─     return [3.14159, 3.1416]

julia> @code_warntype Interval{Float64}(π)
Body::Interval{Float64}
1 ─     return [3.14159, 3.1416]
```

That's very mysterious to me, especially since the definition of both the constructor and of the `atomic` function are the same for `Float32` and `Float16`.

Kolaru:
The problem comes from the fact julia use specific mechanism to convert `Irrational` to `Float64` or `Float32`, probably to exploit some hardware support. We can dodge the problem using generated functions (a feature I've always wanted to use anyway =D):

```julia
@generated function Interval{T}(x::Irrational) where T
    res = atomic(Interval{T}, x())
    return :($res)
end
```

This works as expected, precomputing the value the first time the function is called and subsequently reusing it. I'll thus be able to come with a PR to superseed the current one.

lbenet:
Can you clarify a bit more why it works with generated functions?

Kolaru:
Generated functions work by allowing you to compute the body of your functions from the types that are passed to it. In our case that allow us to to the following:

1. Since each irrational has its own type, we can get the number by instantiating the type, e.g. we can get `π` by doing `Irrational{:π}()`. In the example it's what `x()` does, getting the value from the irrational type.
2. We compute the atomic interval around the irrational for the type we need (the `res` variable).
3. We define the body of the function to only consist of returning that value. It could be made a bit more explicit without omitting the return statement `:(return $res)`.

So for each type, this define the function to simply return a constant.

lbenet:
Thanks a lot for the explanation!

Kolaru:
The test suite is currently testing for thing like

```jl
Interval{Float64}(π) == Interval{Float64}(float(π))
```

which fails with my proposed definition. Can I go forward and delete this kind if tests or is there a reason to have it that way ?

dpsanders:
Yes please go ahead and change whatever you need to. Thanks!

Kolaru:
Superseeded by #338. Thanks you very much for your input on the matter @mfherbst !

mfherbst:
You're welcome. I'm happy to have started the discussion ;).

PR #338, opened 2019-12-06
Better constructors for irrationals

Kolaru:
This superseeds PR #337 by implementing a new constructor for irrational, in such a way no overhead are produced.

### DIFF
--- a/src/intervals/trigonometric.jl
+++ b/src/intervals/trigonometric.jl
@@ -10,12 +10,11 @@ For efficiency, does not check that the constant is positive.
 """
 multiply_by_positive_constant(α, x::Interval) = @round(α*x.lo, α*x.hi)
 
-half_pi(::Type{Float64}) = multiply_by_positive_constant(0.5, Interval{Float64}(π))
-half_pi(::Type{T}) where {T} = 0.5 * Interval{T}(π)
-half_pi(x::T) where {T<:AbstractFloat} = half_pi(T)
+half_pi(::Type{T}) where {T} = multiply_by_positive_constant(convert(T, 0.5), Interval{T}(π))
+half_pi(::T) where T = half_pi(T)
 
-two_pi(::Type{Float64}) = multiply_by_positive_constant(2.0, Interval{Float64}(π))
-two_pi(::Type{T}) where {T} = 2 * Interval{T}(π)
+two_pi(::Type{T}) where {T} = multiply_by_positive_constant(convert(T, 2.0), Interval{T}(π))
+two_pi(::T) where T = two_pi(T)
 
 range_atan(::Type{T}) where {T<:Real} = Interval(-(Interval{T}(π).hi), Interval{T}(π).hi)
 half_range_atan(::Type{T}) where {T} = (temp = half_pi(T); Interval(-(temp.hi), temp.hi) )
--- a/src/intervals/trigonometric.jl
+++ b/src/intervals/trigonometric.jl
@@ -2,7 +2,7 @@
 
 
 # hard code this for efficiency:
-const one_over_half_pi_interval = inv(0.5 * pi_interval(Float64))
+const one_over_half_pi_interval = inv(0.5 * Interval{Float64}(π))
 
 """
 Multiply an interval by a positive constant.
@@ -10,16 +10,16 @@ For efficiency, does not check that the constant is positive.
 """
 multiply_by_positive_constant(α, x::Interval) = @round(α*x.lo, α*x.hi)
 
-half_pi(::Type{Float64}) = multiply_by_positive_constant(0.5, pi_interval(Float64))
-half_pi(::Type{T}) where {T} = 0.5 * pi_interval(T)
+half_pi(::Type{Float64}) = multiply_by_positive_constant(0.5, Interval{Float64}(π))
+half_pi(::Type{T}) where {T} = 0.5 * Interval{T}(π)
 half_pi(x::T) where {T<:AbstractFloat} = half_pi(T)
 
-two_pi(::Type{Float64}) = multiply_by_positive_constant(2.0, pi_interval(Float64))
-two_pi(::Type{T}) where {T} = 2 * pi_interval(T)
+two_pi(::Type{Float64}) = multiply_by_positive_constant(2.0, Interval{Float64}(π))
+two_pi(::Type{T}) where {T} = 2 * Interval{T}(π)
 
-range_atan(::Type{T}) where {T<:Real} = Interval(-(pi_interval(T).hi), pi_interval(T).hi)
+range_atan(::Type{T}) where {T<:Real} = Interval(-(Interval{T}(π).hi), Interval{T}(π).hi)
 half_range_atan(::Type{T}) where {T} = (temp = half_pi(T); Interval(-(temp.hi), temp.hi) )
-pos_range_atan(::Type{T}) where {T<:Real} = Interval(zero(T), pi_interval(T).hi)
+pos_range_atan(::Type{T}) where {T<:Real} = Interval(zero(T), Interval{T}(π).hi)
 
 const halfpi = pi / 2.0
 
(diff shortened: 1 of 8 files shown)


src/intervals/trigonometric.jl
```diff
@@ -2,24 +2,24 @@
...
 """
 Multiply an interval by a positive constant.
 For efficiency, does not check that the constant is positive.
 """
 multiply_by_positive_constant(α, x::Interval) = @round(α*x.lo, α*x.hi)
 
-half_pi(::Type{Float64}) = multiply_by_positive_constant(0.5, pi_interval(Float64))
-half_pi(::Type{T}) where {T} = 0.5 * pi_interval(T)
+half_pi(::Type{Float64}) = multiply_by_positive_constant(0.5, Interval{Float64}(π))
+half_pi(::Type{T}) where {T} = 0.5 * Interval{T}(π)
```
lbenet: Naive question: would it help here to use `multiply_by_positive_constant`? (Same question to other lines below)

test/display_tests/display.jl
```diff
@@ -12,15 +12,13 @@ setprecision(Interval, Float64)
         a = 1..2
         b = -1.1..1.3
         c = Interval(pi)
-        d = @interval(π)
```
lbenet: Is there any difference in using `Interval(pi)` or `@interval(π)`? If not, why not simply test `Interval(pi) == @interval(π)`?

test/interval_tests/construction.jl
```diff
@@ -26,24 +26,35 @@ const eeuler = Base.MathConstants.e
...
-    @test pi..big(4) == hull(pi_interval(BigFloat), interval(4))
-    @test pi..pi == pi_interval(Float64)
-    @test eeuler..pi == hull(@interval(eeuler), pi_interval(Float64))
+    @test pi..big(4) == hull(Interval{BigFloat}(π), interval(4))
+    @test pi..pi == Interval{Float64}(π)
+    @test eeuler..pi == hull(@interval(eeuler), Interval{Float64}(π))
+    @test Interval(π) == @interval(π)
+    @test Interval(π, π) == Interval(π)
+    @test Interval{Float32}(π, π) == Interval{Float32}(π)
+
```
lbenet: I think it is worth adding similar tests (as the three precious lines) for `eeuler` here, since `typeof(eeuler)` returns `Irrational{:ℯ}`.

test/interval_tests/trig.jl
```diff
@@ -91,15 +91,15 @@ end
...
+        Interval(-Interval{Float64}(π).hi/4, Interval{Float64}(π).hi/4)
     @test atan(@interval(0)) == Interval(0.0, 0.0)
     @test atan(@biginterval(-1, 1)) ⊆ atan(@interval(-1, 1))
 end
 
 @testset "atan" begin
     @test atan(∅, entireinterval()) == ∅
     @test atan(entireinterval(), ∅) == ∅
-    @test atan(@interval(0.0, 1.0), @biginterval(0.0)) == @biginterval(pi/2)
+    @test_broken atan(@interval(0.0, 1.0), @biginterval(0.0)) == @biginterval(pi/2)
```
lbenet: Any clue why this test is broken now?

test/display_tests/display.jl
Kolaru: `@interval(π)` doesn't use the fancy new precomputed stuff, but should return the same value and `Interval(pi) == @interval(π)` is indeed tested below.

I'm not sure I can do the same trick I did on the constructor on the macro, but I'll have a look.

test/interval_tests/trig.jl
Kolaru: It is because float and bigfloats intervals are mixed, so the `Base.Math` implementation is used:

```julia
julia> @which atan(@interval(0, 1), @biginterval(0))
atan(y::Real, x::Real) in Base.Math at math.jl:639
```

The fact that the test wasn't previously broken is probably a coincidence.

dpsanders:
This looks like a great solution except for `BigFloat`, which now does not change with the precision:
```jl
julia> Interval{BigFloat}(pi)
Interval(3.141592653589793238462643383279502884197169399375105820974944592307816406286198, 3.141592653589793238462643383279502884197169399375105820974944592307816406286233)

julia> setprecision(Interval, 1000)
1000

julia> Interval{BigFloat}(pi)
Interval(3.141592653589793238462643383279502884197169399375105820974944592307816406286198, 3.141592653589793238462643383279502884197169399375105820974944592307816406286233)
```

dpsanders:
So I think BigFloat should just be special-cased.

test/display_tests/display.jl
Kolaru: I tried to make `atomic` use the `Interval` constructor, but then trying to use `IntervalArithmetic` crashed Julia... I don't know what's going on, so for now I'll let it as it is.

test/interval_tests/trig.jl
Kolaru: It was only part of the problem apparently, as the test is still broken. I don't understand why though, because the test passes if I run only a this single test file.

Kolaru:
I've addressed the reviews and discovered that MPFR doesn't support Euler constant as well as pi:

```julia
julia> eeuler..big(4)
ERROR: MethodError: no method matching BigFloat(::Irrational{:ℯ}, ::Base.MPFR.MPFRRoundingMode; precision=128)
Closest candidates are:
  BigFloat(::BigFloat, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:184
  BigFloat(::Int32, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:204
  BigFloat(::UInt32, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:204
  ...
Stacktrace:
 [1] #BigFloat#21(::Int32, ::Type{BigFloat}, ::Irrational{:ℯ}, ::RoundingMode{:Down}) at .\mpfr.jl:260
 [2] BigFloat(::Irrational{:ℯ}, ::RoundingMode{:Down}) at .\mpfr.jl:260
 [3] atomic(::Type{Interval{BigFloat}}, ::Irrational{:ℯ}) at C:\Users\Kolaru\.julia\dev\IntervalArithmetic\src\intervals\conversion.jl:89
 [4] ..(::Irrational{:ℯ}, ::BigInt) at C:\Users\Kolaru\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:155
 [5] top-level scope at REPL[35]:1
```

lbenet:
> I've addressed the reviews and discovered that MPFR doesn't support Euler constant as well as pi

I am not sure if I understand what you mean here:
```julia
julia> eeuler = Base.MathConstants.e
ℯ = 2.7182818284590...

julia> big(eeuler)
2.718281828459045235360287471352662497757247093699959574966967627724076630353555

julia> setprecision(1000)
1000

julia> big(eeuler)
2.7182818284590452353602874713526624977572470936999595749669676277240766303535475945713821785251664274274663919320030599218174135966290435729003342952605956307381323286279434907632338298807531952510190115738341879307021540891499348841675092447614606680822648001684774118537423454424371075390777449920695
```

Maybe the way rounding works is somewhat broken...

test/display_tests/display.jl
lbenet: Sorry, I didn't notice that the tests were already there...

Kolaru:
Here is the problematic feature:

```julia
julia> BigFloat(pi, RoundDown)
3.141592653589793238462643383279502884195

julia> BigFloat(eeuler, RoundDown)
ERROR: MethodError: no method matching BigFloat(::Irrational{:ℯ}, ::Base.MPFR.MPFRRoundingMode; precision=128)
Closest candidates are:
  BigFloat(::BigFloat, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:184
  BigFloat(::Int32, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:204
  BigFloat(::UInt32, ::Base.MPFR.MPFRRoundingMode; precision) at mpfr.jl:204
  ...
Stacktrace:
 [1] #BigFloat#21(::Int32, ::Type{BigFloat}, ::Irrational{:ℯ}, ::RoundingMode{:Down}) at .\mpfr.jl:260
 [2] BigFloat(::Irrational{:ℯ}, ::RoundingMode{:Down}) at .\mpfr.jl:260
 [3] top-level scope at REPL[43]:1
```

lbenet:
Maybe we should open an issue in Julia...

Kolaru:
I reported it : https://github.com/JuliaLang/julia/issues/34103

Kolaru:
Also I had a look at the coveralls reported and the decrease seem to be artifacts only (again).

lbenet:
I am in favor of merging this PR. The problems with `ℯ` and `φ` depend on Julia itself and have been reported. 

@dpsanders Do you have any comments or suggestions?

dpsanders:
LGTM, thanks!

lbenet:
Great!. Merging...

PR #340, opened 2019-12-09
Remove extra type constraints

ericphanson:
Ref https://github.com/JuliaIntervals/IntervalOptimisation.jl/pull/45/files#r355474198. As far as I can tell, they are unnecessary. They disallow comparing IntervalBoxes of different numeric types, but it just dispatches to the 1D versions which can handle different types, e.g.

```julia
julia> x1 = -1.0..2.0
[-1, 2]

julia> x2 = big(1.0)..big(3.0)
[1, 3]₂₅₆

julia> x1 ∩ x2
[1, 2]₂₅₆

julia> x1 ∪ x2
[-1, 3]₂₅₆

julia> x1 ⊂ x2
false

julia> x1 ⊆ x2
false

julia> x1 ⊃ x2
false

julia> x1 ⊆ x2
false
```

### DIFF
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -107,3 +107,5 @@ IntervalBox(x::Interval, ::Val{n}) where {n} = IntervalBox(SVector(ntuple( _ -> 
 IntervalBox(x::Interval, n::Int) = IntervalBox(x, Val(n))
 
 dot(x::IntervalBox, y::IntervalBox) = dot(x.v, y.v)
+
+Base.:(==)(x::IntervalBox, y::IntervalBox) = x.v == y.v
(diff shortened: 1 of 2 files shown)


dpsanders:
LGTM, thanks!

Could you please add a couple of tests of the new functionality?

ericphanson:
Sure, added. In the course of doing so, I realized we also need a method for `==`, otherwise the tests with `==` that I added fail. By default, `==` falls back to `===`. I think the reason that's OK for `IntervalBox{N, Float64}` is because everything is `isbits`. But that's not true for BigFloats, so we need to choose our `==` method to compare the static vectors.

ericphanson:
Let me know if there's another test I should add. I don't really understand the coveralls report, and I think each operation I changed is tested.

Kolaru:
Coveralls got slightly confused, but your tests are indeed sufficient. Thanks a lot!

PR #351, opened 2020-01-12
IntervalBox as subtype of AbstractArray

### LINKED ISSUE #350: Declare `IntervalBox` as subtype of `AbstractArray`?
Would it make sense to declare `IntervalBox <: AbstractVector`? Or alternatively `IntervalBox <: AbstractIntervalBox <: AbstractVector`

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/27cf9f47a9a1c4b5f566b96a1e0a5853a75ad68c/src/multidim/intervalbox.jl#L6

I ran into an error trying to evaluate a function-like object that accepts arrays as arguments. Example:

```
using DynamicPolynomials, IntervalArithmetic

@polyvar x y
box = IntervalBox(-1..1,2)

p = x^2 + 2y^2 + 3x*y
println(p([1,2]))   # 15
println(p(box))     # MethodError: no method matching (::Polynomial{true,Int64})(::IntervalBox{2,Float64})
```

ArjunNarayanan:
closes #350 

`IntervalBox` is now declared a sub-type of `AbstractArray` so that matrix-vector operations can be performed on `IntervalBox` objects.

Had to add an explicit method for multiplying `SMatrix` and `IntervalBox` to avoid method ambiguities with methods exported by `StaticArrays`.


mforets:
I guess this PR can be closed..? (answered [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/350)). (Also, conceptually it is not clear the  association of an interval box with a vector).

ArjunNarayanan:
Yes indeed, sorry for leaving this hanging!

PR #353, opened 2020-01-20
Return a concrete interval with negative-integer powers

timholy:
On v0.16.2 I get

```julia
julia> pow(-2 .. -1, -1 .. -1)
∅
```

With this I get
```julia
julia> pow(-2 .. -1, -1 .. -1)
[-1, -0.5]
```


lbenet:
Thanks @timholy for spotting and solving this. LGTM. Merging!

Issue #274, opened 2019-03-28
Export mince function


dpsanders:
```jl
function mince(X, n)
    r = range(X.lo, stop=X.hi, length=n+1)   # stop is required for Julia 1.0
    return [interval(r[i], r[i+1]) or i in 1:length(r)-1]
end
```

lbenet:
It would be great to add it !!!

lbenet:
Closed by #354

PR #354, opened 2020-01-31
Add `mince` for intervals and interval boxes

lbenet:
Solves #274

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -567,3 +567,15 @@ isinteger(a::Interval) = (a.lo == a.hi) && isinteger(a.lo)
 
 convert(::Type{Integer}, a::Interval) = isinteger(a) ?
         convert(Integer, a.lo) : throw(ArgumentError("Cannot convert $a to integer"))
+
+
+"""
+    mince(x::Interval, n)
+
+Splits `x` in `n` intervals of the same diameter, which are returned
+as a vector.
+"""
+function mince(x::Interval, n)
+    nodes = range(x.lo, stop = x.hi, length = n+1)
+    return [Interval(nodes[i], nodes[i+1]) for i in 1:length(nodes)-1]
+end
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -576,6 +576,6 @@ Splits `x` in `n` intervals of the same diameter, which are returned
 as a vector.
 """
 function mince(x::Interval, n)
-    nodes = range(x.lo, stop = x.hi, length = n+1)
+    nodes = range(x.lo, x.hi, length = n+1)
     return [Interval(nodes[i], nodes[i+1]) for i in 1:length(nodes)-1]
 end
--- a/src/intervals/set_operations.jl
+++ b/src/intervals/set_operations.jl
@@ -120,7 +120,7 @@ hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
 hull(a...) = reduce(hull, a)
-hull(a::Vector{Interval{T}}) where {T}= hull(a...)
+hull(a::Vector{Interval{T}}) where {T} = reduce(hull, a)
 
 """
     union(a, b)
@@ -133,7 +133,7 @@ union(a::Interval, b::Interval) = hull(a, b)
 #
 # union(a::Interval, b::Interval) = union(promote(a, b)...)
 union(a::Complex{<:Interval},b::Complex{<:Interval}) = hull(a, b)
-union(a, b, c...) = reduce(union, (a, b, c...))
+union(a...) = reduce(union, a)
 
 """
     setdiff(x::Interval, y::Interval)
--- a/src/intervals/set_operations.jl
+++ b/src/intervals/set_operations.jl
@@ -119,8 +119,8 @@ hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
-hull(a, b...) = reduce(hull, (a, b...))
-
+hull(a...) = reduce(hull, a)
+hull(a::Vector{Interval{T}}) where {T}= hull(a...)
 
 """
     union(a, b)
--- a/src/intervals/set_operations.jl
+++ b/src/intervals/set_operations.jl
@@ -119,6 +119,8 @@ hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
+hull(a, b...) = reduce(hull, (a, b...))
+
 
 """
     union(a, b)
@@ -131,7 +133,7 @@ union(a::Interval, b::Interval) = hull(a, b)
 #
 # union(a::Interval, b::Interval) = union(promote(a, b)...)
 union(a::Complex{<:Interval},b::Complex{<:Interval}) = hull(a, b)
-
+union(a, b, c...) = reduce(union, (a, b, c...))
 
 """
     setdiff(x::Interval, y::Interval)
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -135,11 +135,5 @@ as a vector.
     end
 end
 
-
-function IntervalArithmetic.hull(a::IntervalBox{N,T}, b::IntervalBox{N,T}) where {N,T}
-   v = Interval{T}[]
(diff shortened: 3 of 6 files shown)


src/intervals/arithmetic.jl
```diff
@@ -567,3 +567,15 @@ isinteger(a::Interval) = (a.lo == a.hi) && isinteger(a.lo)
...
+
+
+"""
+    mince(x::Interval, n)
+
+Splits `x` in `n` intervals of the same diameter, which are returned
+as a vector.
+"""
+function mince(x::Interval, n)
+    nodes = range(x.lo, stop = x.hi, length = n+1)
```
dpsanders: You don't need  `stop` I think for Julia >= 1.1?

src/multidim/intervalbox.jl
```diff
@@ -109,3 +109,28 @@ IntervalBox(x::Interval, n::Int) = IntervalBox(x, Val(n))
 dot(x::IntervalBox, y::IntervalBox) = dot(x.v, y.v)
 
 Base.:(==)(x::IntervalBox, y::IntervalBox) = x.v == y.v
+
+
+"""
+    mince(x::IntervalBox, n)
+
+Splits `x` in `n` intervals in each dimension of the same diameter. These
```
dpsanders: For higher dimensions this is a bad idea. Is this version actually useful?

src/multidim/intervalbox.jl
```diff
@@ -109,3 +109,28 @@ IntervalBox(x::Interval, n::Int) = IntervalBox(x, Val(n))
...
+"""
+@generated function mince(x::IntervalBox{N,T}, n) where {N,T}
+    quote
+        nodes_matrix = Array{Interval{T},2}(undef, n, N)
+        for i in 1:N
+            nodes_matrix[1:n,i] .= mince(x[i], n)
+        end
+
+        nodes = IntervalBox{$N,T}[]
+        Base.Cartesian.@nloops $N i _->(1:n) begin
```
dpsanders: I'm pretty sure there's an easier way to do this.

dpsanders: e.g. using something in https://julialang.org/blog/2016/02/iteration/ or in the `Iterators` module.

test/interval_tests/numeric.jl
```diff
@@ -340,3 +340,19 @@ end
 
 
 end
+
+@testset "Mince for `Interval`s" begin
+    v = mince(-1 .. 1, 4)
+    @test v == [-1 .. -0.5, -0.5 .. 0, 0 .. 0.5, 0.5 .. 1]
+    h = hull(hull(v[1], v[2]), hull(v[3], v[4]))
```
dpsanders: You can do `reduce(∪, v)`. We should make `hull(v)` for a vector `v` of intervals if this doesn't already exist.

src/intervals/arithmetic.jl
lbenet: Maybe not; I'll check it now, and if it is not needed I'll push a new commit with this correction.

src/multidim/intervalbox.jl
lbenet: I guess you mean that things grow as N^n; if so, I agree that this may not be ideal. 

I think this may be indeed useful: my motivation comes from the computation of tighter bounds for `TaylorModelN`'s when computing a validated integration. In some cases, e.g., when you have to check that certain region is not crossed this is possibly cheaper than reducing other parameters (e.g. integration step), in the sense of the *very many* operations (and memory) involved in taking the different stepsizes

src/multidim/intervalbox.jl
lbenet: I'll check the link; in any case, my naive checks yield a nice performance. This is also a good time to see if iteration is robust for `IntervalBox`-es.

test/interval_tests/numeric.jl
lbenet: I was implementing it as `hull(a, b, c...) = hull( hull(a,b), c...)`, which works. I'll check it with `reduce` as well, which may be more elegant.

lbenet: Shall I use this PR to address this? 

We should also implement `hull` for `IntervalBox`es, as the hull in each direction. Does this make any sense?

src/intervals/arithmetic.jl
lbenet: You were right; it is not needed. I just pushed a new commit which removes it.

test/interval_tests/numeric.jl
dpsanders: Yes please go ahead and do this in this PR. `hull` definitely makes sense for `IntervalBox`es yes.  Isn't `hull` just the same as `union`?

lbenet: In our package they are identical, since `union` is some synonym of `hull`.

Yet, I think the difference is that `union` is the set union of the intervals, while `hull(a,b)` returns an interval/intervalbox, which contains `a` and `b`.

lbenet: I'll push a commit which implements hull for intervals, and uses splatting for the hull of more than two intervals/interval-boxes.

dpsanders: I believe that `reduce` is more efficient.

lbenet: I've just pushed a new commit which addresses this.

src/intervals/set_operations.jl
```diff
@@ -119,6 +119,8 @@ hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
+hull(a, b...) = reduce(hull, (a, b...))
```
dpsanders: You can write this as 
```jl
hull(a...) = reduce(hull, a)
```

src/multidim/intervalbox.jl
```diff
@@ -134,3 +134,12 @@ as a vector.
         nodes
     end
 end
+
+
+function IntervalArithmetic.hull(a::IntervalBox{N,T}, b::IntervalBox{N,T}) where {N,T}
+   v = Interval{T}[]
+   for ind in 1:N
+       push!(v, hull(a[ind], b[ind]))
```
dpsanders: You don't want to be using an array here, since that will allocate and be inefficient.
Instead just use broadcasting, which is specialised for `SVector`s (on which `IntervalBox`es are based). 

Also note that this is already implemented as `union` of two `IntervalBox`es.

In fact, probably `hull` should just be an alias for `union`?

test/interval_tests/numeric.jl
```diff
@@ -344,15 +344,8 @@ end
 @testset "Mince for `Interval`s" begin
     v = mince(-1 .. 1, 4)
     @test v == [-1 .. -0.5, -0.5 .. 0, 0 .. 0.5, 0.5 .. 1]
-    h = hull(hull(v[1], v[2]), hull(v[3], v[4]))
-    @test h == -1 .. 1
+    @test hull(v...) == -1 .. 1
```
dpsanders: We can define `hull` for a `Vector` of intervals as well.

src/multidim/intervalbox.jl
lbenet: I've implemented something, which works, with `CartesianIndices`, but the performance is worst. I think we can leave like as it is for the time being.

src/intervals/set_operations.jl
lbenet: Thanks! I just pushed a new commit following your suggestion.

src/multidim/intervalbox.jl
lbenet: Good catch, thanks!. Done in the last commit.

test/interval_tests/numeric.jl
lbenet: Done in the last commit.

lbenet:
@dpsanders Any other comments on this PR?

src/intervals/set_operations.jl
```diff
@@ -119,6 +119,8 @@ hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
+hull(a...) = reduce(hull, a)
+hull(a::Vector{Interval{T}}) where {T}= hull(a...)
```
dpsanders: You can just use reduce directly here I think, rather than splatting the vector.

src/intervals/set_operations.jl
```diff
@@ -131,7 +133,7 @@ union(a::Interval, b::Interval) = hull(a, b)
 #
 # union(a::Interval, b::Interval) = union(promote(a, b)...)
 union(a::Complex{<:Interval},b::Complex{<:Interval}) = hull(a, b)
-
+union(a, b, c...) = reduce(union, (a, b, c...))
```
dpsanders: Is there a reason for not using  a... here?

dpsanders:
Just had a couple more comments but otherwise LGTM, thanks!

src/intervals/set_operations.jl
lbenet: Good catch! I guess I simply forgot this.

src/intervals/set_operations.jl
lbenet: No, no reason... I've changed it, and now it uses `reduce`.

lbenet:
I just pushed the changes in a new commit. Thanks for reviewing it again.

lbenet:
Merged...

Issue #349, opened 2020-01-01
Use random samplers correctly


dpsanders:
To allow e.g. `rand(X, 10)`. 

Ref: https://docs.julialang.org/en/v1/stdlib/Random/#Hooking-into-the-Random-API-1

krish8484:
#358

Issue #263, opened 2019-03-05
Change eltype(x::T) to return T


dpsanders:
The current definition of `eltype` is causing broadcasting to fail:
```
julia> using IntervalArithmetic

julia> x = 3..4
[3, 4]

julia> x .* [1]
ERROR: MethodError: no method matching Float64(::Interval{Float64})
```
For this to work, we need `eltype(x::T) where {T<:Interval} = T`, and similarly for `IntervalBox`.

As far as I can tell we don't use `eltype` anywhere, so this should be fine.

dlfivefifty:
This needs to be reopened:
```julia
julia> (1:5).^2 ./ Interval(5)
ERROR: MethodError: no method matching Float64(::Interval{Float64})
Closest candidates are:
  (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:200
  (::Type{T})(::T) where T<:Number at boot.jl:760
  (::Type{T})(::VectorizationBase.Double{T}) where T<:Union{Float16, Float32, Float64, VectorizationBase.Vec{var"#s160", var"#s158"} where {var"#s160", var"#s158"<:Union{Float16, Float32, Float64}}, VectorizationBase.VecUnroll{var"#s157", var"#s156", var"#s152", V} where {var"#s157", var"#s156", var"#s152"<:Union{Float16, Float32, Float64}, V<:Union{Bool, Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, VectorizationBase.Bit, VectorizationBase.AbstractSIMD{var"#s156", var"#s152"}}}} at /Users/sheehanolver/.julia/packages/VectorizationBase/2vKLL/src/special/double.jl:84
  ...
Stacktrace:
 [1] convert(#unused#::Type{Float64}, x::Interval{Float64})
   @ Base ./number.jl:7
 [2] setindex!(A::Vector{Float64}, x::Interval{Float64}, i1::Int64)
   @ Base ./array.jl:839
 [3] macro expansion
   @ ./broadcast.jl:984 [inlined]
 [4] macro expansion
   @ ./simdloop.jl:77 [inlined]
 [5] copyto!
   @ ./broadcast.jl:983 [inlined]
 [6] copyto!
   @ ./broadcast.jl:936 [inlined]
 [7] copy
   @ ./broadcast.jl:908 [inlined]
[...]

PR #358, opened 2020-02-09
Add random samplers

krish8484:


dpsanders:
Thanks for the contribution!

But I think it shouldn't be necessary to explicitly write all these methods.
Please check the documentation at https://docs.julialang.org/en/v1/stdlib/Random/#Generating-random-values-of-custom-types-1
to define a random sampler in the correct way that they all come for "free".

src/rand.jl
```diff
@@ -1,3 +1,14 @@
+using Random
+
 Base.rand(X::Interval{T}) where {T} = X.lo + rand(T) * (X.hi - X.lo)
 
 Base.rand(X::IntervalBox) = rand.(X)
+
+
+Base.eltype(::Type{Interval{Float64}}) = Float64
```
dpsanders: Unfortunately this line disagrees with https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/263.

Is this really necessary for the random samplers? That seems strange. (But it could be.)

krish8484: That line is also used to change the default return type of the random samplers according to the docs, I used it to convert the return type of the function from Interval{Float64} to a Float Vector.

dpsanders: Yes I see. What I'm saying is that that conflicts with the way that we need to define `eltype` for `Interval`s in order for broadcasting to work. Maybe you can ask about this on Slack or raise it as an issue on the JuliaLang issue tracker.

dpsanders: I think the solution is that the `Random.eltype` function should be distinguished from `Base.eltype`. I'm not sure this is possible right now?

krish8484: I reckon we can alter the return type of broadcasting in the `Base.similar` function.
Check [https://docs.julialang.org/en/v1/manual/interfaces/#Selecting-an-appropriate-output-array-1](url)

dpsanders: After asking on Slack it seems that the following solves the issue:
```
julia> Random.gentype(::Type{Interval{T}}) where {T} = T
```

krish8484: Yes it does.

test/rand.jl
```diff
@@ -22,4 +33,11 @@ using StaticArrays
         @test rand(Y) isa SVector{2,T}
     end
 
+    for T in (Float32, Float64, BigFloat)
+        X = Interval{T}(3, 7)
+        Y = rand(X,7)
```
dpsanders: Maybe you can test that the type of `Y` is correct too.
Also, please add a space after the comma.

dpsanders:
LGTM (apart from 1 minor comment) -- thanks for the contribution!

krish8484:
Thanks for the review!

krish8484:
closes #358

krish8484:
fixes #349

dpsanders:
Thanks a lot!

dpsanders:
Ah I forgot that you will need to bump the patch version in Project.toml so that we can release a new version. Could you do that in a separate PR please?

PR #362, opened 2020-02-20
Change definition of eltype and add tests

### LINKED ISSUE #263: Change eltype(x::T) to return T

krish8484:
Closes #263


dpsanders:
Any particular reason for making a new file elt.jl? If you do make a new file it should have a name that is more understandable.

Also, I suggest making a new branch in your fork to make pull requests from.

krish8484:
Functions at `intervals.jl` were interfering with the `eltype` function, so I created a new file for eltype functions namely `elt.jl`. Name will be changed if it is not okay.
Do you have an idea of what is causing the build test to fail?

krish8484:
Works only for Intervals now, let me know if the name of the file needs to be changed.

src/intervals/elt.jl
```diff
@@ -0,0 +1,5 @@
+
+Base.eltype(x::Interval{T}) where {T<:Real} = T
+
+# Base.eltype(x::T) where {T<:IntervalBox} = T
```
dpsanders: We need a definition for IntervalBox as well.

dpsanders: I really don't see the need for a separate file this short.

test/runtests.jl
```diff
@@ -1,4 +1,3 @@
-
 using IntervalArithmetic
 if VERSION < v"0.7.0-DEV.2004"
```
dpsanders: We can remove these old VERSION checks now since we require some version > 1.

dpsanders: Maybe that could/should be a separate PR though.

dpsanders:
LGTM, thanks!

You will need to bump the patch version number too.

dpsanders:
Thanks!

PR #363, opened 2020-02-22
Change definition of eltype and add brodcasting tests

### LINKED ISSUE #263: Change eltype(x::T) to return T

krish8484:
Closes #263


dpsanders:
You don't need to open a new pull request, just update the previous one.

If you need to you can use

`git push -f` 

to force update your remote repository.

krish8484:
Alright! I was using a branch in my fork for this PR. I will close it, Thanks.

Issue #287, opened 2019-05-23
Complex intervals not displayed correctly


dpsanders:
```
z = (-5..5) + (-5..5)*im;
```
[-5, 5]Error showing value of type Complex{Interval{Float64}}:
ERROR: TypeError: non-boolean (Interval{Float64}) used in boolean context
Stacktrace:
```

dpsanders:
cc @yashrajgupta

PR #367, opened 2020-03-08
Replace signbit by interval_signbit, and let signbit(::Interval)=false

### LINKED ISSUE #332: `show` is broken for complex intervals
Hi everybody,
I am using Julia 1.2.0 and I have an issue with the following simple code:

```
x= @interval(0.1,2)
complex(x,x)
```
I would like to see an output like

```
[0.0999999, 2]+[0.0999999, 2]im
```

but instead I get the following error :

```
[0.0999999, 2]Error showing value of type Complex{Interval{Float64}}:
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Complex{Interval{Float64}}) at ./complex.jl:186
 [2] show(::IOContext{REPL.Terminals.TTYTerminal}, ::MIME{Symbol("text/plain")}, ::Complex{Interval{Float64}}) at ./multimedia.jl:47
 [3] display(::REPL.REPLDisplay, ::MIME{Symbol("text/plain")}, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:132
 [4] display(::REPL.REPLDisplay, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:136
 [5] display(::Any) at ./multimedia.jl:323
 [6] #invokelatest#1 at ./essentials.jl:790 [inlined]
 [7] invokelatest at ./essentials.jl:789 [inlined]
 [8] print_response(::IO, ::Any, ::Bool, ::Bool, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:156
 [9] print_response(::REPL.AbstractREPL, ::Any, ::Bool, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:141
 [10] (::getfield(REPL, Symbol("#do_respond#38")){Bool,getfield(REPL, Symbol("##48#57")){REPL.LineEditREPL,REPL.REPLHistoryProvider},REPL.LineEditREPL,REPL.LineEdit.Prompt})(::Any, ::Any, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:718
 [11] #invokelatest#1 at ./essentials.jl:790 [inlined]
 [12] invokelatest at ./essentials.jl:789 [inlined]
 [13] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface, ::REPL.LineEdit.MIState) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/LineEdit.jl:2306
 [14] run_frontend(::REPL.LineEditREPL, ::REPL.REPLBackendRef) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:1038
 [15] run_repl(::REPL.AbstractREPL, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.2/REPL/src/REPL.jl:201
 [16] (::getfield(Base, Symbol("##737#739")){Bool,Bool,Bool,Bool})(::Module) at ./client.jl:390
 [17] #invokelatest#1 at ./essentials.jl:790 [inlined]
 [18] invokelatest at ./essentials.jl:789 [inlined]
 [19] run_main_repl(::Bool, ::Bool, ::Bool, ::Bool, ::Bool) at ./client.jl:374
 [20] exec_options(::Base.JLOptions) at ./client.jl:312
 [21] _start() at ./client.jl:464
```
Moreover, the same code seems to work with Julia 1.0.3.
What do you think? Thanks for your answer!

lbenet:
Fixes #332,  #287 and #402


lbenet:
The idea is to define `interval_signbit` (which returns an interval) so `copysign` behaves properly (cf #273), and fix displaying complex intervals as proposed [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/332#issuecomment-592825327).

cc @dpsanders @dlfivefifty @fp4code

dpsanders:
The question is if this affects anything somewhere else in some function that checks `signbit` to see if a number is negative.

We are coming up against limits of Julia's type system once again.

dpsanders:
Isn't the new `interval_signbit` the same as the `sign` function for an `Interval`?

lbenet:
> The question is if this affects anything somewhere else in some function that checks signbit to see if a number is negative.

Within `IntervalArithmetics` everything is consistent. I have only checked that `IntervalRootFinding.jl` works fine, and it does. 

Incidentally, the standard does mentions (requires?) a `sign` function; the standard states "sign(x) is −1 if x<0; 0 if x=0; and 1 if x>0. It is discontinuous at 0 in its domain".

> Isn't the new interval_signbit the same as the sign function for an Interval?

`interval_signbit` is identical to `signbit` in current master. As you noticed, it is *almost* the same as `sign`. The subtle difference is that sign returns the signs of the edges of the interval, while `interval_signbit` returns the signbit, in reversed order. Then, we get `sign(-1 .. 1) == -1 .. 1` whereas `interval_signbit(-1 .. 1) == 0 .. 1`.

I introduced `interval_signbit` as a simple workaround, to allow having `signbit(::Interval) = false` which permits the correct display of complex intervals, and also have a function which does what `signbit` is currently doing in master; that's the reason that one method of `copysign` and `flipsign` were modified.

lbenet:
I have checked all packages in JuliaRegistries/General that have IntervalArithmetics.jl as a dependency. I went one by one, looking for `signbit` using GitHub search utility. I hope this is somewhat thorough.

I have found only two packages which somehow use `signbit`: https://github.com/JuliaIntervals/MPFI.jl and https://github.com/gwater/NumberIntervals.jl. In `MPFI.jl`, there is a test that involves an end of an interval, so it tests `signbit(::Float64)`. In NumberIntervals.jl there are a couple of tests that do involve `signbit` of a `NumberInterval`; the function is overloaded [here](https://github.com/gwater/NumberIntervals.jl/blob/441343df76307ea0c680bfa196815b87e452cb95/src/nonstandard.jl#L69).

So I guess we should ask @gwater about this PR.

gwater:
I have no objection to removing the `signbit` method from IntervalArithmetic. Since NumberIntervals imports it from Base and implements its own method, I doubt I even need to adapt anything.

dpsanders:
Is this ready to merge? I confirm it fixes the `show` bug for complex intervals.

dlfivefifty:
If I recall the reason I added the `signbit` overload was QR with intervals. Can we add a test for this?

lbenet:
@dlfivefifty I am in favor to add a test, can you please suggest one? 

I just came up with the following, but do not know if this is what you have in mind:
```julia
julia> using IntervalArithmetic, LinearAlgebra

julia> A = [Interval(1) Interval(-1); Interval(1) Interval(1)]
2×2 Array{Interval{Float64},2}:
 [1, 1]  [-1, -1]
 [1, 1]    [1, 1]

julia> Q = qr(A)
QR{Interval{Float64},Array{Interval{Float64},2}}
Q factor:
2×2 LinearAlgebra.QRPackedQ{Interval{Float64},Array{Interval{Float64},2}}:
 [-0.707107, -0.707106]  [-0.707107, -0.707106]
 [-0.707107, -0.707106]   [0.707106, 0.707107] 
R factor:
2×2 Array{Interval{Float64},2}:
    [-1.41422, -1.41421]  [-3.33067e-16, 4.4409e-16]
 [0, 0]                    [1.41421, 1.41422]       

julia> (1/sqrt(2)) ∈ Q.Q[2,2]
true

julia> sqrt(2) ∈ Q.R[2,2]
true
```

Would that be ok?

dpsanders:
Does a qr decompositon actually make sense for interval matrices? What does it mean for two interval vectors to be orthogonal?

lbenet:
I do not know if `qr` makes sense or not for intervals, and you certainly have a good point about questioning the meaning of two interval vectors being orthogonal. My guess is that one could say that the interval vectors contain vectors that are indeed orthogonal.

@dlfivefifty Can you suggest a better test?

dlfivefifty:
QR makes sense in that its the QR decomposition calculated via interval arithmetic. That is, Q and R are intervals containing the true QR decomposition.

I think the test is fine.

gwater:
> 
> 
> QR makes sense in that its the QR decomposition calculated via interval arithmetic. That is, Q and R are intervals containing the true QR decomposition.
> 
> I think the test is fine.

I'd like to remind you of [this issue](https://github.com/JuliaArrays/StaticArrays.jl/issues/450) which is caused almost exactly by this test case (inverting a small matrix of Intervals through an external package which is unaware of Intervals). I fear, adding a test case will only further encourage users to expect results IntervalArithmetic.jl does not currently deliver (namely, safe matrix inversion through external libraries).

lbenet:
So we better leave the tests like this...?

dlfivefifty:
Don’t let the perfect be the enemy of the good: there are cases where interval QR works exceptionally well for inverting matrices, in particular, when A is diagonally dominant. 

Also note that ` \ ` presumably uses Gaussian elimination which is well-known to be ill-conditioned in small neighbourhoods, which interval arithmetic picks up. QR is a very different beast: it’s always stable, interval arithmetic may be a bit pessimistic in its bounds but not in a catastrophically bad way.

I’d actually propose overriding ` \ ` to use QR for interval matrices.

gwater:
I'm certainly not claiming that it's impossible to invert/decompose a matrix - the problem is that IntervalArithmetic.jl has no way of telling users *when* it's impossible (again, see the issue I referenced above). [NumberIntervals.jl](https://github.com/gwater/NumberIntervals.jl) demonstrates how to implement interactions with interval-unaware algorithms safely.

dlfivefifty:
Like I said: don’t let the perfect be enemy of the good. I encountered this issue because I needed it.

gwater:
> 
> 
> Like I said: don’t let the perfect be enemy of the good. I encountered this issue because I needed it.

I guess that is kind of my point: If you need to invert interval matrices, you might benefit from using the NumberIntervals wrapper because it protects you from cases which cannot be solved using Interval arithmetic. And there is no need to test functions from the standard library to guarantee that.

dpsanders:
I'm not sure what you mean by "cannot be solved" here.

dpsanders:
I actually quite like the solution in `NumberIntervals`. It's a clever way of introducing a 3-valued logic by returning `missing` when both `true` and `false` are correct answers for different parts of an interval (if I understand correctly).

gwater:
> 
> 
> I'm not sure what you mean by "cannot be solved" here.

As we saw in https://github.com/JuliaArrays/StaticArrays.jl/issues/450 sometimes external algorithms give somewhat unpredictable results when confronted with IntervalArithmetic.jl's intervals. Therefore I would say some matrices in conjunction with some external algorithms represent problems which IntervalAirthmetic.jl is unable to solve.

gwater:
> 
> 
> I actually quite like the solution in `NumberIntervals`. It's a clever way of introducing a 3-valued logic by returning `missing` when both `true` and `false` are correct answers for different parts of an interval (if I understand correctly).

Thanks, it really copies all of IntervalArithmetics logic for set operations and reimplements number operations with the three-value logic you described. When I finally implemented it is was a lot simpler than I had imagined. (It lacks some of IntervalArithmetic's syntactic sugar though.)

lbenet:
I think @gwater has a point: Let's not include the proposed test, simply because it may give a false impression that things work "out of the box", when they may not.  

What about checking if the changes proposed in this PR do not break the code which motivated the introduction of `signbit`? Could you share the code that motivated `signbit`, @dlfivefifty ?

dlfivefifty:
The code that motivated it was calling QR on an interval valued matrix! So this test is the MWE.

I still don’t see nor believe there’s a problem in QR, which is very different from \

dlfivefifty:
You also don’t want to falsely imply things don’t work out of the box, when they do...

dpsanders:
I agree that we should include the qr test. And we should think about whether we should incorporate the use of three valued logic in the future.

We should also probably overload \ in the package as Sheehan suggests

dpsanders:
@gwater How does the 3-valued logic affect performance though? Since that introduces a weak type instability, right?

gwater:
> @gwater How does the 3-valued logic affect performance though? Since that introduces a weak type instability, right?

I haven't benchmarked it yet but from what I hear these small Union's are quite performant.

dlfivefifty:
> I haven't benchmarked it yet but from what I hear these small Union's are quite performant.

It's roughly as performant as a branch (i.e., adding an `if` statement). So compared to pure `Float64` arithmetic union types will be significantly slower, but probably for interval arithmetic it won't be noticeable.

lbenet:
Rebased to master

lbenet:
Bump

dpsanders:
@lbenet Thanks! Could you please add the test of QR?

dpsanders:
Is this ready?

lbenet:
Yes, I think it is.

gwater:
As far as I can see, this PR implements `signbit` but just makes every interval a positive number, effectively.
```julia
 signbit(::Interval) = false 
```
Is this really a good idea? If you don't want to implement a useful method, I would just leave it unimplemented.

lbenet:
The idea is to solve `show` for complex intervals; check the error message when you try to do it in current master. This is a simple way to get around that error.

dpsanders:
I think we should just extend `Base.show` for complex intervals instead.

Kolaru:
Superseeded by more recent developments.

Issue #215, opened 2018-09-18
Bugs with tiny numbers


dpsanders:
```jl
julia> tiny = interval(0, floatmin())
[0, 2.22508e-308]

julia> huge = interval(floatmax(), Inf)
[1.79769e+308, ∞]

julia> tiny * tiny
[0, 0]

julia> huge * huge
[1.79769e+308, ∞]

julia> huge / tiny
[1.79769e+308, ∞]

julia> tiny / huge
[0, 0]
```

In particular, `tiny*tiny` and `tiny / huge` should not be exactly 0.

cc @JeffreySarnoff

dpsanders:
This may be to do with corner cases in FastRounding.jl.

dpsanders:
Using directed rounding via changing the rounding mode gives correct answers:
```jl
julia> setrounding(Interval, :slow)
:slow

julia> tiny * tiny
[0, 4.94066e-324]

julia> tiny / huge
[0, 4.94066e-324]
```

JeffreySarnoff:
How do you get `2.25e-308 * 2.25e-308` not `0.0` with any rounding mode?
```
frexp(floatmin(Float64))
(0.5, -1021)
```
`floatmin(Float64) * floatmin(Float64) = ldexp(0.5 * 0.5, -1021 -1021) == ldexp(0.25, -2042)`

dpsanders:
Ah, the problem is here:
```jl

julia> x = floatmin()
2.2250738585072014e-308

julia> two_prod(x, x)
(0.0, 0.0)
```
which is correct as far as it goes...

JeffreySarnoff:
are these issues limited to denormals?

dpsanders:
No, I think it's any numbers for which the low-order part of `two_prod` is 0, so `x <= 1e-151` approximately.

JeffreySarnoff:
Is it only products?

JeffreySarnoff:
This is the other side of my note earlier about how low order 0.0 usually are nonzero [although guessing whether the value >0.0 or <0.0 seems difficult]

JeffreySarnoff:
Do you think we need to trap and special case products where an arg < ~1.0e-150?

dpsanders:
It seems like special casing there is necessary, yes :/

I guess there will be the same problem with division for small and large numbers.

JeffreySarnoff:
[...]

PR #370, opened 2020-04-01
Another rounding option

matsueushi:
Ref https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/215. I added a new rounding option using RoundingEmulator.jl as `:emulation` for testing purposes. I switched the rounding mode in the test scripts from `:tight` to `:emulation` and ran all tests, then it worked fine on [my Travis builds](https://github.com/matsueushi/IntervalArithmetic.jl/runs/550758791). Please let me know if I'm missing something.

@dpsanders @krish8484

### DIFF
--- a/src/intervals/rounding.jl
+++ b/src/intervals/rounding.jl
@@ -4,15 +4,15 @@ This is a so-called "traits-based" design, as follows.
 
 The main body of the file defines versions of elementary functions with all allowed
 interval rounding types, e.g.
++(IntervalRounding{:fast}, a, b, RoundDown)
 +(IntervalRounding{:tight}, a, b, RoundDown)
-+(IntervalRounding{:emulation}, a, b, RoundDown)
 +(IntervalRounding{:accurate}, a, b, RoundDown)
 +(IntervalRounding{:slow}, a, b, RoundDown)
 +(IntervalRounding{:none}, a, b, RoundDown)
 
 The current allowed rounding types are
-- :tight     # fast, tight (correct) rounding with errorfree arithmetic via FastRounding.jl
-- :emulation # tight (correct) rounding with improved errorfree arithmetic via RoundingEmulator.jl
+- :fast     # fast, tight (correct) rounding with errorfree arithmetic via FastRounding.jl
+- :tight # tight (correct) rounding with improved errorfree arithmetic via RoundingEmulator.jl
 - :accurate # fast "accurate" rounding using prevfloat and nextfloat  (slightly wider than needed)
 - :slow    # tight (correct) rounding by changing rounding mode (slow)
 - :none     # no rounding (for speed comparisons; no enclosure is guaranteed)
@@ -88,7 +88,7 @@ for (op, f) in ( (:+, :add), (:-, :sub), (:*, :mul), (:/, :div) )
 
             mode2 = Symbol("Round", mode)
 
-            @eval $op(::IntervalRounding{:tight},
+            @eval $op(::IntervalRounding{:fast},
                                 a::$T, b::$T, $mode1) = $ff(a, b, $mode2)
         end
     end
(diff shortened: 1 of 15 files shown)


dpsanders:
Great news, thanks!

dpsanders:
In my initial tests this is very promising indeed, thanks @matsueushi!

I think we should just make this the default option.
cc @lbenet, @Kolaru.

test/interval_tests/rounding.jl
```diff
@@ -34,6 +34,12 @@ setrounding(Interval, :tight)
     @test sin(x) == Interval(0.47942553860420295, 0.479425538604203)
 end
 
+setrounding(Interval, :emulation)
+@testset "Tight rounding using RoundingEmulation.jl" begin
+    @test rounding(Interval) == :emulation
+    @test sin(x) == Interval(0.47942553860420295, 0.479425538604203)
```
dpsanders, line 40: I don't think `sin` is a good test here, since that doesn't actually use the `emulation` functions at all?

dpsanders: Maybe you can add some of the tests from #215 instead.

matsueushi: Yes, I added tests for `tiny` and `huge` of #215.

src/intervals/rounding.jl
```diff
@@ -116,6 +116,23 @@ for T in (Float32, Float64)
 end
 
 
+
```
Kolaru, line 119: Could you add inline comment to clarify what's going on in there ? (after some research in `RoundingEmulator.jl` doc I think I get it, but future us will probably be thankful to have a bit of inline doc here)

Kolaru:
If it solves a bug (even a very unlikely one), we should definitely use this as the default.

Is there a reason to keep the `:tight` mode if it gives the same results? (except for the one which were previously buggy)

It looks like this implementation is falling back to the `:slow` rounding mode for functions that are not defined by `IntervalEmulator`. If `:emulation` is to become/replace the current default, then I think the fallback should be to `:tight`.

Finally it would be good to document the rounding mode, at least in the header of the `setrounding.jl` file.

matsueushi:
Thanks @Kolaru, I added comments to rounding.jl. 

In my understanding, the arithmetic operations provided by`FastRounding` are `+`, `-`, `*`, `inv`, `/` and `sqrt`, and other functions are falling to `:slow`. All these basic arithmetic operations are also handled by 'RoundingEmulator', so I guess even if we change the fallback to `:tight`, `:slow` will be used eventually.

dpsanders:
The difficult-to-read, opaque, brittle, generally disgusting and undocumented code in `rounding.jl` is my fault...  Sorry. But that's what ends up happening with this metaprogramming stuff -- it's just really hard to read.

If anybody has a suggestion for how to clean it up, that would be great.

dpsanders:
I think this is almost ready to be merged. 
Does this PR make `emulation` the default mode?
Maybe we should just rename `emulation` to `tight` or move `tight` to another word.

Although `tight` is not correct for very small and very large values, it is up to twice as fast as `emulation` for some operations, e.g. multiplication, so we might want to keep it around.

I wonder if we should actually bump the minor version number since changing to `emulation` would technically be a breaking change. Or you could just regard it as a bug fix. I propose we just bump the patch version. @matsueushi Could you bump the patch version in `Project.toml` please?

matsueushi:
Then can we rename the `emulation` to `tight` and move the current `tight` to `fast`?

dpsanders:
Yes, great idea!

matsueushi:
I renamed rounding modes and bumped version.

lbenet:
I think we should bump a new minor version once this is merged.

Thanks @matsueushi , really great contribution!

matsueushi:
I found the lower bound of julia compat in Project.toml of `RoundingEmulator` is 1.3, which is higher than 1.1 of `IntervalArithmetic`. What versions of julia is supported by `IntervalArithmetic`? 1.1 or higher?

dpsanders:
Yes I think IntervalArithmetic.jl still in principle supports 1.1 and higher? 
However, I see that we are now only testing with Julia 1.3, so maybe we should either bump that up or someone should actually test it with 1.1.  If we are releasing a new minor version anyway then we should increase this to at least 1.3 I would say.

I seem to remember that there was a change which means that it does *not* work with Julia 1.0.5; I think it was the removal of `setrounding` from Base (ironically, given this PR!)

matsueushi:
Can I increase the lower bound to Julia 1.3 and bump the minor version instead of the patch version? If we support 1.2, I need to make a change on `RoundingEmulator` because `Base.significand_bits` was added in 1.3.

dpsanders:
Yes go ahead. And could you please bump Recipesbase to 1.0 as well?

matsueushi:
Thanks. I'll update Recipesbase and add some comments to NEWS.md.

matsueushi:
Done. I also bumped Polynomials to 0.7 because 0.6 is incompatible with RecipeBase 1.0.

dpsanders:
Thanks very much for this great contribution @matsueushi.

If nobody has any objections, I'll merge it soon.

lbenet:
I'm definitely in favor!

dpsanders:
Many thanks @matsueushi!

PR #377, opened 2020-05-11
Add cbrt function for intervals

### LINKED ISSUE #310: Base.cbrt() causes stack overflow on intervals
how to reproduce
```
julia> cbrt(Interval(1,2))
ERROR: StackOverflowError:
Stacktrace:
 [1] cbrt(::Interval{Float64}) at ./special/cbrt.jl:34 (repeats 80000 times)
```
This happens due to an unfortunate combination of facts:
* `float(Interval(1,2))` returns an `Interval`
* Intervals have supertype `Real`
* Base expects that `float()` returns something a subtype of `AbstractFloat`

possible fixes:
* change supertype of Interval to `AbstractFloat`
* specialize `cbrt()` for Intervals

krish8484:
Fixes #310


src/intervals/functions.jl
```diff
@@ -233,6 +233,15 @@ function sqrt(a::Interval{T}) where T
     @round(sqrt(a.lo), sqrt(a.hi))  # `sqrt` is correctly-rounded
 end
 
+
+
+function cbrt(a::Interval{T}) where T
+    isempty(a) && return a
+
+    Interval(cbrt(a.lo), cbrt(a.hi))
```
dpsanders, line 241: This does not take account of directed rounding.

dpsanders: e.g. try `cbrt(1..3)` instead and compare to the result when you use a BigFloat interval.

krish8484: We will need a similar function like sqrt_round in [FastRounding.jl](https://github.com/JeffreySarnoff/FastRounding.jl/blob/master/src/FastRounding.jl), right?

lbenet: I think you can mimic the code for `sqrt`. Have you tried that?

krish8484:
Some unintentional changes are made due to version issues, I will close this one and fix it in my next PR.

dpsanders:
There's no need to close and reopen. Just force push to this branch with

`git push -f`

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.17.2"
+version = "0.17.0"
```
dpsanders: This isn't right. We should be increasing the version as you had before.

dpsanders: Each minor change like this should increment the "patch version", i.e. the 3rd number.

krish8484: In master, it is 0.17.1 so I initially changed it to 0.17.2 but the tests were conflicting so I reverted the commit.
Should I keep it at 0.17.2?

krish8484:
@dpsanders

krish8484:
@dpsanders

Project.toml
dpsanders: Yes keep at 0.17.2.
What do you mean by "the tests were conflicting"?

dpsanders:
You need to rebase on master I think and fix any conflicts.

From your branch do

`git rebase master`

It will tell you which files are in conflict. You have to fix those then add them as instructed, then do 
`git rebase --continue`

src/intervals/rounding.jl
```diff
@@ -69,6 +69,10 @@ sqrt(a::T, rounding_mode::RoundingMode) where {T<:Rational} = setrounding(float(
     sqrt(float(a))
 end
 
+cbrt(a::T, rounding_mode::RoundingMode) where {T<:Rational} = setrounding(float(T), rounding_mode) do
```
dpsanders, line 72: This `setrounding` stuff feels out of date. Is that what other functions do?

src/intervals/rounding.jl
```diff
@@ -114,12 +118,15 @@ for T in (Float32, Float64)
 
         @eval sqrt(::IntervalRounding{:fast},
                             a::$T, $mode1) = sqrt_round(a, $mode2)
+
+        @eval cbrt(::IntervalRounding{:fast},
+                            a::$T, $mode1) = cbrt_round(a, $mode2)
```
dpsanders, line 123: I don't think there is a `cbrt_round` function? Or is there?

test/interval_tests/numeric.jl
```diff
@@ -353,6 +353,12 @@ end
         @test sqrt(big(2..3)) == Interval(big"1.414213562373095048801688724209698078569671875376948073176679737990732478462102", big"1.732050807568877293527446341505872366942805253810380628055806979451933016908815")
     end
 
+    @testset "cbrt" begin
+        @test cbrt(2..3) == Interval(1.259921049894873, 1.4422495703074085)
+
+        @test cbrt(big(2..3)) == Interval(big"1.259921049894873164767210607278228350570251464701507980081975112155299676513956", big"1.442249570307408382321638310780109588391869253499350577546416194541687596830003")
```
dpsanders, line 359: Maybe you should check that the `big` version is contained inside the float version.
In general we should probably add tests like that to the tests.

dpsanders:
Please don't use `git merge`.  Use `git rebase` as I mentioned.

krish8484:
I have used `git rebase` and made the necessary changes.
Thanks @dpsanders

dpsanders:
You can use `git rebase -i` ("interactive") to remove unnecessary commits and then force push.
There are various manuals online that explain the process. 

Alternatively you can just make a new branch with the correct commits, e.g. by `git cherry-pick`. 

None of this is really necessary for this PR, but they are useful skills to learn.

dpsanders:
LGTM

dpsanders:
Thanks!

dpsanders:
Thanks!

PR #378, opened 2020-05-11
Update plot recipe to use seriesalpha

dpsanders:


lbenet:
Thanks! Let's merge it right away, once tests have passed.

PR #379, opened 2020-05-12
Speed up intersect

dpsanders:
Reverse the order of operations to avoid doing effectively the same check twice.


dpsanders:
```
x = 3..5
y = 4..6
```

Master:
```
julia> @btime $(Ref(x))[] ∩ $(Ref(y))[]
  9.689 ns (0 allocations: 0 bytes)
[4, 5]
```

This PR:
```jl
julia> @btime $(Ref(x))[] ∩ $(Ref(y))[]
  4.134 ns (0 allocations: 0 bytes)
```

(Is the benchmark suite operational?)

PR #388, opened 2020-06-06
[WIP / RFC] Configuration mechanism for directed rounding and fast powers

### LINKED ISSUE #103: Power ^ allocates

dpsanders:
This is a proposal for a better configuration mechanism for changing the directed rounding method (tight or accurate) and for choosing the type of powers (fast or tight). 

Usage:
```jl
IntervalArithmetic.configure!(directed_rounding=:fast, powers=:tight)
```

This would fix several outstanding issues (that I won't list right now).

This is the minimal thing that makes sense to me. The syntax is nasty enough that it feels like something a casual user should not do (unlike the previous `setrounding(Interval, :tight)`); in particular I am proposing *not* to export `configure!`), but light enough that it is not unusable. In any case this is designed to be used a very minimal number of times in any given piece of code.

Tests and docs need updating but I would appreciate feedback.

cc-ing interested parties: @lbenet, @Kolaru, @gwater, @jeffbezanson, @mforets

fixes #103


dpsanders:
Note that the mechanism itself is still the same: globally redefine all functions according to the selected mode. There does not seem to be a workable alternative to this.

Kolaru:
I think this is conceptually fine. One way of another, changing the behavior of function must somehow be a redefinition anyway.

For our own testing purpose, I think it would be good to have a do-block syntax available, as it will likely make tests more readable and safer.

dpsanders:
@Kolaru Thanks for the feedback.

I agree that having a do-block syntax would be good for testing purposes. I'm not sure I'll manage that in this PR though.

mforets:
Bump :smile:  Is this "only" needing to add testing and docs?

dpsanders:
And probably other rounding modes, yes.

OlivierHnt:
What is the current state of this PR?

lucaferranti:
@lbenet I have fixed the merge conflicts so this should be ready for review / merge.

Couple of points

- the PR uses fast power by default, I think we should use tight by default (because it's the current behavior, because otherwise it would be a breaking change and for coherence for the rounding mode, which is tight by default)
- Does the rounding with `FastRounding` disappear in this PR? based on [these lines](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/388/files#diff-f4e42530f3dd219de6586506c637aea2ce7c5364e5487fa28939e617a39c0794R9-R12)
- tests for fast power and non-tight modes missing. A good to have test for fast is the one in #491 (btw that seems to work now on this branch, so that issue probably can be closed)

lucaferranti:
fixes #14

src/IntervalArithmetic.jl
```diff
@@ -80,7 +81,7 @@ export
 
 export showfull
 
-import Base: rounding, setrounding, setprecision
+import Base: setprecision
```
lbenet, line 84: Since we are getting rid of `rounding` and `setrounding`, shouldn't we amend (in parallel) the [docs](https://juliaintervals.github.io/pages/tutorials/tutorialArithmetic/index.html#advanced_options_precision_and_rounding_modes)? Similarly, we should add the corresponding docs of `IntervalArithmetic.set_directed_rounding`

src/intervals/configuration.jl
```diff
@@ -0,0 +1,57 @@
...
+    - `:none`: no rounding; does *not* guarantee that the results are correct.
+
+- `powers`:
+    Method to implement powers. Possible values are
+    - `:tight`: gives tightest possible result by using `BigFloat` internally, and hence is slow
+    - `:fast`: uses repeated multiplication based on `Base.power_by_squaring`, and hence is faster but gives slightly looser results
+"""
+
+const configuration = Dict(:directed_rounding => :tight,
+                           :powers => :fast)
```
lbenet, line 21: I think we should have as default configuration that *both* defaults are `:tight`; It seems to me a bit unnatural to have `(:directed_rounding => :tight` and `:powers => :fast`. Note that tests change to this mode.

src/intervals/configuration.jl
```diff
@@ -0,0 +1,57 @@
...
+
+const allowed_values = Dict(:directed_rounding => (:tight, :fast, :none),
+                            :powers => (:tight, :fast))
+
+function configure!(; directed_rounding=nothing, powers=nothing)
```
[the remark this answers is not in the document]
lbenet: Rebump question...

src/intervals/rounding.jl
```diff
@@ -307,33 +308,32 @@ function _setrounding(::Type{Interval}, rounding_type::Symbol)
 
     end
 
-    current_rounding_type[] = rounding_type
+    # current_rounding_type[] = rounding_type
 end
 
 """
-    setrounding(Interval, rounding_type::Symbol)
+    set_directed_rounding(rounding_type::Symbol)
```
lbenet, line 315: This is a simple reminder to add the proper docs to https://juliaintervals.github.io/pages/tutorials/tutorialArithmetic/index.html#advanced_options_precision_and_rounding_modes

test/interval_tests/intervals.jl
```diff
@@ -10,6 +10,6 @@ include("linear_algebra.jl")
 include("loops.jl")
 include("parsing.jl")
 # include("rounding_macros.jl")
-include("rounding.jl")
+#include("rounding.jl")
```
lbenet, line 13: Shouldn't we adapt the tests to the changes implemented here?

lbenet (commented):
Thanks a lot @lucaferranti for dealing with rebasing to current master. 

In general, LGT. Still,  Ileft few comments, the important ones related to testing this (instead of commenting the tests), and also about changing the defaults. I think the actual only requirement is to bump to the next minor version.

OlivierHnt:
Superseded by PR #593. Let us re-visit this PR if the power mechanism fails to address all the concerns.

PR #390, opened 2020-06-11
Completes mpfi.jl and fi.lib.jl

krish8484:
Added cot, coth, csc, csch, sec and sech for intervals with corresponding tests from ITF1788 suite(mpfi.jl and fi_lib.jl)


dpsanders:
This looks nice, thanks!

It would be great to add versions of these functions in the `:accurate` rounding mode that do not use MPFR, but just use the definition of the functions as the inverse of other trig functions etc.

dpsanders:
That can also be a separate PR if this is ready?

krish8484:
Yes, this is ready. I will work on adding these functions  in `:accurate` rounding mode in another one.

dpsanders:
This looks good, apart from `cot` near pi/2.

krish8484:
I reckon you were on a different branch

Master
```
julia> cot(1.5..1.6)
[-∞, ∞]
```

This Branch
```
julia> cot(1.5..1.6)
[-0.029212, 0.0709149]
```

dpsanders:
Apologies, I think you're right that I must have had the wrong branch checked out.
`cot` indeed does seem to work.

Thanks a lot!

PR #391, opened 2020-06-15
Display Interval{Float32} with f0

### LINKED ISSUE #387: Display Interval{Float32} differently
e.g. `[1f0, 2f0]`

krish8484:
Closes #387

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -204,6 +204,40 @@ function basic_representation(a::Interval, format=nothing)
     output
 end
 
+function basic_representation(a::Interval{Float32}, format=nothing)
+    if isempty(a)
+        return "∅"
+    end
+
+    if format == nothing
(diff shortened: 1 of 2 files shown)


src/display.jl
```diff
@@ -204,6 +204,40 @@ function basic_representation(a::Interval, format=nothing)
     output
 end
 
+function basic_representation(a::Interval{Float32}, format=nothing)
```
dpsanders, line 207: You are duplicating a lot of code. You should be able to do this by multiple dispatch instead for just the bits that change depending on whether you have Float64 or Float32 or BigFloat.

dpsanders:
Since this is definitely a net positive, I'm happy to merge for now, but at some point the code should definitely be refactored to reduce duplication (although this is not really urgent).

Thanks a lot!

PR #399, opened 2020-07-17
Completes libieeep1788_reduction.jl

krish8484:
Adds and completes @yashrajgupta's open PR #288. All tests from libieeep1788_reduction.jl pass now.


src/intervals/reduction.jl
```diff
@@ -0,0 +1,24 @@
+# This file is part of the IntervalArithmetic.jl package; MIT licensed
+
+"""
+The file contains the reduction functions stated in the section 12.12.12 with correctly rounded results.
+r is the rounding direction which takes 0.0 for RoundingToZero , 0.5 for RoundingNearest , +Inf for RoundingUp and -Inf for RoundingDown.
+"""
+function vector_sum(v :: Vector{T} , r :: RoundingMode) where T
+    sum1 = sum(BigFloat.(v))
```
dpsanders, line 8: This sum might introduce rounding errors? How do you know which precision to use?

dpsanders: Same for all the other functions.

krish8484: According to the standard, for reduction functions, `evaluation is as if in exact arithmetic up to the ﬁnal rounding, with no possibility of intermediate overﬂow or underﬂow`. Do we have an alternative of `sum` for `BigFloats`?

src/IntervalArithmetic.jl
```diff
@@ -64,7 +64,8 @@ export
     cancelminus, cancelplus, isunbounded,
     .., @I_str, ±,
     pow, extended_div,
-    setformat, @format
+    setformat, @format , vector_sum, vector_dot,
+    vector_sumAbs, vector_sumSquare
```
Kolaru, line 68: I think that would be more Julian to extend base function with the following methods:

```julia
sum(v::Vector, r::RoundingMode)
sum(f::Function, v::Vector, r::RoundingMode)
dot(v::Vector, r::RoundingMode)
```
That way `vector_sumAbs` and `vector_sumSquare` are given by respectively `sum(abs, v, r)` and `sum(abs2, v, r)`.

src/intervals/reduction.jl
Kolaru: I think the only way is to do the loop ourselves, either explicitely or using `reduce`, and using the rounded `+` e.g.

```julia
sum(v::Vector, r::RoundingMode) = reduce((x, y) -> +(x, y, r), v)
```

src/IntervalArithmetic.jl
dpsanders: Strictly speaking that is type piracy, but I agree that makes more sense.

src/intervals/reduction.jl
dpsanders: I believe these functions are supposed to use long accumulators, i.e. with high precision.

src/intervals/reduction.jl
```diff
@@ -0,0 +1,24 @@
...
+    return T(sum1, r)
+end
+
+function vector_dot(v :: Vector{T}, u :: Vector{T}, r :: RoundingMode) where T
+    sum1 = sum(BigFloat.(v) .* BigFloat.(u))
```
dpsanders, line 13: We need to make sure that the precision of the `BigFloat` is high enough that the result will be exact.

PR #403, opened 2020-08-03
Completes libieeep1788_tests_overlap.jl

krish8484:
This PR aims to satisfy tests of the file `libieeep1788_tests_overlap.jl`, and thereby completing #266.


dpsanders:
I thought `overlap` was already implemented before?

I think we would want it to return `Symbol`s, rather than `String`s.

lucaferranti:
superseeded by #518

PR #414, opened 2020-09-21
Extend BigFloat fallback to other floating-point types

mfherbst:
Makes the fallback also work for Float32 and Float16.


dpsanders:
Thanks, the support for Float32 and Float16 definitely needs some work in general. 

Could you please add some tests?

mfherbst:
Just like

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a1cbb81223e350a23b6d03f0de9456734b88a982/test/interval_tests/numeric.jl#L357

dpsanders:
Yes exactly. Basically just to make sure that the code path is used and works correctly, to prevent this being removed in the future.

mfherbst:
Turned out that explicit testing was a good call @dpsanders :smile:. I found quite a few other thing that were needed to get it to work and along the way also made a few more functions Float32-ready.

dpsanders:
Hah, I'm very not surprised. The few times I've tried to use `Float32` everything indeed broke pretty quickly.
That needs to be a focus moving forward.

Thanks for looking into this!

dpsanders:
This looks nice, thanks!

I'm wondering if `@typeinterval` could just be a method of `@interval`? I think it could.

mfherbst:
I think so too, but could not get it to work so far. I'll take another look.

mfherbst:
Ok. I came up with something reasonable. It's not perfect as `@interval Float32 1.0` does not work ... you need to write `@interval Float32 1.0 1.0` explicitly. Getting the former to work is, however, hard because you do not have access to types for dispatch in macros. So the two-argument version `@interval 1.0 1.2` and `@interval Float32 1.0` are hard to distinguish.

Kolaru:
> So the two-argument version @interval 1.0 1.2 and @interval Float32 1.0 are hard to distinguish.

This could be bypassed by reworking the macro more fundamentally. Basically what it does is wrapping number literal as string so that they are not automatically parsed. It should be possible to let it just pass the wrapped literals to `atomic` and let `atomic` do the dispatching at run time (thus having access to type info).

This may be beyond the scope of this PR though, so unless you are very motivated by macro magic, it can be done at a latter point.

Also I tracked the failing test on windows, they are due to some missing method in MPFR for `Float16`.

Changing the method https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/f80b6835814cfb05f3b73173d25a55b50914ff10/src/intervals/rounding.jl#L57 to
```julia
# use higher precision float parser to get round issues on Windows
@static if Sys.iswindows()
    function parse(::Type{Float64}, s::AbstractString, r::RoundingMode)
        a = setprecision(BigFloat, 53) do
                setrounding(BigFloat, r) do
                    parse(BigFloat, s)   # correctly takes account of rounding mode
                end
            end

        return Float64(a, r)
    end

    function parse(::Type{T}, s::AbstractString, r::RoundingMode) where {T <: Union{Float16, Float32}}
        return T(parse(Float64, s, r), r)
    end
end
```

solve the problem. I've added a static windows check too, since according to the comment the issue should only happen with windows anyway. If you could integrate this change and check it does break anything on non-windows plateform, I think the PR would be good to be merged =).

src/IntervalArithmetic.jl
```diff
@@ -52,7 +52,7 @@ import .Broadcast: broadcasted
 export
     AbstractInterval, Interval,
     interval,
-    @interval, @biginterval, @floatinterval, @make_interval,
+    @interval, @biginterval, @floatinterval, @typeinterval,
```
Kolaru, line 55: A change removed `@typeinterval` so no need to export it anymore.

mfherbst:
> This may be beyond the scope of this PR though, so unless you are very motivated by macro magic, it can be done at a latter point.

Sounds fun, but I'm afraid I lack the time to do this at the moment. I'll add your suggestions and than we see with the CI.

Kolaru:
The coverage is decreasing because coveralls runs on a linux machine so the windows only lines are never hit. Therefore everything is fine, thanks a lot!

mfherbst:
No probs. Thanks for merging!

Issue #415, opened 2020-09-21
Improve `Float32` support


dpsanders:
`Interval{Float32}` support is very patchy, c.f. e.g. #414

`Float16` is less useful since arithmetic operations on `Float16` just promote to `Float32` anyway.

remi-garcia:
Couldn't tests be written to enforce a Float32 (and a Float8 and Float16 "just in case") support?
By that I mean rewriting most tests using `for` loops on types.
https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a1cbb81223e350a23b6d03f0de9456734b88a982/test/interval_tests/bisect.jl#L1-L32
would became
```Julia
using IntervalArithmetic
using Test


@testset "`bisect` function" begin
    FloatTypes = [Float8, Float16, Float32, Float64]
    for T in FloatTypes
        X = 0..1
        @test bisect(X, T(0.5)) == (T(0)..T(0.5), T(0.5)..T(1))
        @test bisect(X, T(0.25)) == (T(0)..T(0.25), T(0.25)..T(1))

        @test bisect(X) == (interval(T(0.0), T(0.49609375)), interval(T(0.49609375), T(1.0)))

        # [...]
    end
end
```

dpsanders:
Yes that's a great idea. (Note that `Float8` doesn't exist.)

remi-garcia:
With julia 1.6.7
```julia
julia> interval(Float16(1))/interval(Float16(0.1))
ERROR: MethodError: no method matching rounding_raw(::Type{Float16})
Closest candidates are:
  rounding_raw(::Type{BigFloat}) at mpfr.jl:121
  rounding_raw(::Type{var"#s77"} where var"#s77"<:Union{Float32, Float64}) at rounding.jl:155
Stacktrace:
 [1] setrounding(f::IntervalArithmetic.var"#29#30"{Float16, Float16}, #unused#::Type{Float16}, rounding::RoundingMode{:Down})
   @ Base.Rounding ./rounding.jl:173
 [2] /(#unused#::IntervalArithmetic.IntervalRounding{:tight}, a::Float16, b::Float16, #unused#::RoundingMode{:Down})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/EquAX/src/intervals/rounding.jl:177
 [3] /(a::Float16, b::Float16, r::RoundingMode{:Down})
[...]

PR #418, opened 2020-09-23
Type consistency for ..

### LINKED ISSUE #416: Interval creation with .. returns inconsistent types
I find it strange that:
```Julia
julia> typeof(Float16(0)..Float32(0.5))
Interval{Float32}

julia> typeof(Int(0)..Float32(0.5))
Interval{Float64}
```
I would think that `typeof(Int(0)..Float32(0.5)) == Interval{Float32}`

remi-garcia:
Fix #416

### DIFF
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -144,6 +144,14 @@ function ..(a::T, b::S) where {T, S}
     interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
 end
 
+function ..(a::T, b::S) where {T<:Integer, S}
+    interval(atomic(Interval{S}, a).lo, atomic(Interval{S}, b).hi)
+end
+
+function ..(a::T, b::S) where {T, S<:Integer}
+    interval(atomic(Interval{T}, a).lo, atomic(Interval{T}, b).hi)
+end
+
 function ..(a::T, b::Irrational{S}) where {T, S}
     R = promote_type(T, Irrational{S})
     interval(atomic(Interval{R}, a).lo, R(b, RoundUp))
--- a/test/interval_tests/types.jl
+++ b/test/interval_tests/types.jl
@@ -1,0 +1,28 @@
+using IntervalArithmetic
+using Test
+
+
+@testset "types consistency" begin
+    int_types = [Int8, Int16, Int32, Int64, Int128]
+    float_types = [Float16, Float32, Float64]
+    considered_types = vcat(int_types, float_types
+    interval_functions = [.., interval, Interval]
+    for interval_constructors in interval_functions
+        for T in considered_types
+            for S in considered_types
+                @test typeof(interval_constructors(T(0), S(1)) == promote_type(T, S)
+            end
+            for S in float_types
+                @test typeof(interval_constructors(T(0), S(∞)) == promote_type(T, S)
+                @test typeof(interval_constructors(S(-∞), T(0)) == promote_type(T, S)
+                @test typeof(interval_constructors(T(0), S(0.1)) == promote_type(T, S)
+                @test typeof(interval_constructors(S(-0.1), T(0)) == promote_type(T, S)
+            end
+        end
+        for T in float_types
+            for S in float_types
+                @test typeof(interval_constructors(T(-0.1), S(0.2)) == promote_type(T, S)
+            end
+        end
+    end
+end


remi-garcia:
Should also improve Float32 support (#415)

src/intervals/intervals.jl
```diff
@@ -144,6 +144,14 @@ function ..(a::T, b::S) where {T, S}
     interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
 end
 
+function ..(a::T, b::S) where {T<:Integer, S<:AbstractFloat}
+    interval(atomic(Interval{S}, a).lo, atomic(Interval{S}, b).hi)
+end
+
+function ..(a::T, b::S) where {T<:AbstractFloat, S<:Integer}
+    interval(atomic(Interval{T}, a).lo, atomic(Interval{T}, b).hi)
```
dpsanders, line 152: Can't you just use promotion somehow?

remi-garcia: I could have use `promote_type` but in my understanding having more methods is more efficient than relying on functions that deal with types.
```Julia
function ..(a::T, b::S) where {T, S}
    R = promote_type(T, S)
    interval(atomic(Interval{R}, a).lo, atomic(Interval{R}, b).hi)
end
```
should work. Yet it probably is inefficient for `..(::Floatxx, ::Floatyy)`.

dpsanders: I'm not sure what you mean by "having more methods is more efficient than relying on functions that deal with types".  Julia compiles functions to make specialised versions corresponding to the types of the inputs.

I'm not sure why you think that this code will be inefficient.
Rather than guessing what might be inefficient, just use BenchmarkTools.jl to test the different possibilities to see which actually are faster or slower.

remi-garcia: I though that calling `promote_type` would have been slow. But, you were write:
```Julia
julia> using BenchmarkTools

julia> function foo(a::T, b::S) where {T, S}
           interval(IntervalArithmetic.atomic(Interval{T}, a).lo, IntervalArithmetic.atomic(Interval{S}, b).hi)
       end
foo (generic function with 1 method)

julia> function foo(a::T, b::S) where {T<:Integer, S<:AbstractFloat}
           interval(IntervalArithmetic.atomic(Interval{S}, a).lo, IntervalArithmetic.atomic(Interval{S}, b).hi)
       end
foo (generic function with 2 methods)

julia> function foo(a::T, b::S) where {T<:AbstractFloat, S<:Integer}
           interval(IntervalArithmetic.atomic(Interval{T}, a).lo, IntervalArithmetic.atomic(Interval{T}, b).hi)
       end
foo (generic function with 3 methods)

julia> function foo(a::T, b::Irrational{S}) where {T, S}
           R = promote_type(T, Irrational{S})
           interval(IntervalArithmetic.atomic(Interval{R}, a).lo, R(b, RoundUp))
       end
foo (generic function with 4 methods)

julia> function foo(a::Irrational{T}, b::S) where {T, S}
           R = promote_type(Irrational{T}, S)
           return Interval(R(a, RoundDown), IntervalArithmetic.atomic(Interval{R}, b).hi)
       end
foo (generic function with 5 methods)

julia> function foo(a::Irrational{T}, b::Irrational{S}) where {T, S}
           R = promote_type(Irrational{T}, Irrational{S})
           return Interval(a, b)
       end
foo (generic function with 6 methods)

julia> function bar(a::T, b::S) where {T, S}
           R = promote_type(T, S)
           interval(IntervalArithmetic.atomic(Interval{R}, a).lo, IntervalArithmetic.atomic(Interval{R}, b).hi)
       end
bar (generic function with 1 method)

julia> function bar(a::T, b::Irrational{S}) where {T, S}
           R = promote_type(T, Irrational{S})
           interval(IntervalArithmetic.atomic(Interval{R}, a).lo, R(b, RoundUp))
       end
bar (generic function with 2 methods)

julia> function bar(a::Irrational{T}, b::S) where {T, S}
           R = promote_type(Irrational{T}, S)
           return Interval(R(a, RoundDown), IntervalArithmetic.atomic(Interval{R}, b).hi)
       end
bar (generic function with 3 methods)

julia> function bar(a::Irrational{T}, b::Irrational{S}) where {T, S}
           R = promote_type(Irrational{T}, Irrational{S})
           return Interval(a, b)
       end
bar (generic function with 4 methods)

julia> @benchmark foo(2, 3)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.207 ns (0.00% GC)
  median time:      2.219 ns (0.00% GC)
  mean time:        2.307 ns (0.00% GC)
  maximum time:     21.396 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark bar(2, 3)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     2.208 ns (0.00% GC)
  median time:      2.221 ns (0.00% GC)
  mean time:        2.273 ns (0.00% GC)
  maximum time:     26.497 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

julia> @benchmark foo(2.5, 3)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     212.750 ns (0.00% GC)
  median time:      218.305 ns (0.00% GC)
  mean time:        223.686 ns (0.00% GC)
  maximum time:     434.141 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     525

julia> @benchmark bar(2.5, 3)
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     216.353 ns (0.00% GC)
  median time:      219.258 ns (0.00% GC)
  mean time:        227.008 ns (0.00% GC)
  maximum time:     503.979 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     516
```

remi-garcia:
As @dpsanders suggested, in the last commit `promote_type` is preferred over multiple methods.

I also corrected two `Interval` that, I think, should have been `interval`.

lbenet:
@remi-garcia First, sorry for our long delay to get back to this. 

Could you please rebase to master and push again?

remi-garcia:
I used the GUI "Fetch upstream". I think it went fine.

lucaferranti:
```julia
@test_logs (:warn,) @test isempty(force_interval(NaN, 3))
```

this is the test causing problem. Using `Interval` causes problem with invalid input (either bound is `NaN`).

cc @lbenet

lbenet:
@lucaferranti is right; replacing `force_interval` by `Interval` in that specific test yields an error; using `interval` in that case is fine, because it returns the empty interval; this was done #461.

I think `force_interval` is some left-over, which we replaced by the current behavior of `Interval` that does not check the correctness of the interval. Then, we can delete that function altogether and replace it simply by `Interval` when needed. The corresponding tests could also be deleted, since they test essentially the constructors.

Kolaru:
I think it is fine to remove `force_interval` altogether but it would be breaking, since `force_interval` reorder the bounds to make it valid when needed, something `Interval` (rightully) never does.

So we would be removing one feature.

lbenet:
Good point! The reason I proposed to remove it (overlooking that it reorders the bounds to make it a proper interval) is because it is only used to construct `IntervalBox`es from arrays, and it is not exported. I do not know what would be the impact of that. In any case, we can keep it, correct the problematic test, and eventually consider if we want such feaure.

lucaferranti:
I dont think it would be breaking, since it is used only internally and is not exported

Issue #376, opened 2020-05-07
methods which rely on big53() are not thread-safe


gwater:
Because `big53()` calls [`setrounding()` from `Base`](https://docs.julialang.org/en/v1/base/numbers/#Base.Rounding.setrounding-Tuple{Type,Any}) functions like `^`, `atan()` and `(:tanh, :asinh, :acosh, :atanh)` are currently not thread-safe on Intervals. Ideally all mathematical functions should be thread-safe so Intervals can also be dropped into multi-threaded code.

dpsanders:
This is a general problem with mpfr which has the precision as a global variable. 

Maybe it's finally time to make a version of BigFloat that's parameterized by the precision?

lbenet:
I agree that it would be fantastic to have all mathematical functions thread-safe, and understand the issue with `big53`; thanks for ringing this up!

I though that in order to have the correct rounding of the usual functions for `Interval{Float64}` (except for `^`) we were exploiting [CRlimb.jl](https://github.com/JuliaIntervals/CRlibm.jl). Are these functions (for `Interval{Float64}`) then thread-safe or not?

dpsanders:
The problem is that some functions. are not available in CRlibm, so we fall back to BigFloat for those.

dpsanders:
Actually I think most of those functions are expressible in terms of other functions like exp? In which case we could forget about correct rounding and just use those expressions.

lbenet:
I agree; this may  yield slightly wider intervals, but the functions will be thread-safe

dpsanders:
The `BigFloat` situation is a long-standing issue that I opened a long time ago: https://github.com/JuliaLang/julia/issues/10040

dpsanders:
As usual we need some API to specify which versions of these functions should be used. This is becoming a pressing issue.

gwater:
Please excuse me for being a bit blunt: Why do we need to keep every bit of old code around as an option through some vaporware API? This approach to maintaining the JuliaIntervals universe has been piling up unresolved issues in various packages over multiple years now. And indirectly, it has been leading to me slowly forking or reimplementing the bits that I actually want to use. Maybe using the exponentials is not the best solution but if I write a pull request that will have to wait for a nonexistent API solution to be merged, I'd rather put my time to better use. IntervalArithmetic.jl is really well validated, and it would be a shame to lose out on features like multi-threading (which I imagine more and more of the Julia ecosystem will adopt now).

[...]

PR #420, opened 2020-09-28
fix: make big53() thread safe using lock

gwater:
Base.setprecision() is currently not marked thread safe

addresses #376


gwater:
from my side this PR is complete

Kolaru:
Could you add some inline comments with maybe a reference to #376? Since thread-safety is not explicitly tested (I can understand why) I fear we might remove the lock by mistake in the future otherwise.

gwater:
Good point, I made the code a bit more expressive and added a note

dpsanders:
Thanks for this! Unfortunately there's now a conflict with another just-merged PR but I'm happy to try and fix it.

gwater:
Ok, I think the merge works now

Kolaru:
Perfect, thanks a lot!

PR #421, opened 2020-09-28
Separate implementation of sinpi / cospi that ensure that e.g. radius…

### LINKED ISSUE #412: `sinpi` returns unnecessarily wide intervals
It looks like `sinpi(x)`, where `x` is an interval, is defined internally as `sin(x*\pi)`. Unfortunately this yields intervals that could be tightened; for instance:
```
julia> sinpi(1..1)
[-3.21625e-16, 1.22465e-16]
```
while it should be possible to determine that the result is exactly 0. 

If `sin` were defined in terms of `sinpi`, instead of the opposite, with an implementation analogous to the current one, then this problem would disappear. I tried to submit a PR, but some tests fail because of a related issue I just reported in CRlibm (https://github.com/JuliaIntervals/CRlibm.jl/issues/39).

fph:
…(sinpi(0.5..0.5)) == 0. Fixes #412.

Now all tests pass, but running for instance `cospi(@biginterval(0.5))` returns a StackOverflowError on my machine (Ubuntu 20.04, Julia 1.5.0).

```julia
julia> cospi(@biginterval(0.5))
ERROR: StackOverflowError:
Stacktrace:
 [1] cospi(::BigFloat, ::RoundingMode{:Down}) at /home/sad/.julia/packages/CRlibm/NFCH5/src/CRlibm.jl:137 (repeats 79984 times)
```


fph:
This seems very weird -- the tests pass on Travis but not on Appveyor. Any idea why? Maybe it is because one of the two environments is using MPFR fallback mode and the other isn't?

dpsanders:
I'm not sure what happened but there are now lots of other commits included in this PR.
Could you please rebase against master?

fph:
Yes, sorry, I was trying to rebase today but I must have done something wrong with git. I will get the PR in order and then get back to you soon.

Issue #428, opened 2020-12-08
Exception when subtracting an Int64 scalar from a Interval{Rational{Int64}}


k-cybulski:
I encountered an exception while trying to subtract an Int64 integer from a Interval{Rational{Int64}} rational interval. REPL output below:
```julia
julia> using IntervalArithmetic

julia> x_0 = interval(1 // 3, 2 // 5)
[1//3, 2//5]

julia> x_0 - 1
ERROR: MethodError: no method matching atomic(::Type{Interval{Rational{Int64}}}, ::Interval{Rational{Int64}})
Closest candidates are:
  atomic(::Type{Interval{T}}, ::Interval) where T<:AbstractFloat at /home/user/.julia/packages/IntervalArithmetic/aGHTY/src/intervals/conversion.jl:122
  atomic(::Type{Interval{T}}, ::S) where {T<:AbstractFloat, S<:Real} at /home/user/.julia/packages/IntervalArithmetic/aGHTY/src/intervals/conversion.jl:95
  atomic(::Type{Interval{Rational{Int64}}}, ::Irrational) at /home/user/.julia/packages/IntervalArithmetic/aGHTY/src/intervals/conversion.jl:132
  ...
Stacktrace:
 [1] convert(::Type{Interval{Rational{Int64}}}, ::Interval{Rational{Int64}}) at /home/user/.julia/packages/IntervalArithmetic/aGHTY/src/intervals/conversion.jl:20
 [2] _promote at ./promotion.jl:259 [inlined]
 [3] promote at ./promotion.jl:282 [inlined]
 [4] -(::Interval{Rational{Int64}}, ::Int64) at ./promotion.jl:313
 [5] top-level scope at REPL[7]:1
```

I am on IntervalArithmetic v0.16.7. This occurred both on Julia 1.1.1 and on 1.5.1.

dpsanders:
Thanks for the report. 
The package is not really designed to support rational intervals, as you can see.

I'm guessing it should be easy to fix by following the chain of errors -- we would welcome a PR, if you would like to do so.

dpsanders:
As a workaround you can use
```
julia> x_0 - interval(1//1)
[-2//3, -3//5]
```

k-cybulski:
I see, I had a feeling this might be the case. I can't think of many practical uses for supporting rationals other than for education, e.g. learning interval arithmetic with rationals as an example, so I suppose this isn't a particularly high priority.

[...]

PR #429, opened 2020-12-09
Support Rational - Int operations

rushabh-v:
Fixes: https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/428


dpsanders:
Thanks! If you use `interval` instead of `@interval` you should be able to test equality with the rationals. `@interval` does a conversion to floats instead.

dpsanders:
Thanks!

PR #440, opened 2021-02-18
Add CRlibm-1.0.0 compatibility

c42f:
This allows IntervalArithmetic to be used with the binarybuilder version of CRlibm.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.17.7"
+version = "0.17.8"
 
 [deps]
 CRlibm = "96374032-68de-5a5b-8d9e-752f78720389"
@@ -15,7 +15,7 @@ SetRounding = "3cc68bcd-71a2-5612-b932-767ffbe40ab0"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
-CRlibm = "0.7, 0.8"
+CRlibm = "0.7, 0.8, 1.0.1"
 FastRounding = "0.2"
 Polynomials = "0.7"
 RecipesBase = "1.0"


dpsanders:
Thanks! I was labouring under the impression that I had already done this, but apparently not...

c42f:
Are these real failures due to the updated build of crlibm?

x86, julia-1.4, julia-1.5

```
minimal_log10_test: Test Failed at C:\projects\intervalarithmetic-jl\test\ITF1788_tests\libieeep1788_tests_elem.jl:3460
  Expression: log10(Interval(5.0e-324, 10.0)) == Interval(-323.3062153431158, 1.0)
   Evaluated: Interval(-323.3062153431158, 1.0000000000000002) == Interval(-323.3062153431158, 1.0)
Stacktrace:
 [1] top-level scope at C:\projects\intervalarithmetic-jl\test\ITF1788_tests\libieeep1788_tests_elem.jl:3460
 [2] top-level scope at C:\buildbot\worker\package_win32\build\usr\share\julia\stdlib\v1.5\Test\src\Test.jl:1115
 [3] top-level scope at C:\projects\intervalarithmetic-jl\test\ITF1788_tests\libieeep1788_tests_elem.jl:3446
Test Summary:      | Pass  Fail  Total
minimal_log10_test |   19     1     20
```

Nightly failures seem unrelated (I think they're https://github.com/JuliaLang/julia/issues/38427 ?)

```
fma consistency: Error During Test at C:\projects\intervalarithmetic-jl\test\interval_tests\consistency.jl:76
  Test threw exception
  Expression: fma(a, zero(a), c) == c
  could not load symbol "fesetround":
  The specified procedure could not be found. 
```

dpsanders:
Those seem to be real failures from incorrect rounding.

Clearly `log10(10.0)` should give `1.0` with correct rounding. 
`x86` means 32-bit I think? 

The weird thing is that almost all tests pass except this one. I don't know what's going on or how to fix this, except by 
reverting to using MPFR for that function.

The CRlibm library no longer seems to be under development, so there is not even an upstream to report to...

c42f:
If this was working in the previous local build (build.jl) of CRLibm and it's not working in the binary builder version, perhaps it's due to compilation flags in the code to create CRLibm_jll ?  Then perhaps we should report it to Yggdrasil?

Any thoughts @giordano ?

giordano:
> Any thoughts @giordano ?

No idea what's the issue, I guess someone needs to investigate whether the build process in BinaryBuilder did something different from what the local build used to do. The log of compilation with BinaryBuilder is in the tarball itself

c42f:
Ok thanks.

So I see that prior to https://github.com/JuliaIntervals/CRlibm.jl/pull/41, CRlibm wasn't used on windows — a warning was emitted and things fell back to MPFR.

So this may be the first time that code path has ever been tested with the windows build.

c42f:
I believe this is a bug in CRlibm on 32 bit. It's not restricted to windows. On the official linux 86 build of Julia 1.5.3, I have:

```
julia> ccall((:log10_ru, CRlibm.CRlibm_jll.libcrlibm), Float64, (Float64,), 10.0)
1.0000000000000002
```

I don't think this is due to the binarybuilder build, I expect it was just never tested before on a 32 bit system.

dpsanders:
Thanks for the detective work, @c42f! 

How do you suggest we handle this? This is from an automatically-generated test file.

c42f:
Well it depends on our appetite for debugging CRlibm ;-)

I did some extra investigation in https://github.com/JuliaIntervals/CRlibm.jl/issues/45 and found that our build of CRlibm is badly broken on 32 bit when using the obvious build options and I didn't find a simple way to fix it. My suggestion would be to fall back to MPFR on all 32 bit systems.

lucaferranti:
done in #503

Issue #442, opened 2021-03-06
remove setrounding from functions implementation


lucaferranti:
Some functions do not work on 1.6 on windows. Particularly, I found the functions `abs, mig, fma, sqrt` . The core problem seems that they use `setrounding`. `abs` calls `mig` and `mig` and `fma` use directly `setrounding` (source code [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/arithmetic.jl#L248:L301) ). `sqrt` I am not sure, [this code](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/rounding.jl#L220) does use set rounding, but the other functions defined there (inv, tanh etc.) do not fail. Here's the full stacktrace for reference
```
julia> sqrt(1..2)
ERROR: could not load symbol "fesetround":
The specified procedure could not be found.
Stacktrace:
 [1] setrounding_raw
   @ ~\.julia\packages\SetRounding\MjtxK\src\SetRounding.jl:45 [inlined]
 [2] setrounding(#unused#::Type{Float64}, r::RoundingMode{:Down})
   @ SetRounding ~\.julia\packages\SetRounding\MjtxK\src\SetRounding.jl:47
 [3] setrounding(f::IntervalArithmetic.var"#45#46"{Float64}, #unused#::Type{Float64}, rounding::RoundingMode{:Down})
   @ Base.Rounding .\rounding.jl:174
 [4] sqrt
   @ ~\.julia\packages\IntervalArithmetic\odPjL\src\intervals\rounding.jl:220 [inlined]
 [5] sqrt
   @ ~\.julia\packages\IntervalArithmetic\odPjL\src\intervals\rounding.jl:232 [inlined]
 [6] sqrt
   @ ~\.julia\packages\IntervalArithmetic\odPjL\src\intervals\rounding.jl:284 [inlined]
 [7] sqrt(a::Interval{Float64})
   @ IntervalArithmetic ~\.julia\packages\IntervalArithmetic\odPjL\src\intervals\functions.jl:233
 [8] top-level scope
   @ REPL[29]:1

julia> mig(1..2)
ERROR: could not load symbol "fesetround":
The specified procedure could not be found.
Stacktrace:
 [1] setrounding_raw
   @ ~\.julia\packages\SetRounding\MjtxK\src\SetRounding.jl:45 [inlined]
[...]

PR #443, opened 2021-03-10
remove setrounding from functions

lucaferranti:
The functions using `setrounding`  ¨~do not seem to work properly with 1.6 on windows~ now work, but it was discussed to remove setrounding from the implementation. For what I can say, the functions using setrounding are 

- [x] mig (and hence abs, which calls mig)
- [x] fma
- [ ] sqrt

Related issues: #442

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -235,21 +235,72 @@ end
 //(a::Interval, b::Interval) = a / b    # to deal with rationals
 
 
-function min_ignore_nans(args...)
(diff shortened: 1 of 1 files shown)


lucaferranti:
@dpsanders @lbenet I think this could be merged at this point and investigate the `sqrt` thingy later in a separate issue / PR, what do you think?

OlivierHnt:
This PR may have improvements from what was implemented in PR #593.

@lucaferranti what do you think? Otherwise let us close the PR.

PR #445, opened 2021-03-19
migrate to github actions

lucaferranti:


lucaferranti:
This should be ready now. One note about the documentation:

I added in the readme the link to the documentation in the webpage and disabled documenter. If wanted, I can readd documenter. The webpage is missing an article/paragraph about decorated intervals (I'll add in the upcoming days) otherwise it has all the information that the old documenter docs had.

Issue #438, opened 2021-02-05
Add `zero(X::IntervalBox)`


dpsanders:
zero(X::IntervalBox) = zero.(X)

PR #446, opened 2021-03-22
Add special interval constructors

mforets:
This PR adds special constructors for zero interval / box and symmetric box that are commonly used in other packages (such as TaylorModels.jl or ReachabilityAnalysis.jl).

### DIFF
--- a/src/intervals/special.jl
+++ b/src/intervals/special.jl
@@ -83,3 +83,11 @@ contains_zero(X::Interval{T}) where {T} = zero(T) ∈ X
 # """
 # widen(x::Interval{T}) where T<:AbstractFloat =
 #     Interval(prevfloat(x.lo), nextfloat(x.hi))
+
+"""
+    zero_interval(T)
+
+Return the zero interval `Interval(0, 0)` in the numeric type `T`.
+"""
+zero_interval(::Type{T}) where T<:Real = zero(Interval{T})
+zero_interval(x::Interval{T}) where T<:Real = zero(Interval{T})
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -140,3 +140,18 @@ end
 
 hull(a::IntervalBox{N,T}, b::IntervalBox{N,T}) where {N,T} = IntervalBox(hull.(a[:], b[:]))
 hull(a::Vector{IntervalBox{N,T}}) where {N,T} = hull(a...)
+
+"""
+    zero_box(N, T)
+
+Return the zero interval box of dimension `N` in the numeric type `T`.
+"""
+zero_box(N, ::Type{T}) where T<:Real = IntervalBox(zero_interval(T), N)
+
+"""
+    symmetric_box(N, T)
+
+Return the symmetric interval box of dimension `N` in the numeric type `T`,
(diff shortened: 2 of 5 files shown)


src/intervals/special.jl
```diff
@@ -83,3 +83,11 @@ contains_zero(X::Interval{T}) where {T} = zero(T) ∈ X
...
 # widen(x::Interval{T}) where T<:AbstractFloat =
 #     Interval(prevfloat(x.lo), nextfloat(x.hi))
+
+"""
+    zero_interval(T)
+
+Return the zero interval `Interval(0, 0)` in the numeric type `T`.
+"""
+zero_interval(::Type{T}) where T<:Real = zero(Interval{T})
+zero_interval(x::Interval{T}) where T<:Real = zero(Interval{T})
```
lbenet, line 93: Note that there exists `zero` which does precisely this; see [src/intervals/arithmetic.jl](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/b70940dabdbefd8eb4be32d7bde6df72174dd3f7/src/intervals/arithmetic.jl#L55).

src/multidim/intervalbox.jl
```diff
@@ -140,3 +140,18 @@ end
 
 hull(a::IntervalBox{N,T}, b::IntervalBox{N,T}) where {N,T} = IntervalBox(hull.(a[:], b[:]))
 hull(a::Vector{IntervalBox{N,T}}) where {N,T} = hull(a...)
+
+"""
+    zero_box(N, T)
+
+Return the zero interval box of dimension `N` in the numeric type `T`.
+"""
+zero_box(N, ::Type{T}) where T<:Real = IntervalBox(zero_interval(T), N)
```
lbenet, line 149: What about adding `zero(::Type{IntervalBox{N,T}}`?

lbenet (commented):
Thanks for bringing this functionality here, which I think is the right place to have it!.

I have left two comments related to `zero_interval`: there exists `zero` (overloaded from Base) which already has this functionality, though I think the equivalent for `IntervalBox`es doesn't exist. As for `symmetric_box`, I am in favor of having it!

src/intervals/special.jl
mforets: true. so do you suggest to remove `zero_interval(Float64)` (since it can be achieved by `zero(Interval{Float64})`?

src/multidim/intervalbox.jl
mforets: yes, we could add a new method such that eg. `zero(IntervalBox{2, Float64})` is the same as `zero_box(2, Float64)`. this is in fact open issue https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/438

mforets:
about the comments: just tell me which names you prefer so i update the branch accordingly :)

lbenet:
In my opinion, for both cases, `zero` is preferable.

mforets:
alright, the last commit updates the methods to `zero`. This should close #438.

mforets:
This PR is ready from my side, let me know if you'd like to consider further changes.

lbenet (approved):

PR #452, opened 2021-04-02
Draft of multi-intervals (unions of disjoint intervals)

AnderGray:
Are you guys interested in an implementation of Multi-Intervals?

Multi-Intervals are sets of reals defined by unions of disjoint intervals. This PR includes:

- Constructors, which remove any empty sets and envelope intervals which intersect
- Arithmetic of Multi-Intervals, including with intervals and scalars
- Complement functions

Usage:

```Julia
    julia> a = interval(0,2) ∪ interval(3,3.6) ∪ interval(3.5,4)
    [0, 2] ∪ [3, 4]

    julia> b = interval(1,2) ∪ interval(4,5) ∪ ∅
    [1, 2] ∪ [4, 5]

    julia> c = a * b 
    [0, 10] ∪ [12, 20]
    
    julia> complement(c)
    [-∞, 0] ∪ [10, 12] ∪ [20, ∞]
```

Note that the arithmetic doesn't broadcast on the elements of the multi-interval, but does every combination. The result is then condensed if any intervals overlap.

I'm agnostic to the name "Multi-Interval", it's what we call them but maybe there's a better name.

This is a draft and just includes a stand alone file which uses IntervalArithmetic.jl, and does not yet integrate it into the module.

Extension to interval boxes should also be possible.

### DIFF
--- a/src/intervals/multi_intervals.jl
+++ b/src/intervals/multi_intervals.jl
@@ -1,6 +1,6 @@
 using IntervalArithmetic
 import Base: getindex, ∪
-import Base: +, -, *, /, min, max, ^, log, <, >
+import Base: +, -, *, /, min, max, ^, log, <, >, exp, sin, cos, tan
 import IntervalArithmetic: hull
 
 """
@@ -131,6 +131,10 @@ for op in (:+, :-, :*, :/, :min, :max, :^, :log, :<, :>)
     @eval ($op)( x::IntervalM, n::Real) = intervalM(broadcast($op, x.v, n))
     @eval ($op)( n::Real, x::IntervalM) = intervalM(broadcast($op, n, x.v))
     
+end
+
+for op in (:-, :sin, :cos, :tan, :exp, :log)
+    @eval ($op)( x::IntervalM) = intervalM(broadcast($op, x.v))
 end
 
 ###
--- a/src/intervals/multi_intervals.jl
+++ b/src/intervals/multi_intervals.jl
@@ -36,9 +36,9 @@ intervalM(x :: IntervalM, y :: Interval) = intervalM([x.v; y])
 intervalM(x :: IntervalM, y :: IntervalM) = intervalM([x.v; y.v])
 ∪(x :: IntervalM, y :: IntervalM) = intervalM(x,y)
 
-# MultiInterval can act like a vector
+# Get index from multi-interval array
 getindex(x :: IntervalM, ind :: Integer) = getindex(x.v,ind)
-getindex(x :: IntervalM, ind :: Array{ <: Integer}) = getindex(x.v,ind)
+getindex(x :: IntervalM, ind :: Array{ <:Integer}) = getindex(x.v,ind)
 
 # Remove ∅ from Multi-Interval
 function removeEmpties(x :: IntervalM)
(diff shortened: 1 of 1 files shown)


dpsanders:
This is a great idea; in fact I had an implementation somewhere.

I actually think this belongs in a different package, though. Personally I would call this `IntervalUnion`, but I quite like `MultiInterval`.
Have you seen the papers by Schichl et al. on this stuff?

AnderGray:
Sure, maybe it's better standalone!

Mincing and boxes would have to be figured out, as well as subset-hood and stuff. But I don't think it's difficult.
I haven't, can you send a reference?

dpsanders:
https://www.researchgate.net/publication/309170886_Interval_unions

https://www.researchgate.net/publication/316372412_Using_interval_unions_to_solve_linear_systems_of_equations_with_uncertainties

Kolaru:
I think it is definitely very cool. Please let us know here if you decide to turn that into its own package, especially if you would like some feedbacks.

Here are few of them already:
- The Julia standard style is to use `snake_case` for function names and `CamelCase` for types.
- Performance may become a real issue quickly when the number of intervals in the union grows (I expect exponential growth in the number of intervals in some cases). You may want to consider strategies to deal with that early in the developpment process (e.g. limiting the number of intervals in the union).

dpsanders:
Here's another relevant paper, where they try to deal with the point raised by @Kolaru by introducing criteria for joining together boxes in the union:

https://link.springer.com/chapter/10.1007/978-3-030-31041-7_14

AnderGray:
Thanks for the references and feedback. I'll have a read through them. I'm thinking I like Interval Unions more, since they're called this in the lit.

- Cheers, I wasn't aware that was the style! I'll make the changes
- Definitely. My first thought was finding which intervals are closest together (complements with the smallest widths) and enveloping them until a max number is reached. The suggestion by @dpsanders is probably more sophisticated  

Again agnostic about it being it's own package. 

Pros: Users of IntervalArithmetic don't need it in generally. And it would be easier to manage if it was stand alone.

Cons: I can think of two situations where you get multi-intervals from standard intervals:

- sqrt
- Conditional statements with intervals

For example consider the following function:

```Julia
function interval_ifs(A)
    if A < 3
        A = A + 2 
    else
        A = A - 2
    end
    return A
end

A = interval(0,5)
interval_ifs(A)
```

I believe that the answer should be `[-2, 1] U [5, 7]`.

In cases with uncertain conditional statements, I think the interval should be bisected at 3, with each part going down different branches and then unioned afterwards. Similar for while and for loops.

Although we can't do this yet, I think with Julia's metaprogamming we should be able to do some nice stuff. Maybe even with Intermediate representation: https://github.com/FluxML/IRTools.jl

IntervalUnions would play a role I think, but you could also envelop the union and get `[-2, 7]` and stay with intervals

dpsanders:
We definitely do not want to mix up interval unions with normal intervals. You should have to explicitly start with an interval union to explicitly allow interval unions in the result. So it makes sense for it to be a separate package.

AnderGray:
I agree. Thanks!

AnderGray:
[IntervalUnionArithmetic.jl](https://github.com/AnderGray/IntervalUnionArithmetic.jl) has been created

PR #453, opened 2021-04-03
Improve support for `Interval{<:Integer}`

timholy:
When intervals are components of bigger `struct`s that impose
`eltype`-consistency, it would be nice not to have to force the
`eltype` to be an `AbstractFloat`. It's been possible to construct
`Interval{<:Integer}` for some time, but many methods throw errors.
This reduces the fragility of the package, hopefully without introducing
new numeric errors.

This also takes the liberty of bumping the patch version in hopes
of getting this into a release, but I'm happy to remove that if you'd
prefer to manage that separately.

### DIFF
--- a/src/intervals/precision.jl
+++ b/src/intervals/precision.jl
@@ -21,10 +21,12 @@ function bigequiv(a::Interval{T}) where {T <: AbstractFloat}
         return atomic(Interval{BigFloat}, a)
     end
 end
+bigequiv(a::Interval{<:Integer}) = Interval(BigFloat(a.lo), BigFloat(a.hi))
 
 bigequiv(x::AbstractFloat) = guarded_setprecision(precision(x)) do
     return BigFloat(x)
 end
+bigequiv(x::Integer) = big(x)
 
 # NOTE: prevents multiple threads from calling setprecision() concurrently
 const precision_lock = ReentrantLock()
--- a/test/interval_tests/numeric.jl
+++ b/test/interval_tests/numeric.jl
@@ -6,12 +6,12 @@ using Test
 setprecision(Interval, Float64)
 # # setrounding(Interval, :narrow)
 
-@testset "brodcasting tests" begin
+@testset "broadcasting tests" begin
     x = 3..12
     y = 4..5
     a = 3
     b = 12
-    
+
     @test sqrt(sum(x.^2 .+ y.^2)) == 5..13
 
     for i in 1:20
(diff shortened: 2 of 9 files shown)


src/intervals/precision.jl
```diff
@@ -21,10 +21,12 @@ function bigequiv(a::Interval{T}) where {T <: AbstractFloat}
         return atomic(Interval{BigFloat}, a)
     end
 end
+bigequiv(a::Interval{<:Integer}) = Interval(big(a.lo), big(a.hi))
```
timholy, line 24: This is the change I'm least sure of. This seems safer than `bigequiv(float(a))` but of course it will be much slower.

dpsanders:
Thanks for this @timholy, and apologies for the delay.

We had made a deliberate choice near the start of writing the package not to support intervals of integers, but if it's easy to add then that seems like a good idea. I'm interested in what the use case would be.

I'll try to look at the details soon.

src/intervals/precision.jl
```diff
@@ -21,10 +21,12 @@ function bigequiv(a::Interval{T}) where {T <: AbstractFloat}
         return atomic(Interval{BigFloat}, a)
     end
 end
+bigequiv(a::Interval{<:Integer}) = Interval(BigFloat(a.lo), BigFloat(a.hi))
```
dpsanders, line 24: Is there a reason for not just using `big`?

test/interval_tests/numeric.jl
```diff
@@ -173,8 +187,11 @@ setprecision(Interval, Float64)
 
     @test exp2(@biginterval(1//2)) ⊆ exp2(@interval(1//2))
     @test exp2(Interval(1024.0)) == Interval(1.7976931348623157e308, Inf)
+    @test exp2(Interval{Int}(2, 3)) == Interval{Int}(4, 8)
     @test exp10(@biginterval(1//2)) ⊆ exp10(@interval(1//2))
     @test exp10(Interval(308.5)) == Interval(1.7976931348623157e308, Inf)
+    @test exp10(Interval{Int}(2, 3)) == Interval{Int}(100, 1000)
```
dpsanders, line 193: Doesn't `exp10` usually return floats?

dpsanders (commented):
In general this looks very nice, thanks!

dpsanders:
@timholy: I think `bigequiv(x)` is the same as `big(float(x))` so I don't think we need to export that function?

PR #454, opened 2021-04-12
Avoid NaN caused by using rationalize within atomic

michaellass:
The `@interval` macro can produce `[NaN, NaN]` in case `rationalize` chooses a denominator that does not fit into the datatype used for the boundaries. Example:

```

julia> setprecision(Interval, Float16, 53)
53

julia> @interval(1.00001)
[NaN, NaN]
```

This pull request works around this by adding a corresponing check to `atomic`. Above example is added as a unit test. New behavior:

```

julia> setprecision(Interval, Float16, 53)
53

julia> @interval(1.00001)
[1, 1.00098]
```

### DIFF
--- a/src/intervals/conversion.jl
+++ b/src/intervals/conversion.jl
@@ -106,6 +106,9 @@ function atomic(::Type{Interval{T}}, x::S) where {T<:AbstractFloat, S<:AbstractF
 
     xrat = rationalize(x)
 
+    # Prevent generating NaN if denominator is too large
+    isinf(T(xrat.den)) && return Interval{T}( T(x, RoundDown), T(x, RoundUp) )
+
     # This prevents that xrat returns a 0//1 when x is very small
     # or 1//0 when x is too large but finite
     if (x != zero(x) && xrat == 0) || isinf(xrat)
--- a/test/interval_tests/construction.jl
+++ b/test/interval_tests/construction.jl
@@ -186,6 +186,9 @@ const eeuler = Base.MathConstants.e
     setprecision(Interval, 53)
     a = big(1)//3
     @test @interval(a) == Interval(big(3.3333333333333331e-01), big(3.3333333333333337e-01))
+
+    setprecision(Interval, Float16, 53)
+    @test @interval(1.00001) == Interval{Float16}(1, 1.00098)
 end
 
 @testset "Big intervals" begin


dpsanders:
Thanks and sorry for the delay.

We have moved away from using `@interval` with a global precision mode.
Is this problem reproducible using e.g. `..`?

michaellass:
As far as I can tell, it's only a problem with the `@interval` macro:

```

julia> setprecision(Interval, Float16, 53)
53

julia> @interval(1.00001)
[NaN, NaN]

julia> interval(1.00001)
[1.00001, 1.00002]

julia> Interval(1.00001)
[1.00001, 1.00002]

julia> 1.00001..1.00001
[1, 1.00002]

julia> 1.00001 ± 0
[1, 1.00002]
```

Is moving away from `@interval` an ongoing process? I'm new to the package and after reading the tutorial, it seemed to be one of the preferred ways to define intervals.

lbenet:
Be aware that `interval` and `Interval` usually default to `Float64`. Using your example:
```julia
julia> @interval(1.00001)
[NaN, NaN]

julia> typeof(ans)
Interval{Float16}

julia> interval(1.00001)
[1.00001, 1.00002]

julia> typeof(ans)
Interval{Float64}
```
The other examples you quote produce `Interval{Float64}` as well.

dpsanders:
Which tutorial did you read? 
`@interval` has not been the preferred way for quite a while, but the documentation has not kept up.

The preferred way to create an interval is

`1.1..2.2` or `interval(1.1, 2.2)`

dpsanders:
Also, `Interval{Float16}` is probably not well supported yet, and indeed `Float16` is not well supported in Julia. You are probably better off using `Interval{Float32}`

lbenet:
Not sure if the following is helpful:
```julia
julia> interval(prevfloat(Float16(1.0)), nextfloat(Float16(1.0)))
Interval(0.9995, 1.001)

julia> typeof(ans)
Interval{Float16}
```

lucaferranti:
at least the docstring is outdated
```
julia> @doc @interval
  The @interval macro is the main method to create an interval. It converts each expression into a narrow interval
  that is guaranteed to contain the true value passed by the user in the one or two expressions passed to it. When
  passed two expressions, it takes the hull of the resulting intervals to give a guaranteed containing interval.

  Examples:

      @interval(0.1)

      @interval(0.1, 0.2)

      @interval(1/3, 1/6)

      @interval(1/3^2)
```

The [current tutorial](https://juliaintervals.github.io/pages/tutorials/tutorialArithmetic/) does not define `@interval` as preferred, but it does not emphasize that `..` is preferred either, so I guess it might be misleading.

michaellass:
It's not really described as being preferred. However, there are lots of advantages described in the tutorial so my takeaway was "just use that variant":
https://juliaintervals.github.io/pages/tutorials/tutorialArithmetic/#interval_macro_and_interval_method

My experiments are maybe a bit off from the typical use of this package. I deliberately wanted to observe the impact of rounding in low precision types like Float16. I can also do that using BigFloat and setting the number of mantissa bits to 11 though. Also the suggestion by @lbenet makes sense.

So maybe this is mainly a documentation issue. But as long as `@interval` is available, it can also try to avoid this corner case.

lucaferranti:
You are right that that part is misleading, I opened an issue in the site repository [here](https://github.com/JuliaIntervals/juliaintervals.github.io/issues/7). I was planning to restructure that tutorial anyway as it's quite overwhelming at the moment, hopefully after that it will be clearer. If you found other parts unclear, feel free to comment on that issue or open a new one.

michaellass:
Closing since use of the `@interval` macro is discouraged.

PR #458, opened 2021-04-30
fix dist for decorated intervals

lucaferranti:
`dist` for decorated intervals was accidentally defined as unary function [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/decorations/functions.jl#L49-L55)

before:
```
julia> a = @decorated 1 2
[1, 2]_com

julia> b = @decorated 3 4
[3, 4]_com

julia> dist(a, b)
ERROR: MethodError: no method matching dist(::DecoratedInterval{Float64}, ::DecoratedInterval{Float64})
Stacktrace:
 [1] top-level scope
   @ REPL[13]:1
```

after:
```
julia> dist(a, b)
2.0
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.18.1"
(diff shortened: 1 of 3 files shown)


lucaferranti:
this should be ready to go 😄

dpsanders:
LGTM. Please bump the patch version, merge and release. Thanks!

PR #460, opened 2021-05-05
update interval-macro docstring

lucaferranti:
The docstring of `@interval` is outdated as it still states that the macro is the main way to create an interval. This can sometimes be misleading as pointed out in #454 .


lucaferranti:
taking care of this in #461

Issue #457, opened 2021-04-23
Should Interval(Inf, Inf) give the empty set?


dpsanders:
I think `Interval(Inf, Inf)` should give the empty set, rather than throwing an error as it does now.

dpsanders:
cc @Kolaru, @lbenet

Kolaru:
I can't see a scenario where an empty interval would give wrong result, so it looks fine to me. And the standard doesn't say much about it (p. 33):

![image](https://user-images.githubusercontent.com/17861033/116402811-21438580-a82d-11eb-96b5-ced38ccaa21f.png)

lbenet:
The following example may show that `Interval(Inf, Inf)` is tricky as a replacement for the empty set.
```julia
julia> Interval(1,2) < Interval(Inf, Inf)
true
```
I think the standard dictates that the result should be `false`, as it currently does:
```julia
julia> emptyinterval() < Interval(1,2)
false

julia> Interval(1,2) < emptyinterval()
false
```

lucaferranti:
If I understand correctly, the idea was that `Interval(Inf, Inf)` would _return_ an empty interval and not represent it.

At the moment the `Interval` constructor (capital i) does not perform any check, so all kind of prohibited intervals are allowed

```
julia> Interval(Inf, Inf)
[∞, ∞]

julia> Interval(2, 1)
[2, 1]
```
versus
```
julia> interval(Inf, Inf)
ERROR: ArgumentError: `[Inf, Inf]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
Stacktrace:
 [1] interval(a::Float64, b::Float64)
   @ IntervalArithmetic ~\.julia\packages\IntervalArithmetic\zFxqb\src\intervals\intervals.jl:111
 [2] top-level scope
   @ REPL[27]:1

julia> interval(2, 1)
ERROR: ArgumentError: `[2, 1]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
Stacktrace:
 [1] interval(a::Int64, b::Int64)
   @ IntervalArithmetic ~\.julia\packages\IntervalArithmetic\zFxqb\src\intervals\intervals.jl:111
 [2] top-level scope
   @ REPL[28]:1
```

[...]

PR #461, opened 2021-05-15
Lf/invalid as empty

lucaferranti:
It seems that the standard requires invalid inputs to return an empty interval (see discussion in #457 ), this is also the octave behavior. 

Now the package returns an empty interval but also prints a warning (like octave does), a few examples

````
julia> 3..1
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:107
∅

julia> interval(Inf, Inf)
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:107
∅

julia> interval(NaN, π)
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:107
∅

julia> @decorated 3 1
[NaN, NaN]_ill

julia> @decorated NaN 3
[NaN, NaN]_ill

julia> @decorated Inf Inf
[NaN, NaN]_ill

julia> @decorated Inf -Inf
∅_trv
```

### DIFF
--- a/src/decorations/intervals.jl
+++ b/src/decorations/intervals.jl
@@ -47,7 +47,7 @@ DecoratedInterval(I::Interval{T}, d::DECORATION) where T<:AbstractFloat =
     DecoratedInterval{T}(I, d)
 
 function DecoratedInterval(a::T, b::T, d::DECORATION) where T<:Real
-    a > b && return nai(T)
+    is_valid_interval(a, b) || return nai(T)
     DecoratedInterval(Interval(a,b), d)
 end
 
@@ -63,7 +63,7 @@ DecoratedInterval(a::T, b::S, d::DECORATION) where {T<:Real, S<:Real} =
 DecoratedInterval(I::Interval) = DecoratedInterval(I, decoration(I))
 
 function DecoratedInterval(a::T, b::T) where T<:Real
-    a > b && return nai(T)
+    is_valid_interval(a, b) || return nai(T)
     DecoratedInterval(Interval(a,b))
 end
 
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -79,11 +79,7 @@ function is_valid_interval(a::Real, b::Real)
     # println("isvalid()")
 
     if isnan(a) || isnan(b)
-        if isnan(a) && isnan(b)
-            return true
-        else
-            return false
-        end
+        return false
     end
 
     if a > b
@@ -106,9 +102,10 @@ end
 
 `interval(a, b)` checks whether [a, b] is a valid `Interval`, which is the case if `-∞ <= a <= b <= ∞`, using the (non-exported) `is_valid_interval` function. If so, then an `Interval(a, b)` object is returned; if not, then an error is thrown.
 """
-function interval(a::Real, b::Real)
+function interval(a::T, b::S) where {T<:Real, S<:Real}
     if !is_valid_interval(a, b)
-        throw(ArgumentError("`[$a, $b]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`."))
+        @warn "Invalid input, empty interval is returned"
+        return emptyinterval(promote_type(T, S, Float64))
     end
 
     return Interval(a, b)
--- a/test/decoration_tests/decoration_tests.jl
+++ b/test/decoration_tests/decoration_tests.jl
@@ -54,6 +54,15 @@ let b
     b = @decorated 3 4
 
     @test dist(a, b) == 2.0
+
+    # invalid input
+    @test isnai(@decorated(3, 1, com))
+    @test isnai(@decorated(3, 1))
+    @test isnai(@decorated(Inf, Inf))
+    @test isnai(@decorated(-Inf, -Inf))
+    @test isnai(@decorated(NaN, 3))
+    @test isnai(@decorated(3, NaN))
+    @test isnai(@decorated(NaN, NaN))
 end
 
 @testset "Convert string to DecoratedInterval" begin
--- a/test/interval_tests/consistency.jl
+++ b/test/interval_tests/consistency.jl
@@ -380,10 +380,10 @@ setprecision(Interval, Float64)
         @test interval(1, 2) == Interval(1, 2)
 
         @test inf(Interval(3, 2)) == 3
-        @test_throws ArgumentError interval(3, 2)
+        @test isempty(interval(3, 2))
 
         @test sup(Interval(Inf, Inf)) == Inf
-        @test_throws ArgumentError interval(Inf, Inf)
+        @test isempty(interval(Inf, Inf))
 
     end
 
--- a/test/interval_tests/construction.jl
+++ b/test/interval_tests/construction.jl
@@ -69,20 +69,20 @@ const eeuler = Base.MathConstants.e
 
     # Disallowed conversions with a > b
 
-    @test_throws ArgumentError interval(2, 1)
-    @test_throws ArgumentError interval(big(2), big(1))
-    @test_throws ArgumentError interval(BigInt(1), 1//10)
-    @test_throws ArgumentError interval(1, 0.1)
-    @test_throws ArgumentError interval(big(1), big(0.1))
+    @test isempty(interval(2, 1))
+    @test isempty(interval(big(2), big(1)))
+    @test isempty(interval(BigInt(1), 1//10))
+    @test isempty(interval(1, 0.1))
+    @test isempty(interval(big(1), big(0.1)))
 
-    @test_throws ArgumentError @interval(2, 1)
-    @test_throws ArgumentError @interval(big(2), big(1))
-    @test_throws ArgumentError @interval(big(1), 1//10)
-    @test_throws ArgumentError @interval(1, 0.1)
-    @test_throws ArgumentError @interval(big(1), big(0.1))
-    @test_throws ArgumentError interval(Inf)
-    @test_throws ArgumentError interval(-Inf, -Inf)
-    @test_throws ArgumentError interval(Inf, Inf)
+    @test isempty(@interval(2, 1))
+    @test isempty(@interval(big(2), big(1)))
+    @test isempty(@interval(big(1), 1//10))
+    @test isempty(@interval(1, 0.1))
+    @test isempty(@interval(big(1), big(0.1)))
+    @test isempty(interval(Inf))
+    @test isempty(interval(-Inf, -Inf))
+    @test isempty(interval(Inf, Inf))
 
     # Conversion to Interval without type
     @test convert(Interval, 1) == Interval(1.0)
@@ -237,10 +237,10 @@ end
     a = big(0.1)..2
     @test typeof(a) == Interval{BigFloat}
 
-    @test_throws ArgumentError 2..1
-    @test_throws ArgumentError π..1
-    @test_throws ArgumentError π..eeuler
-    @test_throws ArgumentError 4..π
+    @test isempty(2..1)
+    @test isempty(π..1)
+    @test isempty(π..eeuler)
+    @test isempty(4..π)
     @test 1..π == Interval(1, π)
 end
 
@@ -292,9 +292,10 @@ end
 end
 
 # issue 192:
-@testset "Disallow a single NaN in an interval" begin
-    @test_throws ArgumentError interval(NaN, 2)
-    @test_throws ArgumentError interval(Inf, NaN)
+@testset "Disallow NaN in an interval" begin
+    @test isempty(interval(NaN, 2))
+    @test isempty(interval(Inf, NaN))
+    @test isempty(interval(NaN, NaN))
 end
 
 # issue 206:
@@ -342,7 +343,7 @@ end
 end
 
 @testset "a..b with a > b" begin
-    @test_throws ArgumentError 3..2
+    @test isempty(3..2)
 end
 
 @testset "Hashing of Intervals" begin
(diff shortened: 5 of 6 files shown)


src/intervals/intervals.jl
```diff
@@ -106,9 +102,10 @@ end
 
 `interval(a, b)` checks whether [a, b] is a valid `Interval`, which is the case if `-∞ <= a <= b <= ∞`, using the (non-exported) `is_valid_interval` function. If so, then an `Interval(a, b)` object is returned; if not, then an error is thrown.
 """
-function interval(a::Real, b::Real)
+function interval(a::T, b::S) where {T<:Real, S<:Real}
     if !is_valid_interval(a, b)
-        throw(ArgumentError("`[$a, $b]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`."))
+        @warn "Invalid input, empty interval is returned"
+        return emptyinterval(promote_type(T, S, Float64))
```
lucaferranti, line 108: This causes problems for invalid input of rationals, as the output would then be `Interval{Float64}` instead of `Interval{Rational}`, the `Float64` is there to avoid problems when both inputs are integers, but I guess that's not the smartest way to handle it

src/decorations/intervals.jl
```diff
@@ -47,7 +47,7 @@ DecoratedInterval(I::Interval{T}, d::DECORATION) where T<:AbstractFloat =
     DecoratedInterval{T}(I, d)
 
 function DecoratedInterval(a::T, b::T, d::DECORATION) where T<:Real
-    a > b && return nai(T)
```
lucaferranti, line 50: the `nai(T)` does not work if `T` is rational, so e.g. `DecoratedInterval(3//1, 2//1)` throws an uncatched exception
This is also in the current master version

lucaferranti:
is it possible to suppress warnings during tests? I guess printing all the warnings is not very nice/useful?

src/intervals/intervals.jl
lbenet, line 108: The idea of having `interval` was to check the validity of the construction, which does not happen if you use `Interval`. This yields some speed-ups since the `is`s are avoided. While I think it is ok to return a `nai` for `DecoratedInterval`s, I think that we should return an error for bare intervals.

lbenet: What about defining `Interval(1//0, -1//0)` as the `nai` for rational intervals?

lucaferranti: `Interval` still does not make any check, so e.g. `Interval(3, 1)` works like before. The validity check is only in `interval` (and in the methods that depend on it) so for example
before:
```
 julia> Interval(3, 1)
[3, 1]

julia> interval(3, 1)
ERROR: ArgumentError: `[3, 1]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
Stacktrace:
 [1] interval(a::Int64, b::Int64)
   @ IntervalArithmetic ~\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:111
 [2] top-level scope
   @ REPL[60]:1

```

after:
```
julia> Interval(3, 1)
[3, 1]

julia> interval(3, 1)
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\intervals.jl:107
∅
```

dpsanders:
> is it possible to suppress warnings during tests? I guess printing all the warnings is not very nice/useful?

Looks like you can test the warning using `@test_logs`

src/intervals/intervals.jl
lbenet: That's precisely my point: I think using `interval` should throw an error, instead of a warning, as it is proposed. This behavior could be changed for a specific "flavor".

dpsanders:
Apparently the standard requires an empty interval to be returned when
constructing incorrect intervals. There could maybe be a flag passed to
`interval` yo specify the behaviour, but we would need to benchmark that.

src/intervals/intervals.jl
lucaferranti: The motivation for returning an empty interval was the standard extract posted in https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/457#issuecomment-830855196

My understanding from that was that it should return an empty interval and signal the behavior (print the warning). Mathematically the set `{x in R | 2<=x<=1}` is the empty set, so I see no reason why `2..1` should not return an empty interval. 

I guess throwing an error might be more useful for debugging though. If somewhere in the code you try to build an invalid interval, interrupting the computation and pointing to the line where this happened would be more useful than silently constructing an empty interval and let it propagate through all computations.

lucaferranti: This refers to my code review below?
Anyway, I think this would work because `nai` is only allowed as decorated interval, so you can distinguish `[1//0, -1//0]_trv` (the empty interval) from `[1//0, -1//0]_ill` (the NaI). This might require manually adjusting a couple of functions, but other then that it should work. Anyway, maybe this deserves its own issue/PR for further discussion?

lbenet: Sorry for misplacing my comment. Rather than manually requiring some adjustments, dispatching for `Interval{Rational{T}}` is also perhaps a way to proceed.

lucaferranti: I also opened an issue #462

src/decorations/intervals.jl
lucaferranti: that's quite beyond the scope of this PR, see #462

lucaferranti:
The current implementation of `..` was calling `atomic` before checking the validity of the input, the commit above this message fixes it. 

In the file `ieee1788-exceptions.itl` of the ITL testsuite there is a test
```
b-numsToInterval +infinity -infinity = [empty] signal UndefinedOperation;
```

so basically `interval(Inf, -Inf)` should return the empty interval **and** flag a warning.I modified the definition of `is_valid_interval` to be conformant to this, but personally I think printing the warning in this case is a little redundant, since `[Inf, -Inf]` is the representation of the empty interval, opinions on this?

src/intervals/intervals.jl
```diff
@@ -141,21 +133,32 @@ include("complex.jl")
...
+    end
+    Interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
 end
 
 function ..(a::T, b::Irrational{S}) where {T, S}
+    if !is_valid_interval(a, b)
+        @warn "Invalid input, empty interval is returned"
+        return emptyinterval(promote_type(T, Irrational{S}, Float64))
```
lbenet, line 146: Possible type instability as above; if `a=Float32(Inf)`, this will return `emptyinterval(Float64)` (instead of `emptyinterval(Float32)`).

src/intervals/intervals.jl
```diff
@@ -141,21 +133,32 @@ include("complex.jl")
...
+    end
     R = promote_type(T, Irrational{S})
-    interval(atomic(Interval{R}, a).lo, R(b, RoundUp))
+    Interval(atomic(Interval{R}, a).lo, R(b, RoundUp))
 end
 
 function ..(a::Irrational{T}, b::S) where {T, S}
+    if !is_valid_interval(a, b)
+        @warn "Invalid input, empty interval is returned"
+        return emptyinterval(promote_type(Irrational{T}, S, Float64))
```
lbenet, line 155: Possible type instability as above

lbenet (commented):
I've added few comments, some may be related to a possible type instability, which should be easy to check.

src/intervals/intervals.jl
```diff
@@ -79,20 +80,10 @@ function is_valid_interval(a::Real, b::Real)
...
-        if isinf(a) && isinf(b)
-            return true  # empty interval = [∞,-∞]
-        else
-            return false
-        end
-    end
+    a > b && return false
 
     if a == Inf || b == -Inf
         return false
```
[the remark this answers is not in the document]
lucaferranti: the if block is for the cases when both `a` and `b` are `Inf` or `-Inf`

src/intervals/intervals.jl
```diff
@@ -141,21 +133,32 @@ include("complex.jl")
 # Syntax for intervals
 
 function ..(a::T, b::S) where {T, S}
-    interval(atomic(Interval{T}, a).lo, atomic(Interval{S}, b).hi)
+    if !is_valid_interval(a, b)
+        @warn "Invalid input, empty interval is returned"
+        return emptyinterval(promote_type(T, S, Float64))
```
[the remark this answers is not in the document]
lucaferranti: true that! When I wrote that line, I also thought it looks very wrong. The reason was that if both `a` and `b` are integers and invalid (e.g. a > b), then the function would have called `emptyinterval(Int)`, which would have led to an error. I now fixed this by defining
`emptyinterval(::Type{<:Integer}) = emptyinterval(Float64)`

src/intervals/intervals.jl
```diff
@@ -25,7 +25,8 @@ struct Interval{T<:Real} <: AbstractInterval{T}
                 new(a, b)
 
             else
-                throw(ArgumentError("Interval of form [$a, $b] not allowed. Must have a ≤ b to construct interval(a, b)."))
+                @warn "Invalid input, empty interval is returned"
+                return new(T(Inf), T(-Inf))
```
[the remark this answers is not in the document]
lucaferranti: wouldn't that cause a stack overflow? the inner constructor would call `emptyinterval()` which would call the inner constructor, which would call `emptyinterval()` etc. etc.

btw, are `T(Inf)` and `convert(T, Inf)` equivalent? preferences on which one to use?

src/intervals/intervals.jl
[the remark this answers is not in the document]
lbenet: I guess the only case (i can think now) where the `if` block does something is for `a == b == Inf` or `a == b == -Inf`, where the conditional at line 86 would not be considered. So, let's leave it there.

src/intervals/intervals.jl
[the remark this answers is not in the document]
lbenet: Good point about the stack overflow!

Regarding your question, I have no special preference.

src/intervals/intervals.jl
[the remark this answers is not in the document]
lbenet: I though that you thought about that case, too :-) Yet, I guess an outer constructor would be called before, and then conversion to `Float64` or something else would take place. Is this correct? Have you checked it?

lbenet (approved):
LGTM!

lucaferranti:
Great! added a couple of more tests and as a far as I'm concerned, this is ready for merge now. This `fixes` step 6 of #465 . 
I am next going to fix the constructors for decorated intervals (steps 4 and 10 of #465 ) and next investigate the parsing of strings.

Should we already bump the version and release or wait a few days for the upcoming PRs?

src/intervals/intervals.jl
```diff
@@ -25,7 +25,8 @@ struct Interval{T<:Real} <: AbstractInterval{T}
                 new(a, b)
 
             else
-                throw(ArgumentError("Interval of form [$a, $b] not allowed. Must have a ≤ b to construct interval(a, b)."))
+                @warn "Invalid input, empty interval is returned"
```
Kolaru, line 28: Would it be reasonnable to never check the validity here, get rid of the `validity_check` setting and document that to get IEEE compliant and thouroughly validated interval once should use either `interval` or `DecoratedInterval`?

Related but not directly in the scope of this PR, I wonder if it wouldn't be clearer for users to have `Interval` as the default IEEE compliant one and use some trick to get an `unsafe_interval` function for internal operations.

If you thing it is too tangential to the issue tackled here, I can add a separate issue.

src/intervals/intervals.jl
```diff
@@ -106,9 +97,10 @@ end
 
 `interval(a, b)` checks whether [a, b] is a valid `Interval`, which is the case if `-∞ <= a <= b <= ∞`, using the (non-exported) `is_valid_interval` function. If so, then an `Interval(a, b)` object is returned; if not, then an error is thrown.
 """
-function interval(a::Real, b::Real)
+function interval(a::T, b::S) where {T<:Real, S<:Real}
```
Kolaru, line 100: This docstring is no longer correct, it never throws an error.

Also precising that this is the way of creating IEEE compliant interval would be great.

lucaferranti: right! I had a separated PR (#460 ) for the docstrings which I had opened before this. The idea was to rebase that once this is merged and fix the docstrings there but indeed... it makes no sense! I'll close that PR and take care of the PRs here :D

src/intervals/intervals.jl
lucaferranti: `validity_check` is indeed set by default to `false` and since you have functions like `..` and `interval` I think it could be removed without too much harm.

The `Interval` vs `interval` might be an interesting discussion. One thing that puzzles me is the asymmetry of bare and decorated intervals, in the sense that for bare the constructors `Interval` doesn't do any checks and you should use `interval` or `..`, but for decorated intervals the constructors `DecoratedIntervals` does checks (well a few issues there, next PR on the way for that :D ). I think it might be important to have a mechanism to construct "unsafe intervals" which merely puts the given inputs as lower and upper bound, at least for testing purposes. Maybe this can be addressed in a separate issue?

src/intervals/intervals.jl
lucaferranti: updated the docstring. I didn't include the note about IEEE because I think the note about IEEE compliance may be more appropriate after the package is actually IEEE compliant

Kolaru: I think it is fine like that. This will be discussed in #468 where will have to sort out what we want each constructor to do anyway.

Even if the full package is not yet compliant, I think it will be good to declare our intent on which constructor corresponds to the IEEE constructors.

Kolaru:
Thanks a lot!

PR #467, opened 2021-05-29
Improve mince and add method for non-uniform partition of IntervalBox

### LINKED ISSUE #444: Generalize IntervalBox split for non-uniform partitions
Currently, `mince` for interval boxes receives an integer `n` to divide the box n times along each dimension,

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/b70940dabdbefd8eb4be32d7bde6df72174dd3f7/src/multidim/intervalbox.jl#L124

Sometimes it is useful to have a function that accepts a non-uniform partition, say `[2, 3]` to split the first dimension two times and the second dimension 3 times, obtaning 6 boxes. (This is what `split(::AbstractHyperrectangle, block::AbstractVector)` does.)

lbenet:
Closes #444

### DIFF
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -577,5 +577,5 @@ as a vector.
 """
 function mince(x::Interval, n)
     nodes = range(x.lo, x.hi, length = n+1)
-    return [Interval(nodes[i], nodes[i+1]) for i in 1:length(nodes)-1]
+    return @. Interval(nodes[1:n], nodes[2:n+1])
 end
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -575,7 +575,7 @@ convert(::Type{Integer}, a::Interval) = isinteger(a) ?
 Splits `x` in `n` intervals of the same diameter, which are returned
 as a vector.
 """
-function mince(x::Interval, n)
+@inline function mince(x::Interval, n)
     nodes = range(x.lo, x.hi, length = n+1)
     return Interval.(nodes[1:n], nodes[2:n+1])
 end
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -577,5 +577,5 @@ as a vector.
 """
 function mince(x::Interval, n)
     nodes = range(x.lo, x.hi, length = n+1)
-    return @. Interval(nodes[1:n], nodes[2:n+1])
+    return Interval.(nodes[1:n], nodes[2:n+1])
 end
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -141,8 +141,7 @@ end
 @generated function mince(x::IntervalBox{N,T}, tup::NTuple{N,Int}) where {N,T}
     quote
         n = maximum(tup)
-        nodes_matrix = Array{Interval{T},2}(undef, n, N)
-        fill!(nodes_matrix, emptyinterval(T))
+        nodes_matrix = fill(emptyinterval(T), n, N)
         for i in 1:N
             nodes_matrix[1:tup[i],i] .= mince(x[i], tup[i])
         end
--- a/src/multidim/intervalbox.jl
+++ b/src/multidim/intervalbox.jl
@@ -115,45 +115,30 @@ Base.:(==)(x::IntervalBox, y::IntervalBox) = x.v == y.v
 
 
 """
-    mince(x::IntervalBox, n)
+    mince(x::IntervalBox, n::Int)
 
 Splits `x` in `n` intervals in each dimension of the same diameter. These
 intervals are combined in all possible `IntervalBox`-es, which are returned
 as a vector.
 """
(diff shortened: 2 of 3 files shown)


lbenet:
Some benchmarking:

- Current master:
```
julia> @benchmark mince(ii, ndiv) setup=(ii=-1 .. 1, ndiv=100_000_000)
BenchmarkTools.Trial: 
  memory estimate:  152.59 MiB
  allocs estimate:  2
  --------------
  minimum time:     45.498 ms (0.00% GC)
  median time:      64.020 ms (26.33% GC)
  mean time:        58.819 ms (17.82% GC)
  maximum time:     120.155 ms (58.98% GC)
  --------------
  samples:          85
  evals/sample:     1
```

- This PR
```
julia> @benchmark mince(ii, ndiv) setup=(ii=-1 .. 1, ndiv=100_000_000)
BenchmarkTools.Trial: 
  memory estimate:  152.59 MiB
  allocs estimate:  2
  --------------
  minimum time:     41.535 ms (0.00% GC)
  median time:      53.344 ms (16.84% GC)
  mean time:        54.540 ms (19.06% GC)
  maximum time:     117.183 ms (60.92% GC)
  --------------
  samples:          92
  evals/sample:     1
```

src/intervals/arithmetic.jl
```diff
@@ -577,5 +577,5 @@ as a vector.
 """
 function mince(x::Interval, n)
     nodes = range(x.lo, x.hi, length = n+1)
-    return [Interval(nodes[i], nodes[i+1]) for i in 1:length(nodes)-1]
+    return @. Interval(nodes[1:n], nodes[2:n+1])
```
Kolaru, line 580:
```suggestion
    return Interval.(nodes[1:n], nodes[2:n+1])
```
Should be equivalent and avoid the (rather uncommon) macro.

lbenet: Thanks for the suggestion!

src/multidim/intervalbox.jl
```diff
@@ -138,6 +138,26 @@ as a vector.
     end
 end
 
+@generated function mince(x::IntervalBox{N,T}, tup::NTuple{N,Int}) where {N,T}
+    quote
+        n = maximum(tup)
+        nodes_matrix = Array{Interval{T},2}(undef, n, N)
+        fill!(nodes_matrix, emptyinterval(T))
```
Kolaru, line 145:
```suggestion
        nodes_matrix = fill(emptyinterval(T), n, N)
```
should be equivalent.

src/multidim/intervalbox.jl
```diff
@@ -138,6 +138,26 @@ as a vector.
...
+        for i in 1:N
+            nodes_matrix[1:tup[i],i] .= mince(x[i], tup[i])
+        end
+
+        nodes = IntervalBox{$N,T}[]
+        Base.Cartesian.@nloops $N i _->(1:n) begin
```
Kolaru, line 151: Does this part requires carefull review? I am not familiar with the macros used, but I could delve into it if needed.

src/multidim/intervalbox.jl
lbenet: Thanks! Done!

src/multidim/intervalbox.jl
lbenet: Thanks for your review @Kolaru and also for offering getting into the details of this part of the code. 

Here, I simply used the original method, and adapt it to obtain the desired behavior; maybe there are better ways, but for the time being, this is good enough. The original code was inspired from [this section of the manual](https://docs.julialang.org/en/v1/devdocs/cartesian/#Base.Cartesian), which admittedly I have mostly forgotten.

Kolaru: That sounded like a fun multidimensional challenge so I spent some time on it and I believe the following is equivalent:
```julia
function mince(x::IntervalBox{N,T}, ncuts::NTuple{N,Int}) where {N,T}
    minced_intervals = [mince(x[i], ncuts[i]) for i in 1:N]
    minced_boxes = Vector{IntervalBox{N,T}}(undef, prod(ncuts))

    for (k, cut_indices) in enumerate(CartesianIndices(ncuts))
        ibox = IntervalBox([minced_intervals[i][cut_indices[i]] for i in 1:N])
        minced_boxes[k] = ibox
    end
    return minced_boxes
end
```

lbenet: Thanks @Kolaru for the suggestion! I am also changing the other method (avoiding then the `@generated`). I would like to [include you as coauthor](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#required-co-author-information) of this commit; is it ok with you? (I already have the commit ready, but prefer to wait for your reply before pushing it.)

Kolaru: Sure let me know if it needs any input from my side.

lbenet: I just pushed it, so you can have a look

Kolaru:
All look good, thanks.

lbenet:
Thanks a lot!

PR #469, opened 2021-06-02
fix decorated intervals constructors issues

### LINKED ISSUE #170: [NaN, NaN] should not be allowed
I believe the standard specifies that `[NaN, NaN]`, i.e. `NaI`, is not a valid interval *except* if it's decorated? (But I may be wrong.)

### LINKED ISSUE #462: nai() not working with rational intervals
At the moment `nai(Rational)` returns an error

```
julia> nai(Rational)
ERROR: InexactError: Int64(NaN)
Stacktrace:
 [1] Int64
   @ .\float.jl:723 [inlined]
 [2] rationalize(#unused#::Type{Int64}, x::Float64, tol::Int64)
   @ Base .\rational.jl:161
 [3] #rationalize#184
   @ .\rational.jl:216 [inlined]
 [4] Rational
   @ .\rational.jl:119 [inlined]
 [5] Rational
   @ .\rational.jl:123 [inlined]
 [6] convert(#unused#::Type{Rational}, x::Float64)
   @ Base .\number.jl:7
 [7] nai(#unused#::Type{Rational})
   @ IntervalArithmetic ~\.julia\dev\IntervalArithmetic\src\decorations\functions.jl:14
 [8] top-level scope
   @ REPL[25]:1
```

as Luis proposed, one solution might be to define

```
nai(::Rational) = DecoratedInterval(Interval(1//0, -1//0), ill)
```

I think this would work because `NaI` is allowed only as decorated interval and not as bare, hence it would not be confused with `[1//0, -1//0]_trv` which is the empty interval. This might require manually adjusting a couple of functions (at least hardcoding the printing for `NaI` and possibly checking that the current definitions of `isnai` and `isnan` still work), but other than that I don't see why it would not work.

lucaferranti:
## Changes

- fixed issues with decorated intervals constructors (see examples below)
- `interval(nai())` now returns an empty interval instead of `[NaN, NaN]`, it also prints a warning, this is mentioned in the standard
- changed printing of `nai()` from `[NaN, NaN]_ill` to `[NaI]_ill` (happy to switch it back if you prefer the former
- nai(Rational) wasn't supported before (see #462 ), now it returns `[1//0, -1//0]_ill`, which also motivates the change in printing above

## Examples
- before
```
julia> DecoratedInterval(Inf)
[∞, ∞]_dac

julia> DecoratedInterval(1, Inf, com)
[1, ∞]_com

julia> DecoratedInterval(1//2, 3//4)
ERROR: MethodError: no method matching atomic(::Type{Interval{Interval{Rational{Int64}}}}, ::Interval{Interval{Rational{Int64}}})
Closest candidates are:
  atomic(::Type{Interval{T}}, ::Interval) where T<:AbstractFloat at C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\conversion.jl:122
  atomic(::Type{Interval{T}}, ::S) where {T<:AbstractFloat, S<:Real} at C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\conversion.jl:79

julia> nai(Rational)
ERROR: InexactError: Int64(NaN)

julia> interval(nai())
[NaN, NaN]

julia> DecoratedInterval(1, 2, ill)
ERROR: ArgumentError: cannot parse "[NaN, NaN]_ill" as BigFloat

julia> DecoratedInterval(3, 1, ill)
ERROR: InexactError: Int64(NaN)
```

- after
```
julia> DecoratedInterval(Inf)
[NaI]_ill

julia> DecoratedInterval(1, Inf, com)
[1, ∞]_dac

julia> DecoratedInterval(1//2, 3//4)
[1//2, 3//4]_com

julia> nai(Rational)
[NaI]_ill

julia> interval(nai())
┌ Warning: trying to access interval part of NaI, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\decorations\intervals.jl:89
∅

julia> DecoratedInterval(1, 2, ill)
[NaI]_ill

julia> DecoratedInterval(3, 1, ill)
[NaI]_ill
```

## Fixed issues
- fixes #170 
- fixes #462 

## Open problems
- [x] functions that are supposed to return `NaN` still don't work with rational intervals, because I don't think there's a rational equivalent of `NaN` (see #462 ), e.g. `inf(nai(Rational))` fails. **edit** to be addressed in a separated PR
- [ ] While the original ITF tests also test that a warning is thrown when a decorated interval is constructed from an invalid input, I removed the warnings because I think the `NaI` itself is flagging something went wrong. I understand printing the warning for bare intervals, to distinguish between empty interval itself and empty interval because of exceptional behavior, but I think for decorated intervals NaI itself conveys this information. I think also the standard says something like this, but I need to double check, what do you think?

## Question
- why is `decorations` set to `false` by default in the display parameters? I think if someone wants to work with decorated intervals, it is more likely they want to see the decoration?

### DIFF
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -11,7 +11,7 @@ one(::Type{DecoratedInterval{T}}) where T<:Real = DecoratedInterval(one(T))
 
 # NaI: not-an-interval
 """`NaI` not-an-interval: [NaN, NaN]."""
-nai(::Type{<:Int}) = nai(Float64)
+                            nai(::Type{<:Integer}) = nai(Float64)
 nai(::Type{Rational}) = nai(Rational{Int})
 nai(::Type{Rational{T}}) where {T<:Real} = DecoratedInterval{Rational{T}}(Interval{Rational{T}}(1//0, -1//0), ill)
 nai(::Type{T}) where T<:Real = DecoratedInterval{T}(Interval{T}(convert(T, NaN), convert(T, NaN)), ill)
@@ -35,7 +35,10 @@ bool_binary_functions = (
 )
 
 for f in bool_functions
-    @eval $(f)(xx::DecoratedInterval) = $(f)(interval(xx))
+    @eval function $(f)(xx::DecoratedInterval)
+        isnai(xx) && return false
+        return $(f)(interval(xx))
+    end
 end
 
 for f in bool_binary_functions
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -11,12 +11,15 @@ one(::Type{DecoratedInterval{T}}) where T<:Real = DecoratedInterval(one(T))
 
 # NaI: not-an-interval
 """`NaI` not-an-interval: [NaN, NaN]."""
-nai(::Type{T}) where T<:Real = DecoratedInterval(Interval{T}(convert(T, NaN), convert(T, NaN)), ill)
+nai(::Type{<:Int}) = nai(Float64)
+nai(::Type{Rational}) = nai(Rational{Int})
+nai(::Type{Rational{T}}) where {T<:Real} = DecoratedInterval{Rational{T}}(Interval{Rational{T}}(1//0, -1//0), ill)
+nai(::Type{T}) where T<:Real = DecoratedInterval{T}(Interval{T}(convert(T, NaN), convert(T, NaN)), ill)
 nai(x::Interval{T}) where T<:Real = nai(T)
 nai(x::DecoratedInterval{T}) where T<:Real = nai(T)
 nai() = nai(precision(Interval)[1])
 isnai(x::Interval) = isnan(x.lo) || isnan(x.hi)
-isnai(x::DecoratedInterval) = isnai(interval(x)) && x.decoration == ill
+isnai(x::DecoratedInterval) = x.decoration == ill
 isnan(x::Interval) = isnai(x)
 
 ## Bool functions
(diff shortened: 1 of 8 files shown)


src/decorations/functions.jl
```diff
@@ -51,7 +54,12 @@ scalar_functions = (
 )
 
 for f in scalar_functions
-    @eval $(f)(xx::DecoratedInterval{T}) where T = $f(interval(xx))
+    @eval begin 
+        function $(f)(x::DecoratedInterval{T}) where T
+            isnai(x) && return T(NaN)
```
lucaferranti, line 59: this does not work if T is a rational type.

lucaferranti:
why cannot I see the tests running? 🤔

lucaferranti:
according to the standard the interval part of NaI should be the empty interval and a warning should be flagged. I changed `interval(x::DecoratedInterval)` accordingly, but now it's a little bit annoying because it prints the warning also when `interval` is used internally, e.g.

```
julia> nai() + DecoratedInterval(1..2)
┌ Warning: trying to access interval part of NaI, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\decorations\intervals.jl:89
[NaI]

julia> cos(nai())
┌ Warning: trying to access interval part of NaI, empty interval is returned
└ @ IntervalArithmetic C:\Users\lucaa\.julia\dev\IntervalArithmetic\src\decorations\intervals.jl:89
[NaI]
```

possible solutions:
1) use `interval(x::DecoratedInterval) = x.interval` internally and `interval_part` as exported function which checks for `NaI` (or other names)
2) check for `NaI` at the beginning of each operation, but I think this would hurt performance a little?

lucaferranti:
taking care of this for 1.0-dev

PR #477, opened 2021-06-15
Add `floatmin` and `floatmax`

dlfivefifty:


dpsanders:
What's the use case for this?

dpsanders:
Currently there's
```jl
julia> @interval(Inf)
[1.79769e+308, ∞]
```

(Note that in general we now prefer using `interval(a, b)` instead of `@interval(a, b)` since the former is *much* faster. However, infinity is a special case.

dlfivefifty:
I use it in the convergence check in the adaptive QR implemented in InfiniteLinearAlgebra.jl

lucaferranti:
```
julia> @interval(Inf)
[1.79769e+308, ∞]
```

that's not the case anymore on master, it has been changed in #461

PR #480, opened 2021-06-24
overload isapprox for intervals

### LINKED ISSUE #479: Approximate equality with intervals
in the comparison below, i expected `m ≈ interval(0.295, 0.395)` to hold.

```julia
julia> m
[0.294999, 0.395001]

julia> m.lo ≈ 0.295
true

julia> m.hi ≈ 0.395
true

julia> m ≈ interval(0.295, 0.395)
false

julia> @which m ≈ interval(0.295, 0.395)
isapprox(x::Number, y::Number; atol, rtol, nans, norm) in Base at floatfuncs.jl:278
```
see also https://github.com/lucaferranti/IntervalLinearAlgebra.jl/blob/main/src/utils.jl#L1

lucaferranti:
##
`isapprox` wasn't overloaded for intervals and was using the fallback `isapprox(a::Number, b::Number)`. This PR defines the method `isapprox(a::Interval, b::Interval)`

## before
```julia
julia> a = 0.3..0.7
[0.299999, 0.700001]

julia> b = a + 1e-10
[0.3, 0.700001]

julia> a ≈ b
false
```

## after
```julia
julia> a = 0.3..0.7
[0.299999, 0.700001]

julia> b = a + 1e-10
[0.3, 0.700001]

julia> a ≈ b
true
```

## Related issues:
fixes #479

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -20,7 +20,7 @@ export ×, dot
 
 import Base:
     +, -, *, /, //, fma,
-    <, >, ==, !=, ⊆, ^, <=,
+    <, >, ==, !=, ⊆, ^, <=, isapprox,
     in, zero, one, eps, typemin, typemax, abs, abs2, real, min, max,
     sqrt, exp, log, sin, cos, tan, cot, inv, cbrt, csc, hypot, sec,
     exp2, exp10, log2, log10,
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -50,6 +50,7 @@ function strictprecedes(a::Interval, b::Interval)
 end
 const ≺ = strictprecedes # \prec
 
+isapprox(a::Interval, b::Interval; kwargs...) = isapprox(a.lo, b.lo; kwargs...) && isapprox(a.hi, b.hi; kwargs...)
 
 # zero, one, typemin, typemax
 zero(a::Interval{T}) where T<:Real = Interval(zero(T))
--- a/test/interval_tests/numeric.jl
+++ b/test/interval_tests/numeric.jl
@@ -200,6 +200,10 @@ end
 
     # real
     @test real(@interval(-1, 1)) == Interval(-1, 1)
+
+    a = 0.3..0.7
+    b = a + 1e-10
+    @test a ≈ b
 end
 
 @testset "Rational tests" begin


lucaferranti:
Hups, apologies for rushing into the PR without the proper background check, I was aware of the issues/PRs you linked but when I opened this I did not connect the dots.

I can close this PR for the time being, but I do have a couple of comments:

>  it may break code written for numbers in unexpected ways.
This is happening with the current version, which is falling back to `Base.isapprox`, which uses `norm` which for objects of type `Number` uses the absolute value, which for intervals is an interval, which makes no sense in this contest. 

Regardless of what the correct behavior of isapprox should be, the current one is not, 
```julia
julia> a = 1..(1+1e-16)
[1, 1]

julia> b = a + 1e-308
[1, 1.00001]

julia> a ≈ b
false

julia> a.lo ≈ b.hi
true
```
I chose this example because in this case *any* number in a is approximately equal to *any* number in b (with the default tolerance settings), so this should be true in both philosophies. Always returning `missing` for `==` and `isapprox` would cut the head off the snake, but I am a little skeptical it would be practically useful in bigger applications. A proper implementation of b) would require modal logic I think, something like `[1, 2] ≈ [1, 2]` is possibly true and `a ≈ b` from the example above is definitely true.

Moreover, I would say that generic code written for `Number` would break not because some functions are overloaded for `Interval` as here using philosophy a), but because some functions *aren't*, but if you have examples in mind where overloading `==` and `isapprox` breaks, please do prove me wrong :)

lucaferranti:
Also, reading at the old related issues, my main take-home understanding (but I wouldn't be surprised if I misunderstood something) is that: _while the package is mainly following philosphy (a), having `Interval <: Real` is a contradiction, because intervals thought as sets do not fulfill the axioms of real numbers`.

I agree with this! Technically speaking, floating-point numbers do not obey the axioms of real numbers either, since e.g. `NaN` has no additive inverse. However, it is true that not even "ideal interval" with infinite precision etc. obey the axioms of real numbers e.g. distributivity does not hold. 

Hence I agree that having `Interval <: Real` leads to a semantical contradiction and I understand why considering philosophy (b). 

However, what about having `Interval <: Number` I am not an expert in type/set theory of foundations in mathematics in general, but I would say that numbers (whatever a number is) do not have any particularly strict axiom that would be broken by intervals and operations defined as in the standard here?

I am sure you had already thought about this and I apologize if it feels like we are having just another instance of an old conversation. I will re-read the discussions in the related issues and PRs to make sure I don't waste anyone's time by reinventing the wheel. For what is worth, I do think that repeating an old conversation after some time is not necessarily a waste of time, as it can bring up new insights 😄

Kolaru:
First no need to apologize, I pointed out old issues because I thought you may find them interesting, not because I expected you to know about them :) (this goes as far back as #2 so it is quite a rabbit hole)

> Always returning `missing` for `==` and `isapprox` would cut the head off the snake, but I am a little skeptical it would be practically useful in bigger applications.

I think that underlines the core of the problem. You are right, always returning `missing` is not useful. It will simply break any code that expects `==` or `isapprox` to return a boolean. But this is safe. If the code never runs, it will never produce wrong result. The tricky part is that we do not have any other way to guarantee that the code that was given `Interval` is producing guaranteed result when boolean comparison are involved.

> if you have examples in mind where overloading `==` and `isapprox` breaks, please do prove me wrong :)

An example would be a possible Taylor expansion of a function `f(a -b)`
```julia
if a ≈ b
    # Do a Taylor expansion of the function around x
    return f(0) + (a - b) f'(0)
else
    # Go the expensive route
    return f(a - b)
end
```
If you input the interval in you example `a = b = (3..7)` the first branch will be used, but `a - b == (-4..14)` which can not be considered to be close to zero and is certainly not proper for a first order Taylor expansion.

It relies of one of the properties of numbers that intervals are missing (`a == b` => `a - b == 0`).

That kind of branching is not unusual for the computation of special functions like the zeta function (however it generally uses inequalities I think).

> what about having `Interval <: Number`

A lot of packages require `<:Real` to signal they want a number that is not a complex number. Removing `Interval <: Real` would prevent interoperability. Currently the consensus (IIRC) in #271 was to implement strict rules for boolean functions, at least for the default flavor, while keeping `Interval <: Real`.

Another course of action (briefly discussed on slack) would be to encourage removal of this type restriction everywhere and mostly rely on duck typing. But that would be a more global discussion. This is however tangential to the problem of boolean comparison of intervals.

lucaferranti:
thank you for the example, it was an interesting one, also apologies for returning to this so seldomly. I promise that next time I'll answer faster 😄 

The current implementation of `==` at the moment is

```julia
function ==(a::Interval, b::Interval)
    isempty(a) && isempty(b) && return true
    a.lo == b.lo && a.hi == b.hi
end
```

which is in accordance to philosophy a) of your original message. This may change in the future and if in the future we will have the flavour based system with different mechanisms, but since at the moment the "set flavor" is being used for equality, I don't understand why not use the same for approximate inequality, that is here in the PR I am mainly after coherence in the package.

The current fall back implementation is failing because of 
```julia
norm(x-y) <= max(atol, rtol*max(norm(x), norm(y))))
```

and `norm(a::Interval)` is just returning `abs(a)`, which is an interval and not a positive real.

OlivierHnt:
@lucaferranti what is the status of this PR? Perhaps rebase and define some `isapprox_interval` function?

OlivierHnt:
I am closing this PR since it seems abandoned. Feel free to re-open it if you want to revive it at some point, or just open a new one.

PR #481, opened 2021-07-03
Update Project.toml

mforets:


mforets:
The build failure seems unrelated.

lucaferranti:
I also encountered this "mystical" failure on macos a couple of times. I simply pushed an empty commit to retrigger the workflow and then it passed 😕

mforets:
Thanks for checking @dpsanders. Can we have a new release?

The reason (for my use case) is that the master branch has important updates of the `mince` function, so new users adding the package may don't see those changes immediately.

mforets:
Bump? @lucaferranti

lucaferranti:
sorry for the delay, I was browsing the commits after the last release and I think [this](https://github.com/JuliaIntervals/IntervalArithmetic.jl/commit/c7602f4941b550e6a8f11aa2d4cfe774a60f8afc) is breaking, since before it `interval(2, 1)` was erroring and now it returns the empty interval and prints a warning (as the standard requires). So maybe should update the minor release? cc @dpsanders

lucaferranti:
(to be fair, that was my commit so you can blame me for this :) )

dpsanders:
Yes I think this should be a new minor version since it is breaking, as @lucaferranti points out.

PR #482, opened 2021-07-18
Dispatch for exponentiation on Float32 intervals

mfherbst:
Currently this fails:

```julia
Interval(1.0f0) ^ Interval(1.0f0)
```
while this works
```julia
Interval(1.0) ^ Interval(1.0)
```

With this change both things will work, but I'm not too sure if it's the right thing to do. 

Happy to add tests etc.


lbenet:
Thanks @mfherbst and sorry for the long delay to get into this. It would be nice to have some tests. Aside from that, LGTM!

PR #492, opened 2021-09-19
Lf complex interior

lucaferranti:
added isinterior for complex intervals

### before

```julia
julia> a = (1..2) + (3..4)*im
[1, 2] + [3, 4]im

julia> isinterior(a, a)
ERROR: MethodError: no method matching isinterior(::Complex{Interval{Float64}}, ::Complex{Interval{Float64}})
Stacktrace:
 [1] top-level scope
   @ REPL[32]:1
```

### After

```julia
julia> a = (1..2) + (3..4)*im
[1, 2] + [3, 4]im

julia> isinterior(a, a)
false
```


lbenet (approved):
Thanks a lot for addressing this! LGTM.

PR #497, opened 2021-10-12
Use `@ccall` instead of `ccall`

### LINKED ISSUE #494: Showing incorrect number in Julia 1.7-rc1 macOS ARM
Heads up for Julia 1.7, the `show` method for `Interval`s seems to be broken on the macOS ARM version of Julia. With 1.7-beta, it was segfaulting, but now with 1.7-rc1 it's just showing the intervals incorrectly. My guess is it has to do with [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/7e3f494967a37f0c39a2753f3df1539826e58487/src/display.jl#L163).

```julia

julia> a = 1..2
[1.61731e+13636, 1.61731e+13636]

julia> dump(a)
Interval{Float64}
  lo: Float64 1.0
  hi: Float64 2.0
```

```julia
(jl_Zvx3xi) pkg> st
      Status `/private/var/folders/jk/zbhsb6bd25b9swwhb_3d6v340000gn/T/jl_Zvx3xi/Project.toml`
  [d1acc4aa] IntervalArithmetic v0.19.2

julia> versioninfo()
Julia Version 1.7.0-rc1
Commit 9eade6195e (2021-09-12 06:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin20.5.0)
  CPU: Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, cyclone)
Environment:
  JULIA_NUM_THREADS = 8
```

giordano:
How'd you feel about requiring Julia v1.5?  This change makes the package work on `aarch64-apple-darwin` (aka Apple Silicon, for example the M1 processor).  For the background issue see https://github.com/JuliaLang/julia/issues/42614.

Fix #494.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.19.2"
+version = "0.20.0"
 
 [deps]
 CRlibm = "96374032-68de-5a5b-8d9e-752f78720389"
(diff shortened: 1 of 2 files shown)


lucaferranti:
Would this also solve #494?

giordano:
Yes

lucaferranti:
@lbenet should we go ahead and merge this? I don't think there's much we can do for testing atm

Issue #473, opened 2021-06-03
Normalise negative zero to zero


dpsanders:
```jl
julia> interval(0.0, -0.0)
[0, -0]
```

We should definitely normalise -0.0 to 0.0.

lucaferranti:
Section 12.12.8 Numeric functions of intervals

> In formats that have a signed zero, a Level 1 value of 0 shall be returned as −0 by inf, and +0 by all other
functions in this subclause

so if I understand correctly we should map 0 to +0.0 except for `inf` which should return `-0.0`

lucaferranti:
but the only cases I can think of in practice when this would matter are

1. printing, like in the example you posted
2. when using `===`

lbenet:
A (naive) possible implementation (for printing):
```julia
julia> myprint(x) = !iszero(x) ? x : zero(x)
myprint (generic function with 1 method)

julia> myprint(0.0)
0.0

julia> myprint(-0.0)
0.0
```

lucaferranti:
The ITL test suite has indeed 
```
    inf [0.0,infinity] = -0.0;
    inf [-0.0,infinity] = -0.0;
    inf [-0.0,0.0] = -0.0;
    inf [0.0,-0.0] = -0.0;
    inf [0.0,0.0] = -0.0;
    inf [-0.0,-0.0] = -0.0;
```
 
so maybe we should define

```
    inf(x::Interval) = iszero(x.lo) ? -zero(x.lo) : x.lo
    sup(x::Interval) = iszero(x.hi) ? zero(x.hi) : x.hi
```
    
I don't recall the details of printing, but if `inf` and `sup` are used to access the boundaries of the interval, than this would fix that too. 

This also strengthens my idea that we should use `===` in test, mainly for two reasons: 1) detect this corner case, 2) avoid checking decoration of intervals separately for decorated intervals.

lucaferranti:
WoW this opened Pandora's box! I generated the tests in ITF1788.jl using === and now there are many more tests failing because of that, e.g.

```
julia> /(interval(-Inf,-15.0), interval(3.0,Inf))
[-∞, -0]
```
expected value `Interval(-Inf, 0.0)`

and
```
julia> cancelplus(interval(1.0, 5.0), interval(-5.0,-1.0))
[-0, 0]
```
expected value `Interval(0.0, 0.0)`

Kolaru:
[...]

PR #500, opened 2021-12-13
Fix display of -0.0 to 0.0

lbenet:
See [this comment](https://github.com/JuliaIntervals/ITF1788.jl/issues/4#issuecomment-992043408) for the motivation.

### DIFF
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -315,7 +315,7 @@ function atan(yy::DecoratedInterval{T}, xx::DecoratedInterval{T}) where T
     if zero(T) ∈ y
         zero(T) ∈ x && return DecoratedInterval(r, trv)
         if x.hi < zero(T)
-            signbit(y.lo) && return DecoratedInterval(r, min(d, def))
+            y.lo < zero(T) && return DecoratedInterval(r, min(d, def))
             return DecoratedInterval(r, min(d, dac))
         end
     end
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -314,8 +314,9 @@ function atan(yy::DecoratedInterval{T}, xx::DecoratedInterval{T}) where T
     # Check cases when decoration is trv and decays (from com or dac)
     if zero(T) ∈ y
         zero(T) ∈ x && return DecoratedInterval(r, trv)
-        if x.hi < zero(T) && y.lo != y.hi && signbit(y.lo) && Int(d) > 2
-            return DecoratedInterval(r, decay(d))
+        if x.hi < zero(T)
+            signbit(y.lo) && return DecoratedInterval(r, min(d, def))
+            return DecoratedInterval(r, min(d, dac))
         end
     end
     DecoratedInterval(r, d)
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -302,7 +302,7 @@ end
 
 
 # Infimum and supremum of an interval
-inf(a::Interval) = a.lo
+inf(a::Interval) = ifelse(iszero(a.lo) && !signbit(a.lo), copysign(a.lo, -1), a.lo)
 sup(a::Interval) = a.hi
 
 
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -40,7 +40,6 @@ struct Interval{T<:Real} <: AbstractInterval{T}
 end
 
 @inline _normalisezero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
-@inline _normalisezero(a::Integer) = _normalisezero(float(a))
 
 
 ## Outer constructors
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -19,9 +19,11 @@ struct Interval{T<:Real} <: AbstractInterval{T}
 
     function Interval{T}(a::Real, b::Real) where T<:Real
 
+        a = _normalisezero(a)
+        b = _normalisezero(b)
+
         if validity_check
-            a = _normalisezero(a)
-            b = _normalisezero(b)
+
             if is_valid_interval(a, b)
                 return new(a, b)
 
@@ -38,7 +40,7 @@ struct Interval{T<:Real} <: AbstractInterval{T}
 end
 
 @inline _normalisezero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
-@inline _normalisezero(a::Integer) = float(a)
+@inline _normalisezero(a::Integer) = _normalisezero(float(a))
 
 
 ## Outer constructors
--- a/src/intervals/intervals.jl
+++ b/src/intervals/intervals.jl
@@ -20,9 +20,10 @@ struct Interval{T<:Real} <: AbstractInterval{T}
     function Interval{T}(a::Real, b::Real) where T<:Real
 
         if validity_check
-
+            a = _normalisezero(a)
+            b = _normalisezero(b)
             if is_valid_interval(a, b)
-                new(a, b)
+                return new(a, b)
 
             else
                 @warn "Invalid input, empty interval is returned"
@@ -36,6 +37,8 @@ struct Interval{T<:Real} <: AbstractInterval{T}
     end
 end
 
+@inline _normalisezero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
+@inline _normalisezero(a::Integer) = float(a)
 
 
 ## Outer constructors
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.20.1"
+version = "0.20.2"
 
 [deps]
 CRlibm = "96374032-68de-5a5b-8d9e-752f78720389"
--- a/src/display.jl
+++ b/src/display.jl
@@ -185,6 +185,8 @@ function basic_representation(a::Interval, format=nothing)
 
     local output
 
+    a = Interval(_signofzero(a.lo), a.hi)
+
     if format == :standard
 
         aa = round_string(a.lo, sigfigs, RoundDown)
@@ -218,6 +220,8 @@ function basic_representation(a::Interval{Float32}, format=nothing)
     sigfigs = display_params.sigfigs
 
     local output
+
+    a = Interval(_signofzero(a.lo), a.hi)
 
     if format == :standard
 
@@ -267,6 +271,8 @@ function basic_representation(a::Interval{Rational{T}}, format=nothing) where
 
     output
 end
+
+@inline _signofzero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
 
 function subscriptify(n::Integer)
     dig = reverse(digits(n))
--- a/src/display.jl
+++ b/src/display.jl
@@ -185,8 +185,6 @@ function basic_representation(a::Interval, format=nothing)
 
     local output
 
-    a = Interval(_signofzero(a.lo), a.hi)
-
     if format == :standard
 
         aa = round_string(a.lo, sigfigs, RoundDown)
@@ -220,8 +218,6 @@ function basic_representation(a::Interval{Float32}, format=nothing)
     sigfigs = display_params.sigfigs
 
     local output
-
-    a = Interval(_signofzero(a.lo), a.hi)
 
     if format == :standard
 
@@ -271,8 +267,6 @@ function basic_representation(a::Interval{Rational{T}}, format=nothing) where
 
     output
 end
-
-@inline _signofzero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
 
 function subscriptify(n::Integer)
     dig = reverse(digits(n))


lbenet:
Example:
```julia
# Currently in master
julia> f(x) = 1 - x^4 + x^5

julia> ff = f(0..1)
[-0, 2]

# With this PR
julia> f(x) = 1 - x^4 + x^5

julia> ff = f(0..1)
[0, 2]
```

lucaferranti (approved):

lucaferranti:
From the printing side great!
Do you think this can close #473 ?

lbenet:
Almost, but not quite. The following still doesn't look nice:
```julia
julia> interval(0.0, -0.0)
[0, -0]
```
Maybe having that zero returns always with positive sign is the best.

Oder opinions?

lucaferranti:
I think calling the `_signofzero` to the upper bound would fix it?

Regarding the issue I linked: is the -0 vs 0 just a printing thing or do we actually have to make sure that 0 has the bitstring of -0 when it's in the lower bound and the bitstring of +0 when is in the upper bound?

schillic:
> Regarding the issue I linked: is the -0 vs 0 just a printing thing or do we actually have to make sure that 0 has the bitstring of -0 when it's in the lower bound and the bitstring of +0 when is in the upper bound? This is anyway tangential to this PR

I think ±0 are considered identical (two representations of the same number).

```julia
julia> x = -0.0
-0.0

julia> y = 0.0
0.0

julia> x < y
false

julia> x > y
false

julia> x == y
true
``

**Comment by lucaferranti** [comment]
they have different bit representation
```julia
julia> +0.0 === -0.0
false

julia> bitstring(+0.0)
"0000000000000000000000000000000000000000000000000000000000000000"

julia> bitstring(-0.0)
"1000000000000000000000000000000000000000000000000000000000000000"
```
There are a couple of points in this standard where this small nuisance matters

sec 12.12.8
> In formats that have a signed zero, a Level 1 value of 0 shall be returned as −0 by inf, and +0 by all other
functions in this subclause.

hence e.g. 
```julia
julia> inf(0..0) |> bitstring
"0000000000000000000000000000000000000000000000000000000000000000"
```

I would need to double check the ITF1788 testsuite to see if that's the only place where you actually check that you get a -0, but I think it is.

SUMMA SUMMARUM:

1. This PR normalises the printing, which is good, and I think that mutatis mutandis it can be merged
2. My comment in that issue (and on that other issue in ITF1788.jl) is that by the standard in some cases (I think only with `inf`) we need to make sure to actually return the right 0, but this can be addressed also in a separated PR I think.

lbenet:
The signed zero is a floating-point arithmetic feature, if i recall correctly, to have a unique answer for the `inv` function.

@lucaferranti Am I understanding you correct that the interval standard imposes that `Interval(0,2)` should indeed be displayed as `[-0,2]`? What about `Interval(0.0,0.0)`?

(I'm having a busy morning. I'll try to get back to this in the afternoon)

lbenet:
I think this commit now fixes #473. Using it we have:
```julia
julia> -0 .. 0
[0, 0]

julia> dump( -0 .. 0)
Interval{Float64}
  lo: Float64 0.0
  hi: Float64 0.0

julia> dump( 0 .. -0)
Interval{Float64}
  lo: Float64 0.0
  hi: Float64 0.0

julia> inf(0 .. 0)
-0.0

julia> inf(-0 .. 0)
-0.0

julia> sup( -0 .. -0)
0.0
```

src/intervals/arithmetic.jl
```diff
@@ -302,7 +302,7 @@ end
 
 
 # Infimum and supremum of an interval
-inf(a::Interval) = a.lo
+inf(a::Interval) = ifelse(iszero(a.lo) && !signbit(a.lo), copysign(a.lo, -1), a.lo)
 sup(a::Interval) = a.hi
```
lucaferranti, line 306: should a similar check also be here for `sup` I am mainly thinking the case `sup(interval(0.0, -0.0))`. Maybe something like
```julia
sup(a::Interval) = ifelse(iszero(a.hi), abs(a.hi), a.hi)
```

or using the copysign and signbit if it's more performant.

src/intervals/intervals.jl
```diff
@@ -36,6 +37,8 @@ struct Interval{T<:Real} <: AbstractInterval{T}
     end
 end
 
+@inline _normalisezero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
+@inline _normalisezero(a::Integer) = float(a)
```
lucaferranti, line 41: is this needed? I think integer is already converted to float by the constructor at some point?

lucaferranti:
@lbenet I left a couple of comments but overall LGTM

src/intervals/intervals.jl
```diff
@@ -20,9 +20,10 @@ struct Interval{T<:Real} <: AbstractInterval{T}
     function Interval{T}(a::Real, b::Real) where T<:Real
 
         if validity_check
-
+            a = _normalisezero(a)
```
lucaferranti, line 23: I think `Interval` is by default with `validity_check` turned off? I wonder how often this would be effectively used

src/intervals/arithmetic.jl
lbenet: I don't think anything is needed with regards to `sup`, unless there is some specification in the standard that we haven't considered yet. 

As shown [above](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/500#issuecomment-993103088), an interval `a = -0 .. -0` is such that `a.lo=0.0` and `a.hi=0.0`, so the signbit is already taken care of. That's the reason `sup(a)` is simply `a.hi`.

src/intervals/intervals.jl
lbenet: The first one yes; the second one I am not yet sure... There is a corner case I'm trying to figure out...

src/intervals/intervals.jl
lbenet: Good catch! These lines should have been before the `if`... I'll push a new commit with that change.

lucaferranti: for example
```julia
julia> a = interval(0.0, -0.0)
[0, -0]

julia> sup(a)
-0.0
```

I wonder whether it would make more sense to have the check for +0 and -0 only in the `inf` `sup` functions and the normalisation only in the printing, what do you think?

src/intervals/intervals.jl
lbenet: I'm checking it right now. The subtlety with that line is that `Integer`s have no signbit, which was the reason I original included it.

src/intervals/intervals.jl
lbenet: I think your observation does solve the corner case, which probably was due to `interval`. The one I was checking was `1 - (0..1)`. Before I was getting `[-0, 1]`; moving outside the `if` those lines I'm getting `[0, 1]`, but there is a (decorated) test not passing...

src/intervals/arithmetic.jl
lucaferranti: yes I think if you move the normalisation before the `if validity_check` then it would work. I was mainly thinking to the example [below](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/500#discussion_r768718450)

src/intervals/intervals.jl
lbenet: Just pushed a new commit. It does take care of the corner case, or of your example, but there is a problem with one decorated test... Any idea?

lucaferranti: I think [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/decorations/functions.jl#L317) is the reason

lucaferranti: mmmm
```julia
julia> yy = DecoratedInterval(Interval(-0.0, 1.0), com)
[-0, 1]_com

julia> yyy = DecoratedInterval(Interval(0.0, 1.0), com)
[0, 1]_com

julia> xx = DecoratedInterval(Interval(-2.0, -0.1), com)
[-2, -0.1]_com

julia> atan(yyy, xx)
[1.67046, 3.1416]_com

julia> atan(yy, xx)
[1.67046, 3.1416]_dac
```

lucaferranti: I think that is a bug in atan? I don't think the behavior should differ whether it's +0 or -0

lbenet: Not sure yet... see [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/af22f4a1c24070de3ae25050526f222f57563cbe/src/decorations/functions.jl#L317). The sign seems to matter, and since we just suppressed it...

lucaferranti: in octave:
```octave
䰛ﮨ>> a = infsupdec(-0.0, 1.0)
a = [0, 1]_com
>> b = infsupdec(-2.0, -0.1)
b = [-2, -0.1]_com
>>
>> atan2(a, b)
ans = [1.6704, 3.1416]_dac
>> aa = infsupdec(0.0, 1.0)
aa = [0, 1]_com
>> atan2(aa, b)
ans = [1.6704, 3.1416]_dac
```

lucaferranti: I think the sign shouldn't matter for 0, that line is buggy I think

lucaferranti: @lbenet I pushed a commit that I think fixed it. Now it passes all tests (at least on my machine, let's see what the CI says) and behaves like the octave package. Before there were some discrepancies. I have to confess I am not fully sure why it works this way either. If you spot a mistake or would prefer to have this fix in a separate PR, I'd be glad to revert.

IntervalArithmetic.jl v0.20.1
```julia
(@v1.6) pkg> st IntervalArithmetic
      Status `~/.julia/environments/v1.6/Project.toml`
  [d1acc4aa] IntervalArithmetic v0.20.1

julia> atan(DecoratedInterval(-2..2), DecoratedInterval(-2.. -1))
[-3.1416, 3.1416]_dac

julia> atan(DecoratedInterval(-2..0), DecoratedInterval(-2.. -1))
[-3.1416, 3.1416]_dac

julia> atan(DecoratedInterval(0..0), DecoratedInterval(-2.. -1))
[3.14159, 3.1416]_com

julia> atan(DecoratedInterval(0..2), DecoratedInterval(-2.. -1))
[2.03444, 3.1416]_com
```

octave
```octave
>> atan2(infsupdec(-2, 2), infsupdec(-2, -1))
ans = [-3.1416, +3.1416]_def
>> atan2(infsupdec(-2, 0), infsupdec(-2, -1))
ans = [-3.1416, +3.1416]_def
>> atan2(infsupdec(0, 0), infsupdec(-2, -1))
ans = [3.1415, 3.1416]_dac
>> atan2(infsupdec(0, 2), infsupdec(-2, -1))
ans = [2.0344, 3.1416]_dac
```

after my latest commit
```julia
julia> atan(DecoratedInterval(-2..2), DecoratedInterval(-2.. -1))
[-3.1416, 3.1416]_def

julia> atan(DecoratedInterval(-2..0), DecoratedInterval(-2.. -1))
[-3.1416, 3.1416]_def

julia> atan(DecoratedInterval(0..0), DecoratedInterval(-2.. -1))
[3.14159, 3.1416]_dac

julia> atan(DecoratedInterval(0..2), DecoratedInterval(-2.. -1))
[2.03444, 3.1416]_dac

```

src/decorations/functions.jl
```diff
@@ -314,8 +314,9 @@ function atan(yy::DecoratedInterval{T}, xx::DecoratedInterval{T}) where T
     # Check cases when decoration is trv and decays (from com or dac)
     if zero(T) ∈ y
         zero(T) ∈ x && return DecoratedInterval(r, trv)
-        if x.hi < zero(T) && y.lo != y.hi && signbit(y.lo) && Int(d) > 2
-            return DecoratedInterval(r, decay(d))
+        if x.hi < zero(T)
+            signbit(y.lo) && return DecoratedInterval(r, min(d, def))
```
lbenet, line 318: I think `signbit(y.lo)` will always be `false` due to the changes i have implemented. (The constructor makes that -0.0 becomes 0.0.) I'm not sure if `signbit(inf(y))` is what is needed...

lucaferranti: what I wanted to achieve in that line is 
```julia
y.lo < zero(T) && return DecoratedInterval(r, min(d, def))
``` 
because of the change you made that should be equivalent to `signbit(y.lo)`. Maybe it's still better to use `y.lo < zero(T)` though, clearer to read and safer to these -0 vs 0 issues.

lbenet: But it seems to work... 😄

lucaferranti: (side note: I think the `decay` function is not needed anymore?)

src/intervals/intervals.jl
lucaferranti: the constructor for integers is
```julia
Interval(a::T, b::T) where T<:Integer = Interval(float(a), float(b))
```
so I think the integer is converted to float before getting to that line.

src/decorations/functions.jl
lbenet: The changes (in this PR) make that an `Interval` with upper/lower bound equal to zero, has a positive zero; so `signbit` is false *always*. Yet, `inf` returns -0.0 if the lower bound is zero, as the standard demands, in which case `signbit` return true. My point is that the `if` in line 318 is never used.

src/intervals/intervals.jl
lbenet: You are right... I'll get rid of it.

src/decorations/functions.jl
lucaferranti: the `if` on line 318 is used when `y.lo` is strictly lower than zero, e.g.
```julia
julia> y = DecoratedInterval(-2..0)
[-2, 0]_com

julia> x = DecoratedInterval(-2.. -1)
[-2, -1]_com

julia> atan(y, x)
[-3.1416, 3.1416]_def

```

By looking at ITL tests and playing with octave you indeed _never_ want to use that line if `y.lo == 0`

lbenet:
I just pushed a new commit, getting rid of the `_normalisezero` method that was not used. Indeed, `y.lo < 0` seems necessary, as you commented. I think this is ready to be merged!

Thanks a lot @lucaferranti for finding the tiny little uncomfortable issue in the `if`!!

lucaferranti (approved):
I think this is ready to be merged too! (maybe bump the patch version?) 🎉

lbenet:
Done!

lbenet:
Merging...

PR #503, opened 2022-01-19
Bump CRlibm to 1 while keeping the existing compats

ven-k:
- The older versions of CRlibm untars the deps which can't be supported on machines where you don't have root access
- CRlibm fails to build with
```
tar: crlibm-1.0beta4/trigo_fast.c: Cannot change ownership to uid 1000, gid 1000: Invalid argument
tar: crlibm-1.0beta4/trigpi.h: Cannot change ownership to uid 1000, gid 1000: Invalid argument
tar: crlibm-1.0beta4/ChangeLog: Cannot change ownership to uid 1000, gid 1000: Invalid argument
tar: crlibm-1.0beta4/triple-double.h: Cannot change ownership to uid 1000, gid 1000: Invalid argument
tar: crlibm-1.0beta4/log2-td.c: Cannot change ownership to uid 1000, gid 1000: Invalid argument
tar: crlibm-1.0beta4/exp-td-standalone.c: Cannot change ownership to uid 1000, gid 1000: Invalid argument
:
:
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.20.2"
+version = "0.20.3"
 
 [deps]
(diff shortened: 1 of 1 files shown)


lucaferranti:
Thanks, I was under the impression that there were some issues with CRlibm 1, but everything seems to work, merging

Issue #502, opened 2021-12-29
Macros can not guarantee tight inclusion of float literals


Kolaru:
Float literals are parsed before being passed to the macro:

```julia
julia> macro isitonetenth(expr)
           @show expr
           @show typeof(expr)
           @show expr == 0.1
       end
@isitonetenth (macro with 1 method)

julia> @isitonetenth 0.100000000000000006
expr = 0.1  # I was expecting Symbol("0.100000000000000006") here
typeof(expr) = Float64  # It has already been parsed to a floating point number
expr == 0.1 = true
true
```

This means that `@interval` does not know what the literal was before rounding, which means it can lead to incorrect enclosures:

```julia
julia> x = @interval 0.100000000000000006
[0.0999999, 0.100001]

julia> big(x.hi)
0.1000000000000000055511151231257827021181583404541015625

julia> BigFloat("0.100000000000000006") in x
false
```
i.e. despite typing `0.100000000000000006` explicitly, the user has no guarantee it actually ends up in the interval.

My conclusion is that with floats literals we can not do better than returning `(prevfloat(x)..nextfloat(x))` with either the `@interval` macro or `atomic`. I plan to include it in the final version of #271, since in passing I ended up rewriting big parts of `atomic`.

dpsanders:
Good point. It is pretty frustrating that the Julia parser works like that!

OlivierHnt:
I am closing this issue since `@interval` has been removed (and if it were to be reintroduced later, it will just be a syntax sugar to wrap every constant literals in an expression with `interval`). Also `atomic` does exactly `(prevfloat(x)..nextfloat(x))`.

Issue #465, opened 2021-05-27
ITF1788 compliance


lucaferranti:
I have recently created the package [ITF1788.jl](https://github.com/JuliaIntervals/ITF1788.jl) which translates ITL to julia tests, runs the tests and marks the failing ones as broken. This should make it easier to see what is missing. Here is a summary (except for reverse functions, I will open an issue for those in IntervalContractors.jl soonish).

- **fi_lib.itl**
  - [x] **1**. some problems with acot (Luis pointed out that this function is recommended and is also missing from MPFR, indeed the name of the file suggests these were tests for FiLib library, not for the standard).
  - [x] **2**. some problems with acoth (same of above)
  
- **ieee1788-exceptions.itl**
  - [x] **3**. `b-textToInterval "[+infinity]" = [empty] signal UndefinedOperation;` fixed in #522
  - [ ] **4**. Interval part of `NaI` should return the empty interval, now it returns `[NaN, NaN]`
  - [x] **5**. `b-textToInterval "[1.0000000000000001, 1.0000000000000002]" = [1.0, 0x1.0000000000001p+0] signal PossiblyUndefinedOperation;` fixed in #522
  
 - **ieee1788-class.itl**
    - [x] **6**. invalid inputs should return an empty interval for bare intervals #457 and #461 e.g. 
     ```
    b-numsToInterval NaN NaN = [empty] signal UndefinedOperation;
    b-numsToInterval 1.0 -1.0 = [empty] signal UndefinedOperation;
    b-numsToInterval -infinity -infinity = [empty] signal UndefinedOperation;
    b-numsToInterval infinity infinity = [empty] signal UndefinedOperation;
    ```
    - [x] **7**. invalid strings should return an empty interval fixed in #522
    - [x] **8**. some problems in parsing intervals from strings, probably due to current implementation being case sensitive fixed in #522
    - [x] **9**. in some cases same problem due to #464
    - [ ] **10**. a few tests failing when constructing decorated intervals.
   
- **ieee1788-elem**
   - [x] **11**. ~roundTiesToAway and roundTiesToEven missing~, they are implemented, I just didn't notice those before.
   
- **ieee1788-num**
[...]

PR #504, opened 2022-01-24
Fix ITF1788 tests (assuming #271 is merged)

lbenet:


lbenet:
This PR builds from @Kolaru's #271, giving a space to cherry-pick commits that add/fix functionality, specifically aiming to fix ITF1788 tests. I plan to push some new commits, related to trigonometric and hyperbolic functions, but I haven't quite finished them.

cc @dpsanders @lucaferranti

lucaferranti:
@lbenet would you find it helpful if I generate all tests with ITF1788.jl and push them here? The package can try the tests and mark the failing ones as broken, this would give a complete list of what is broken and what not.

lbenet:
Yes, and thanks for doing so! Just delete/comment the lines needed (before) related to the precision.

At the moment I'm not taking care about the parsing part, so some tests there maybe broken, see also #502.

lbenet:
Incidentally @lucaferranti , for the `:ieee1788` mode, `≛` and `==` in principle behave equivalently, and similarly `iszero` and `isthinzero` (and perhaps other functions). I've kept @Kolaru's proposal of distinguish them for other modes. I think it would be nice to test with both operators, or have the tests using `≛`.

lucaferranti:
@lbenet added the tests. The broken tests are marked with `@test_skip`, the way I use them is that I ctrl+F `@test_skip` to see what tests are failing. The tests I added are missing the files for reverse functions (because they are in a separate repository) and the tests for the ovelap function (because it's not implemented), so they are not relevant here.

I had previously opened #465 summarizing what was failing in master, hopefully comparing this to that will help to track what is a regression and what is a known bug. 

In the tests I use `===`, the problem with using `==` and or the or `≛` is that according to the standard two decorated intervals are equal regardless of the decoration, but in tests you want to check that the decoration is also correct.

After you introduced the normalization of zero in your previous PR, using `===` doesn't introduce false positives anymore (or maybe it does, actually I haven't checked carefully :) )

A couple of regressions I noticed
- **libieeep1788_num.jl` now some tests with `inf` are failing, this is because now the definition is `inf(x) = x.lo`, whereas the standard says that if 0 is a lower bound it should return -0, this should be easy to fix. e.g. `inf(x) = iszero(x.lo) ? -x.lo : x.lo` or similar.
-  Now `atanh` errors:
  - master
  ```julia
julia> atanh(Interval(0, 1))
[0, ∞]
```

  - this PR
  ```julia
julia> atanh(Interval(0, 1))
ERROR: MethodError: no method matching BigFloat(::Interval{BigFloat}, ::Base.MPFR.MPFRRoundingMode; precision=256)
Closest candidates are:
  BigFloat(::Union{Float16, Float32}, ::Base.MPFR.MPFRRoundingMode; precision) at C:\Users\lucaa\AppData\Local\Programs\Julia-1.7.1\share\julia\base\mpfr.jl:236
  BigFloat(::Rational, ::Base.MPFR.MPFRRoundingMode; precision) at C:\Users\lucaa\AppData\Local\Programs\Julia-1.7.1\share\julia\base\mpfr.jl:239
  BigFloat(::AbstractString, ::Base.MPFR.MPFRRoundingMode; precision) at C:\Users\lucaa\AppData\Local\Programs\Julia-1.7.1\share\julia\base\mpfr.jl:254
  ...
Stacktrace:
 [1] BigFloat(x::Interval{BigFloat}, r::RoundingMode{:Down}; precision::Int32)
   @ Base.MPFR .\mpfr.jl:262
 [2] BigFloat(x::Interval{BigFloat}, r::RoundingMode{:Down})
   @ Base.MPFR .\mpfr.jl:262
 [3] Interval{BigFloat}(a::Interval{BigFloat}, b::Interval{BigFloat})
   @ IntervalArithmetic c:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\construction.jl:70
 [4] atanh(a::Interval{BigFloat})
   @ IntervalArithmetic c:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\arithmetic\hyperbolic.jl:62
 [5] atanh(a::Interval{Float64})
   @ IntervalArithmetic c:\Users\lucaa\.julia\dev\IntervalArithmetic\src\intervals\arithmetic\hyperbolic.jl:69
 [6] top-level scope
   @ REPL[2]:1
```

but I haven't checked yet what has changed, I'll try to have a closer look tomorrow, unless you solve it first :)

Kolaru:
> * **libieeep1788_num.jl** now some tests with `inf` are failing, this is because now the definition is `inf(x) = x.lo`, whereas the standard says that if 0 is a lower bound it should return -0, this should be easy to fix. e.g. `inf(x) = iszero(x.lo) ? -x.lo : x.lo` or similar.

Wouldn't it be easier to have `x.lo` be `-0` directly in those cases?

lbenet:
Thanks a lot @lucaferranti for pushing the updated version.

Few remarks:
- `test/interval_tests/numeric.jl` and `test/interval_tests/linear_algebra.jl` have broken tests, so I'll mark the problematic tests with `@test_skip` for the time being. This may be related to the fact that I'm using the `:ieee1788` as the default pointwise politic.
- The structure and file names of the ITF1788 changed, so I'm readapting. I noticed that `libieeep1788_reduction.jl` has no tests related with intervals(!), and still some tests are broken/skip. Is this expected? Moreover, for decorated intervals, the current suite uses `interval_part` to extract the interval, while the instruction is `interval`. Could you fix this @lucaferranti ?
- Regarding the exceptions, for the decorated intervals, do we need to have a `@warn` according to the standard? Currently we get none.
- `test/test_ITF1788/ieee1788-constructors.jl` still has some problems, related to parsing strings.
- By some reason, some tests of the trigonometric and hyperbolic functions are broken; this is what I'm fixing now.
- I think it's good to use `===`, since this signals differences such as in the sign bit of zero. We ma have to revise that.

lucaferranti:
> Moreover, for decorated intervals, the current suite uses interval_part to extract the interval, while the instruction is interval.
I'm on it 💪 

> The structure and file names of the ITF1788 changed, so I'm readapting. I noticed that libieeep1788_reduction.jl has no tests related with intervals(!), and still some tests are broken/skip
Those tests were included in the original ITL testsuite. They regard section 12.12.12 of the standard. The only test failing is
```julia
@test_skip dot([0x10000000000001p0, 0x1p104], [0x0fffffffffffffp0, -1.0]) === -1.0
```
The standard suggests that a software implementation has routines for accurate dot product and summation. Basically that test is numerically challenging, using naive floating point you get 0, but the true result is -1. The correct result could be achieved using e.g. the Oishi-Rump-Ogita algorithm, which is implemented in `AccurateArithmetic.jl`
```julia
julia> using AccurateArithmetic

julia> dot([0x10000000000001p0, 0x1p104], [0x0fffffffffffffp0, -1.0])
0.0

julia> dot_oro([0x10000000000001p0, 0x1p104], [0x0fffffffffffffp0, -1.0]) #accurate summation
-1.0
```
a simple solution would be to re-export `AccurateArithmetic` and document that accurate summation and dot-product can be achieved with `sum_oro` and `dot_oro`. However, I think this is secondary at this stage

> test/test_ITF1788/ieee1788-constructors.jl still has some problems, related to parsing strings.
Yes, that is a known bug already tracked in #465, not introduced in this PR

> By some reason, some tests of the trigonometric and hyperbolic functions are broken; this is what I'm fixing now.
some issues with `acot` and `acoth` were already present before, see #465 . If I recall correcly, the problem with that is that IntervalArithmetic.jl returns a valid but not tight enclosure, that is the returned interval is a few ULPs wider than the one required by the ITL tests. If you can fix that be my guest, but a word of warning it might be hard without relying on MPFR :).

If you find a solution for these "already existing" bugs, I think it's better to open directly a PR to master and make a new patch-release.

lucaferranti:
> @lucaferranti I've noticed that some of the broken (skip) tests are due to trivial promotion issues. For example, diam(interval(-8.0, 0.0)) === +8 is broken because you are comparing a Float64 (8.0) against an Int (+8). (This is the same as getting false in 0.0 === 0.) Is it possible to regenerate such tests (automatically) with the proper promoted comparisons? If not, I can do it manually.

Thanks, I didn't notice it!! yes I'll fix it!

For the tests I think the best thing to do is the following:

we push *all* the generated ITF1788 tests (except for reverse functions which are in IntervalContractors) both here and to main (via PR). This way we can keep track of what tests were already failing and what new failures are introduced by this PR and the comparison should be easier, what do you think @lbenet ?

lbenet:
Regarding the tests, we should have the reverse tests available, but perhaps have them commented in `run_ITF1788.jl`. This way, we may eventually test them as well.

Regarding the rev functions, in my opinion and for compliance with the standard, we should have them here. `IntervalContractors.jl` will still need the infrastructure developed here, so it adds no further dependency there. Maybe @dpsanders has some opinion on this?

lucaferranti:
@lbenet I've now uploaded the latest generated ITF1788 tests, updated with your comments above and with your latest commit, there was indeed a lot of improvement.

PR #509, opened 2022-03-05
Fix cancelminus

### LINKED ISSUE #508: `cancelminus` specification
The following code
```julia
a = @interval(1e308, 1e308)
b = @interval(1e-308, 1e-308)
@show cancelminus(a, b)
```
prints
```
cancelminus(a, b) = [9.99999e+307, 1.00001e+308]
```
Is it really what the standard mandates?  Shouldn't it be `Interval(prevfloat(Inf), Inf)`?

lbenet:
Fixes #508

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.20.3"
+version = "0.20.4"
 
 [deps]
 CRlibm = "96374032-68de-5a5b-8d9e-752f78720389"
--- a/src/intervals/arithmetic.jl
+++ b/src/intervals/arithmetic.jl
@@ -531,21 +531,21 @@ function cancelminus(a::Interval{T}, b::Interval{T}) where T<:Real
     c_lo = @round_down(a.lo - b.lo)
     c_hi = @round_up(a.hi - b.hi)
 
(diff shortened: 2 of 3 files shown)


lucaferranti (approved):
LGTM, can you bump the project patch version?
Also, I think to understand that the corner case 2 wasn't covered in the ITF1788 test-suite, maybe we should test that somewhere?

lbenet:
Thanks for your review @lucaferranti !

I think it is covered, at least in [this file](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/test/test_ITF1788/libieeep1788_cancel.jl), but I should look at it more careful. Now I can't look at this, but in the afternoon I'll check that and if needed, I add the needed tests.

lucaferranti:
The file was testing the more extreme corner case `cancelminus([M, M], [-M, -M])` where `M = prevfloat(Inf)` [here]https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/test/test_ITF1788/libieeep1788_cancel.jl#L103) I think that case was manually captured by [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/arithmetic.jl#L536) but that didn't generalize to other cases e.g. `M = 1e308` as reported in #508

lbenet:
You are right Luca: the case `cancelminus([M,M], [-M,-M])` with `M` *sufficiently large* (such that `prevfloat(Inf) < 2M`), was not captured consistently by those lines. So I've added few more tests.

Once tests pass, I'll merge and tag the new version.

PR #517, opened 2022-04-22
Fix acot and acoth with tests

lbenet:


lbenet:
Tests are still failing, but comparing [the current tests](https://github.com/JuliaIntervals/IntervalArithmetic.jl/runs/6135375890?check_suite_focus=true#step:6:1062) with those [of the 1.0-dev branch](https://github.com/JuliaIntervals/IntervalArithmetic.jl/runs/6099660580?check_suite_focus=true#step:6:3402), there is a clear improvement.

lucaferranti:
iirc acot and and acoth are not required by the standard, hence it's perfectly fine to return a valid interval, that is an interval which is a little wider than the tightest, as long as it contains it.

lucaferranti (approved):

PR #518, opened 2022-04-23
Return enum elements for overlap, and adapt ITF tests

lbenet:


src/IntervalArithmetic.jl
```diff
@@ -68,7 +68,10 @@ export
     cancelminus, cancelplus, isbounded, isunbounded,
     .., @I_str, ±,
     pow, extended_div, nthroot,
-    setformat, @format
+    setformat, @format,
+    OverlapType, _both_empty, _first_empty, _second_empty, _before,
```
lucaferranti, line 72: if they are exported, maybe it would make more sense to call them `bothEmpty`, `firstEmpty` and so on?

lucaferranti: from today meeting: the issue was that `contains` is a julia function

lbenet:
After today's meeting I've checked [EnumX](https://github.com/fredrikekre/EnumX.jl) and it looks pretty cool. 
- Considering adding it as a new dependence seems ok to me, in the sense that it is a pretty light package. 
- Using it would allow to export only the typename (`Overlap` is perhaps better than `OverlapType`), which brings along the different (enum) instances. 
- We could get rid of the odd naming I proposed initially, e.g. `_contains` or `_first_empty`, in favor of `Overlap.contains` or `Overlap.firstEmpty` (as you suggest) or `Overlap.first_empty`.

@lucaferranti What's you opinion on giving it a try?

lucaferranti:
@lbenet sorry for the delay in answering, I didn't notice the comment, I think we could give it a try, but does it have a significant advantage wrt returning a symbol?

lucaferranti:
well we can try, it's not a biggie :)

lbenet:
I tested using EnumX.jl: it corresponds to what we have in this PR, with a nicer interface, in the sense that only `Overlap` (the struct containing the overlap instances) is exported, instead of all specific instances. Shall I push the commit here so you can have a look and test it?

lucaferranti:
@lbenet sure!

lbenet:
I think this is ready @lucaferranti. Shall we backport it in an adapted form to master?

lucaferranti:
@lbenet LGTM, merging. don't have strong opinions on whether or not we should backport this to master and release a new patch release. If 1.0-dev is going to be merged "soon", then I think we can wait.

Issue #519, opened 2022-05-14
Extend inclusion check for interval and interval box


mforets:
```julia
julia> (0 .. 1) ⊆ IntervalBox(0 .. 1, 1)
ERROR: ArgumentError: [0, 1] ∈ [0, 1]¹ is not defined
Stacktrace:
 [1] in(X::IntervalArithmetic.Interval{Float64}, Y::IntervalBox{1, Float64})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/997vH/src/multidim/intervalbox.jl:82
 [2] issubset(l::IntervalArithmetic.Interval{Float64}, r::IntervalBox{1, Float64})
   @ Base ./abstractset.jl:293
 [3] top-level scope
   @ REPL[39]:1
```

lucaferranti:
what should `(0 .. 1) ⊆ IntervalBox(0 .. 1, 2)` return? Since `IntervalBox` uses `N` (the dimension) as parameter type, we could define

```julia
⊆(a::Interval, b::IntervalBox{1}) = a ⊆ first(b)
``` 

do you need this somewhere in JuliaReach?

mforets:
The solution you propose sounds good to me 👍 
(No need to implement it with `IntervalBox{N}` for `N` higher than one as the default MethodError is usually fine).

> do you need this somewhere in JuliaReach?

Yes, this is to avoid special casing inputs which can be 1-dim or mult-dim intervals.

PR #520, opened 2022-05-15
Update intervalbox.jl

mforets:
Fixes https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/519

---

A few tests could be added [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/test/multidim_tests/multidim.jl#L68), eg.

```julia
x = 0.5 .. 3
a = IntervalBox(A[1]) # 1 .. 2
@test !(x ⊆ a) && a ⊆ x
@test !(x ⊂ a) && a ⊂ x
@test x ∩ a == a ∩ x == A[1]
@test x ∪ a ==  a ∪ x == x
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "IntervalArithmetic"
 uuid = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 repo = "https://github.com/JuliaIntervals/IntervalArithmetic.jl.git"
-version = "0.20.5"
+version = "0.20.6"
 
 [deps]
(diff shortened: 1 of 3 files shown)


lucaferranti:
Could you add those tests and bump the patch version? 

In principle, if we accept those we should also support all other operations (arithmetic etc.) but it can be left for future PRs if/when someone needs those.

lucaferranti (approved):

lucaferranti:
@lbenet should we cherrypick this to 1.0-dev?

PR #522, opened 2022-05-21
fix parser

lucaferranti:


lucaferranti:
@lbenet this fixes almost all errors in the ITF1788 files (except for reverse functions). The remaining ones are not related to parsing but to the DecoratedInterval constructor and hence it might be better to address that in a separate PR.

lucaferranti:
ah also to note that atm the tests are not checking that warnings are thrown when they should. I would like to address warnings checks in a separate PR (after the constructors are fixed)

lucaferranti:
last but not least, if you are wondering where the green checks come from, that's because I temporarily disabled the tests for reverse functions... we know they are failing and there's an issue in IntervalContractors.jl, so better to leave those out for the time being.

lucaferranti:
@lbenet now **ALL** tests related to strings / parsing pass.

FYI the only failing ITF tests (except for reverse functions)
- 8 related to constructing decorated intervals, easy to fix, PR on the way already 💪 🚀
- one related to the accurate dot product

src/parsing.jl
```diff
@@ -32,125 +51,102 @@ in Section 9.7 of the IEEE Std 1788-2015, with some extensions.
...
+        ival, _ = _parse(Interval{T}, s)
+        return ival
+    catch
+        @warn "invalid input, empty interval returned"
+        return emptyinterval(T)
     end
+end
 
-    throw(ArgumentError("string \"$s\" can not be parsed to an interval."))
+function _parse(::Type{Interval{T}}, s::AbstractString) where T
```
Kolaru, line 66: Documenting here (e.g. in a docstring) what `isnotcom` do would be nice.

Also we could document (either here or in the docstring of `parse`) that `parse(Intrerval, s)` never errors, but `_parse(Interval, s)` does. The built-in `parse`, e.g. `parse(Int, "a")`, does error on invalid input, so we are doing something a bit special here.

src/parsing.jl
```diff
@@ -32,125 +51,102 @@ in Section 9.7 of the IEEE Std 1788-2015, with some extensions.
...
     try
-        return parse(p, s)
-    catch e
-        # We avoid CombinedParsers stacktraces because they are hard to read.
-        if !(e isa ArgumentError)
-            rethrow()
-        end
+        ival, _ = _parse(Interval{T}, s)
+        return ival
+    catch
```
Kolaru, line 60: Same as above, I recommend only catching `ArgumentError`.

Kolaru (changes requested):
It is great that the tests now pass =)

My comments are only about exception handling and documentation, all the rest looks good to me.

src/parsing.jl
lucaferranti: I have to confess I am not sure how to do it, is it just

```julia
catch e
  if e isa ArgumentError
    # return empty, NAI
  else
     retrow(e)
  end
end
```

lucaferranti: https://discourse.julialang.org/t/catching-exceptions-by-type/5690 pretty old, but seems so

Kolaru (approved):

test/interval_tests/parsing.jl
```diff
@@ -1,67 +0,0 @@
-using IntervalArithmetic
```
lbenet, line 1: I notice that you have removed the parsing tests (this file). While I a-priori might agree, the question is: are all instances covered by other (ITF1788) tests? Maybe these tests should be simply re-adapted to test other numeric formats, e.g. `Float32`.

test/test_ITF1788/run_ITF1788.jl
```diff
@@ -14,8 +14,9 @@ include("libieeep1788_reduction.jl")
 include("libieeep1788_set.jl")
 include("mpfi.jl")
 
-using IntervalContractors
-include("abs_rev.jl")
-include("libieeep1788_rev.jl")
-include("libieeep1788_mul_rev.jl")
-include("pow_rev.jl")
+# TODO: CHECK CONTRACTORS
```
lbenet, line 17: I think the issue here is that `IntervalContractors.jl` uses certain functionality that has changed in 1.0-dev branch.

src/parsing.jl
```diff
@@ -1,156 +1,241 @@
...
+            If I try to give a decoration that is too high, e.g. [1, Inf]_com, then it
+            should error and return [NaI]. Exception to this is if the interval would be com
+            but becomes dac because of finite precision, e.g. "[1e403]_com" when parse to
+            Interval{Float64} is allowed to become [prevfloat(Inf), Inf]_dac without erroring.
+            The isnotcom flag returned by _parse is used to track if the interval was originally
+            smaller than com or became dac because of overflow.
+        =#
+        dec_given > dec_calc && isnotcom && throw(ArgumentError("invalid decoration $dec for $ival"))
+
+        return DecoratedInterval(ival, min(dec_given, dec_calc))
```
lbenet, line 64: I guess it would be better here to explicitly return `DecoratedInterval{T}(...)`, perhaps (but not sure) for type-stability issues.

src/parsing.jl
```diff
@@ -1,156 +1,241 @@
...
+                lo = T((m - v * ulp) * e, RoundDown)
+                hi = T((m + v * ulp) * e, RoundUp)
+            end
+        end
+    else # number
+        lo = parse_num(T, s, RoundDown)
+        hi = parse_num(T, s, RoundUp)
     end
+    is_valid_interval(lo, hi) && return Interval(lo, hi), isnotcom
+    throw(ArgumentError("input $s can not be parsed as an interval."))
```
lbenet, line 229: Is this equivalent to 
```julia
return interval(lo, hi), isnotcom
```
?

Or is there a case I'm missing?

lbenet (approved):
Thanks @lucaferranti for addressing this! In general, LGTM, and I'm in favor of merging this. 

I left few comments that are optional. Perhaps the only relevant is related to the tests, which instead of being removed could be adapted to test *other* numeric formats (Float32, etc).

src/parsing.jl
```diff
@@ -1,156 +1,241 @@
...
+            The isnotcom flag returned by _parse is used to track if the interval was originally
+            smaller than com or became dac because of overflow.
+        =#
+        dec_given > dec_calc && isnotcom && throw(ArgumentError("invalid decoration $dec for $ival"))
+
+        return DecoratedInterval(ival, min(dec_given, dec_calc))
+    catch e
+        if e isa ArgumentError
+            @warn "invalid input, returning [NaI]"
+            return nai(T)
```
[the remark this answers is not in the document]
lucaferranti: I don't think the warning should be inside `nai()`, as that would emit the warning also when the user explicitly asks for `NaI`. See

```julia
julia> nai()
[NaN, NaN]_ill

julia> parse(DecoratedInterval{Float64}, "[NaI]")
[NaN, NaN]_ill

julia> parse(DecoratedInterval{Float64}, "foobar")
┌ Warning: invalid input, returning [NaI]
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/parsing.jl:67
[NaN, NaN]_ill
```

(unrelated to this PR, but I don't like the current printing of NaI currently, I think it should print `[NaI]` )

test/interval_tests/parsing.jl
lucaferranti: in principle ITF1788 can generate tests for other number format (the original python script not mine). Directly using the testsuite for other number format probably will not work (double rounding for Float32, BigFloat would probably be tighter). 

I heard rumors there's some work for a unified testing framework ;) meanwhile, I'll reinsert some manual tests for other formats

src/parsing.jl
lucaferranti: I don't know :D 
if `ival` is already of type `Interval{T}` then it should be type stable I guess?

src/parsing.jl
lucaferranti: the reason for this is that `_parse` is also called by `parse(DecoratedInterval{Float64}, s)`, using simply `interval` would return `[empty]_trv` instead of `[NaI]` in that case.

src/parsing.jl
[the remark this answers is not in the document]
lbenet: > I don't think the warning should be inside `nai()`, as that would emit the warning also when the user explicitly asks for `NaI`

Fine then with me.

> I don't like the current printing of `NaI` currently, I think it should print `[NaI]`

I may agree also with that.

test/interval_tests/parsing.jl
lbenet: While you may be right with respect to ITF1788, I don't think "other formats" includes extended precision, simply because the "native" way of coding those tests is `Float64`. In any case, reinserting those tests with another format was my point

src/parsing.jl
lbenet: You are right...

src/parsing.jl
lbenet: Understood... Then leave it as it is.

lucaferranti:
@lbenet now added some tests for `float32` and `bigfloat`

lbenet (approved):
LGTM! Thanks a lot!

lbenet:
Merged!

It's **really** nice to see again green colors in the tests!

PR #523, opened 2022-05-22
update pointwise policy

lucaferranti:
### Description
- fix typo politics --> policy
- updated docstring, ieee is the default behavior (at least atm)

### Notes
- It feels a little non-intuitive why it's called pointwise, could it be renamed to BooleanPolicy?
- I would also suggest to rename `all-in` to `certain` or `certainly`, reflecting the naming convections of Boost and Filib++.
- we could also rename `:interval` --> `:booleaninterval`, it's a little longer but maybe clearer (and reflects the name of the struct). if length is deal break, one could use singleton types instead of symbols and get autocompletition

### Other notes
(that is, probably better suited for a different PR

- I am not sure why currently `BooleanInterval(true, true)` and `BooleanInterval(false, false)` are disabled, but I think they should be allowed. Also, since generally [false] = 0 and [true] = 1, (where [ ] are [iverson brackets](https://en.wikipedia.org/wiki/Iverson_bracket) ), it should probably be `BooleanInterval(false, true)`
- at some point we could also add `possible` or `possibly` policy (also preseant in Filib++ and boost), which is basically map missing --> true
 
to summarize, except for ieee which is different, the others could be based on ternary (as kolaru did), and would just be the mappings
- `certainly`: 
  - true --> true
  - false --> false
  - missing --> false
- `possibly`: 
  - true --> true
  - false --> false
  - missing --> true
- `booleaninterval`: 
   - true --> [true, true]
   - false --> [false, false]
   - missing -> [false, true]


Kolaru:
> * It feels a little non-intuitive why it's called pointwise, could it be renamed to BooleanPolicy?

I blame the standard for the non-intuitiveness 😬.

There are two competing things here:
1. A use of the *symbols* usually reserved for boolean operation. That's what the standard provides.
2. An extension of the boolean *operations* themselves. That's where the name "pointwise" comes from, as the natural extension is to compare the interval pointwise.

Initially I was thinking to only adress the second one. If we use the first interpretation instead, `BooleanPolicy` (or even `BooleanOperatorsPolicy`) is indeed a better name.

> * I would also suggest to rename `all-in` to `certain` or `certainly`

Sounds nicer indeed, I think we should go for it.

> `booleaninterval`:
> * true --> [true, true]
> * false --> [false, false]
> * missing -> [false, true]

The reason I disallowed `BooleanInterval(true, true)` is because I was thinking in term of sets, so that `[true, true] == [true] == {true}`. I don't have very strong opinion here, the main point of this policy (that I can see ^^') is to always error when passed into a `if` conditional statement.

src/intervals/interval_operations/pointwise_boolean.jl
```diff
@@ -1,39 +1,27 @@
...
-    This is different from what is described in sections 9.5 of the
-    IEEE Std 1788-2015 (Boolean functions of intervals), since we deal here
-    with what is needed for drop-in replacement of floats by intervals,
-    and not what is reasonnable to do with intervals in a separate execution
-    environment.
-
-    Essentially, julia is better at composability and we are more ambitious
-    that what the standard ever expected, so we have some extra problems
-    to solve.
+#=  This file contains boolean operations defined for intervals.
```
lbenet, line 3: I think we should adapt part of the explanation of @Kolaru that was removed here. First, the `PointwisePolicy` does allow to consider other behavior for the boolean functions on intervals, which does not follow the specifications of the IEEE Std 1788-2015. (This may be interpreted as another flavor, e.g. cset + PoinwisePolicy.) And secondly, because Benoit is right, Julia is better at composability and we are ambitious :-)

src/intervals/interval_operations/pointwise_boolean.jl
```diff
@@ -1,39 +1,27 @@
...
-    PointwisePolitic{P}
+    PointwisePolicy{P}
 
-Define which politic we use to extend pointwise comparison of
+Define which policy we use to extend pointwise comparison of
 
-Valid value for the politic identifier `P` are
+Valid value for the policy identifier `P` are
+    - `ieee1788` (default): Boolean operations as defined in the standard,
+      see section 10.5.10
```
lbenet, line 19: Maybe `:ieee1788` (which is too long) should be renamed to `:ieee`, or allow it as an alias.

lbenet (commented):
In principle, LGTM. 

Yet, I've left two comments: the first one is part of a removed comment explaining the motivation, which I think it should be back perhaps in an adapted form. The second is actualy a suggestion, and is related to the name for the IEEE Std 1788-2015 policy: while `:ieee1788` is fine, it is also long, so maybe we should have `:ieee` as an alias.

src/intervals/interval_operations/pointwise_boolean.jl
```diff
@@ -1,39 +1,32 @@
...
-    and not what is reasonnable to do with intervals in a separate execution
-    environment.
-
-    Essentially, julia is better at composability and we are more ambitious
-    that what the standard ever expected, so we have some extra problems
-    to solve.
+    As there is no common consensus on how these operations should be defined for intervals,
+    in addition to the behavior defined by the standard (sec. ), the file also implements
+    different policies to handle these operations, giving the user to choose the one most
+    suited for their use case.
```
lbenet, line 8: Benoit refers (systematically?) to the standard as "IEEE Std 1788-2015"; I think we should follow that. Is the section number you refer to, section 9.5? (Sorry for being too picky.)

I must admit that I really like the last paragraph you deleted: "Essentially, julia is better at composability and we are more ambitious that what the standard ever expected, so we have some extra problems to solve."

src/intervals/interval_operations/pointwise_boolean.jl
```diff
@@ -1,39 +1,32 @@
...
 """
-    PointwisePolitic{P}
+    PointwisePolicy{P}
 
-Define which politic we use to extend pointwise comparison of
+Define which policy we use to extend pointwise comparison of
 
-Valid value for the politic identifier `P` are
-    - `:is_all` : A boolean operation is extended by asking "is it true for
+Valid value for the policy identifier `P` are
```
lbenet, line 22: value -> values

src/intervals/interval_operations/pointwise_boolean.jl
```diff
@@ -43,26 +36,26 @@ Valid value for the politic identifier `P` are
         possible outcome as a `BooleanInterval`.
         This is safe but very strict, always erroring when an interval is used
         in a conditional statement.
-    - `:ternary` (default) : With this politics return `missing` when the
+    - `:ternary`: With this policys return `missing` when the
```
lbenet, line 39: policys -> policy

lbenet (approved):
LGTM!

Left three comments (the first one is maybe too picky), and the other two are related to typos. Please, go ahead and merge this PR once the typos are corrected.

src/intervals/interval_operations/pointwise_boolean.jl
lucaferranti: > Benoit refers (systematically?) to the standard as "IEEE Std 1788-2015"; I think we should follow that. Is the section number you refer to, section 9.5? (Sorry for being too picky.)

hups sorry I left that hanging :)

> I must admit that I really like the last paragraph you deleted: "Essentially, julia is better at composability and we are more ambitious that what the standard ever expected, so we have some extra problems to solve."

Personally, I find the paragraph a little out of place here. Of course I think julia is cool and great and composability, but the right place for such a claim would be in a tutorial notebook, showing with some examples how one policy fails when combining with another package and another gives the desired results. API docs should stick to the raw facts and implementation details. I also fear it's a little misleading, the "we are more ambitious and have extra problems to solve" sounds to me like in the file new solutions to more advanced problems are proposed, while actually the other non-ieee policies are actually older and what other interval arithmetic libraries (filib, gaol, boost) already did before the standard.

PR #525, opened 2022-05-24
Simplified version of mod(x::Interval, y::Real)

### LINKED ISSUE #129: What to do with mod?

petvana:
This PR cherry-picks `mod(x::Interval, y::Real)` from #178 because it seems to be the most useful case (at least for me), and also easiest to implement.

Cc: @dpsanders @lbenet 

fixes #129

### DIFF
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -378,7 +378,7 @@ end
 Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
 """
 function mod(x::Interval, y::Real)
-    @assert y > 0 "modulo is currently implemented only for a positive divisor."
+    @assert y > zero(y) "modulo is currently implemented only for a positive divisor."
     division = x / y
     fl = floor(division)
     fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -373,3 +373,13 @@ function nthroot(a::Interval{T}, n::Integer) where T
     b = nthroot(bigequiv(a), n)
     return convert(Interval{T}, b)
 end
+
+"""
+Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
+"""
+function mod(x::Interval, y::Real)
+    @assert y > 0 "modulo is currently implemented only for a positive divisor."
+    division = x / y
+    fl = floor(division)
+    fl.lo < fl.hi ? 0..y : y * (division - fl)
+end
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -375,13 +375,18 @@ function nthroot(a::Interval{T}, n::Integer) where T
 end
 
 """
-Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
+Calculate `x::Interval mod y::Real`, limited by `y != 0`.
 """
 function mod(x::Interval, y::Real)
-    @assert y > zero(y) "modulo is currently implemented only for a positive divisor."
+    @assert y != zero(y) """mod(x::Interval, y::Real) 
+is currently implemented only for a strictly positive or negative divisor y."""
     division = x / y
     fl = floor(division)
-    fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
+    if !isthin(fl)
+        return y > zero(y) ? Interval(zero(y), y) : Interval(y, zero(y))
+    else
+        return y * (division - fl)
+    end
 end
 
 mod(x:T, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -375,7 +375,7 @@ function nthroot(a::Interval{T}, n::Integer) where T
 end
 
 """
-Calculate `x::Interval mod y::Real`, limited by `y != 0`.
+Calculate `x::Interval mod y::Real` where y != zero(y) and y is not inteval`.
 """
 function mod(x::Interval, y::Real)
     @assert y != zero(y) """mod(x::Interval, y::Real) 
@@ -389,4 +389,5 @@ is currently implemented only for a strictly positive or negative divisor y."""
     end
 end
 
-mod(x:T, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
+mod(x::Interval, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
+mod(x::Real, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -381,5 +381,5 @@ function mod(x::Interval, y::Real)
     @assert y > 0 "modulo is currently implemented only for a positive divisor."
     division = x / y
     fl = floor(division)
-    fl.lo < fl.hi ? 0..y : y * (division - fl)
+    fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
 end
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -383,3 +383,5 @@ function mod(x::Interval, y::Real)
     fl = floor(division)
     fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
 end
+
+mod(x:T, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -389,5 +389,5 @@ is currently implemented only for a strictly positive or negative divisor y."""
     end
 end
 
-mod(x::Interval, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
-mod(x::Real, y::Interval) where T = throw(ArgumentError("mod not defined for interval as divisor `y`"))
+mod(x::Interval, y::Interval) = throw(ArgumentError("mod not defined for interval as divisor `y`"))
+mod(x::Real, y::Interval) = throw(ArgumentError("mod not defined for interval as divisor `y`"))
--- a/src/intervals/functions.jl
+++ b/src/intervals/functions.jl
@@ -9,7 +9,7 @@
 for T in (:Integer, :Float64, :BigFloat, :Interval)
     @eval ^(a::Interval{Float64}, x::$T) = atomic(Interval{Float64}, bigequiv(a)^x)
 end
-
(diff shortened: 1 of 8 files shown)


src/intervals/functions.jl
```diff
@@ -373,3 +373,13 @@ function nthroot(a::Interval{T}, n::Integer) where T
...
 end
+
+"""
+Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
+"""
+function mod(x::Interval, y::Real)
+    @assert y > 0 "modulo is currently implemented only for a positive divisor."
+    division = x / y
+    fl = floor(division)
+    fl.lo < fl.hi ? 0..y : y * (division - fl)
```
lucaferranti, line 384:
```suggestion
    fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
```

For internal use, the `Interval` constructor is better. The others (`interval`, `..`) are for end users and have some extra checks.

src/intervals/functions.jl
```diff
@@ -373,3 +373,13 @@ function nthroot(a::Interval{T}, n::Integer) where T
...
 end
+
+"""
+Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
+"""
+function mod(x::Interval, y::Real)
+    @assert y > zero(y) "modulo is currently implemented only for a positive divisor."
+    division = x / y
+    fl = floor(division)
+    fl.lo < fl.hi ? Interval(zero(y), y) : y * (division - fl)
```
lbenet, line 384: Picky suggestions: 
- instead of using `fl.lo` and `fl.hi`, can you use `inf(fl)` and `sup(fl)`? As an alternative, you could test `!isthin(fl)` instead of checking that the bounds are different (`fl.lo < fl.hi`).
- include `return `. 

(These are things we are trying to have thoroughly for version 1.0.)

src/intervals/functions.jl
```diff
@@ -373,3 +373,13 @@ function nthroot(a::Interval{T}, n::Integer) where T
     b = nthroot(bigequiv(a), n)
     return convert(Interval{T}, b)
 end
+
+"""
+Calculate `x mod y` where `x` is an interval and `y` is a positive divisor.
+"""
+function mod(x::Interval, y::Real)
+    @assert y > zero(y) "modulo is currently implemented only for a positive divisor."
```
lbenet, line 381: Is it possible to extend this to have strictly negative `y`? I understand that having `0 ∈ y` complicates things....

test/interval_tests/numeric.jl
```diff
@@ -434,3 +434,30 @@ end
     @test nthroot(Interval{BigFloat}(-81, -16), -4) == ∅
     @test nthroot(Interval{BigFloat}(-81, -16), 1) == Interval{BigFloat}(-81, -16)
 end
+
+# approximation used for testing (not to rely on ≈ for intervals)
+# ⪆(x, y) = (x ≈ y) && (x ⊇ y)
+⪆(x::Interval, y::Interval) = x.lo ≈ y.lo && x.hi ≈ y.hi && x ⊇ y
```
lbenet, line 440: I think `x ≈ y` works for intervals (instead of `x.lo ≈ y.lo && x.hi ≈ y.hi`)...

test/interval_tests/numeric.jl
```diff
@@ -434,3 +434,30 @@ end
...
+
+@testset "`mod`" begin
```
lbenet, line 442: Could you also add tests where `y` (the second argument) is an interval?

lbenet (commented):
Thanks a lot for addressing #129. 

In general the pull-request LGTM! I left some suggestions, which I think are easy to implement.

test/interval_tests/numeric.jl
petvana: Unfortunately not
``` julia
julia> 1.00000001 ≈ 1.0
true

julia> 1.0000000001..2.0 ≈ 1.0..2.0
false

```

lbenet: Good point!

petvana: Ref #479 #480

petvana: Also, the returned interval should be at least as wide as the correct result.

test/interval_tests/numeric.jl
petvana: It's not yet implemented, so I've added a todo and test that it throws `TypeError`.

src/intervals/functions.jl
petvana: Updated for strictly negative divisor. Hope its correct.

lbenet: The implementation seems to me correct if we have `y::AbstractFloat` (perhaps also including Rationals, but let me forget about them now), which corresponds to the tests.

However, if `y::Interval` (and we have `Interval <: Real`) then there this function throws errors: e.g., try `mod(1..2, 1..2)`, which I think should return `Interval(0.0, 2.0)`. In this case things are subtle, because `y != zero(y)` is true for `[-1, 1]`, but that interval is not strictly positive or negative. Also, `Interval(zero(y), y)` causes the error mentioned above. I guess this is the reason that `mod` has two methods in #178.

My suggestion is either restrict `y::AbstractFloat`, or include a new method where `y::Interval`.

petvana: This is my mistake. My intention was to constrain `y` NOT to be `Interval`. I've not realized `1..3 isa Real -> true`. However, If I constraint it to `y::AbstractFloat` it does not accept integers for example because `1 isa AbstractFloat -> false`, right?

test/interval_tests/numeric.jl
```diff
@@ -434,3 +434,43 @@ end
...
+    @test mod(x, 2) == mod(x, 2.0) == 0..2
+    @test mod(x, 2.5) == 0..2.5
+    @test mod(x, 0.5) == 0..0.5
+    @test mod(x, -1) == mod(x, -1.0) == -1..0
+    @test mod(x, -2) == mod(x, -2.0) == -2..0
+    @test mod(x, -2.5) == -2.5..0
+    @test mod(x, -0.5) == -0.5..0
+
+    # TODO - implement mod for two intervals
+    @test_throws TypeError mod(1..2, 1.4..1.5)
```
lucaferranti, line 475: this is causing problems on nightly. I think it's better not to rely on what exception the julia internals throw, as they may change (as this example shows), see [this comment](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/525#discussion_r883348123) .

src/intervals/functions.jl
```diff
@@ -373,3 +373,18 @@ function nthroot(a::Interval{T}, n::Integer) where T
     b = nthroot(bigequiv(a), n)
     return convert(Interval{T}, b)
 end
+
+"""
+Calculate `x::Interval mod y::Real`, limited by `y != 0`.
+"""
+function mod(x::Interval, y::Real)
+    @assert y != zero(y) """mod(x::Interval, y::Real)
```
lucaferranti, line 381: an alternative would be to return an empty interval, I think this would be more conformant with the IEEE standard behavior, e.g.

```julia
julia> sqrt(-1.. -1)
∅

julia> log(-2.. -1)
∅
```

petvana: Or we can define `mod(a..b, 0) == 0..0 ⊇ ∅` as Wolfram says
https://www.wolframalpha.com/input?i=limit+x-%3E0%2B+mod%281%2C+x%29

Consequently, for `0 ∈ y`, we can define smth like `mod(x,y) = union(mod(x,(y.lo..y.lo/3)), mod(x, (y.hi/3)..(y.hi)), 0..0)` because the maximum value needs to by in `(y.hi/2)..(y.hi)`

test/interval_tests/numeric.jl
petvana: Solved by defining `ArgumentError` explicitly, as you proposed.

src/intervals/functions.jl
petvana: I propose to solve this in some future PR, and move the discussion to #129, to keep track once this PR merged.

src/intervals/functions.jl
lbenet: `mod` can indeed be generalized for `y::Interval`. It's tricky with respect of having zero within the interval, which is part of the reason I was suggesting to have either a strictly positive or negative `y`. Actually, a motivating example would be to have `y::Irrational`, or actually any (mathematical) real number which is not exactly representable as a `Float64`. Note that #178 includes such an implementation, though it does not include the restriction of strictly positive/negative intervals. The subtleties related to zero at the end are related to the division: `y` appears in the denominator.

lbenet: I unresolved this conversation, so it is easy to track the discussion...

lucaferranti: fwiw in octave

```
>> mod(infsup(-3, 2), infsup(-1, 0.5))
ans = [-1, +0.5]

>> mod(infsup(-3, 2), infsup(0, 0))
ans = [Empty]

>> mod(infsup(-3, 2), infsup(0, 3))
ans = [0, 3]

>> mod(infsup(-3, 2), infsup(1, 3))
ans = [0, 3]

>> mod(infsup(-3, 2), infsup(-3, 2))
ans = [-3, +2]
```

src/intervals/functions.jl
lbenet: > Or we can define `mod(a..b, 0) == 0..0 ⊇ ∅` as Wolfram says https://www.wolframalpha.com/input?i=limit+x-%3E0%2B+mod%281%2C+x%29

Julia (and Wolfram) return `NaN` (`undefined`) for `mod(3.2, 0.0)`, which makes sense due to the division in the algorithm.

> 
> Consequently, for `0 ∈ y`, we can define smth like `mod(x,y) = union(mod(x,(y.lo..y.lo/3)), mod(x, (y.hi/3)..(y.hi)))` because the maximum value needs to by in `(y.hi/2)..(y.hi)`

This idea is interesting, and reminds me a little bit what we have for `extended_div`. That certainly should be addressed in another PR. Yet, I think for (mathematical) reals, `mod(x,y)` is a number in the interval [0,y] for y>0, or [y, 0] if y<0, isn't it?

From the range of `mod(x,y)` for real `y`, then for a strictly positive interval `y`, we should return `[0, sup(y)]`, or if it is strictly negative `[inf(y), 0]`. If `0 ∈ y`, we should return the hull of these intervals.

lbenet: I think my last comment corresponds to [these results](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/525#discussion_r883635954)... Or not?

Kolaru:
Sorry for the extended delay.

Since all tests are passing and all discussions are passing, I am merging. Thanks a lot for your contribution !

PR #527, opened 2022-05-26
update decorated intervals constructors

lucaferranti:
## What

This PR fixes issues in 1.0-dev with decorated interval constructors. After this all related ITF tests pass.

### Check for validity of interval

Added inner constructor that checks if the given decoration is appropriate and behaves accordingly. Also for `ill` the interval part is normalized to `[NaN, NaN]`, which simplifies computations since `NaN` poisons everything

- before
```julia
julia> DecoratedInterval(Interval(0, Inf), com)
[0, ∞]_com

julia> DecoratedInterval(Interval(2, 1), com)
[2, 1]_com

julia> DecoratedInterval(Interval(0, Inf), com)
[0, ∞]_com

julia> DecoratedInterval(Interval(2, 1), com)
[2, 1]_com

julia> a = DecoratedInterval(2, 1)
[2, 1]_ill

julia> inf(a)
2.0
```

- after
```julia
julia> DecoratedInterval(Interval(0, Inf), com)
[0, ∞]_dac

julia> DecoratedInterval(Interval(2, 1), com)
[NaI]

julia> ans.interval
[NaN, NaN]

julia> DecoratedInterval(2, 1)
[NaI]

julia> ans.interval
[NaN, NaN]

julia> inf(a)
NaN
```

there was (and still is) some asymmetry between constructors of decorated intervals and bare intervals. For bare intervals, we have the unsafe constructor `Interval` which does nothing, the safe constructor `interval` and the safe macro with correct rounding `@interval`. For decorated intervals we have the safe macro with correct rounding interval `@decorated`, but only one constructor `DecoratedInterval`, which before this PR was unsafe and after this PR is safe (at the price of extra overhead). Are we happy with this? Or should we have a safe `decorated` constructor and unsafe `DecoratedInterval` constructor? I don't have strong opinions about this.

### Interval part of nai()

according to the standard, the interval part of the `[NaI]` should be the empty interval and a warning should be thrown. This mainly had a few consequences
  1. for a few functions (e.g. inf, sup) `interval` could not be used directly, since `inf(nai())` should still be `NaN`
  2. Now all tests pass, so I guess it's fine but as a drawback, warnings are emitted in the middle
  ```julia
julia> nai() + DecoratedInterval(1, 2)
┌ Warning: trying to access interval part of [NaI], returning empty interval
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/decorations/intervals.jl:63
[NaI]
```
is this fine? The alternatives are
1. use `x.interval` internally. This would probably work but might be a little contradictory with the latest decision to use `inf, sup` internally instead of `.lo, .hi`. On the other side, This matters only when implementing operations for decorated intervals but the end user who writes some generic algorithm with intervals would not note the difference, as would write generic code for `AbstractInterval` and hence would not need to access the interval part.
2. have an interval function `_interval(x::DecoratedInterval) = x.interval` and external `interval` that does the "standard" behavior. Alternatively the internal could be called `interval` and external `intervalPart` or viceversa. 

Personally, I think option 1 would be better in this case.

### printing

- changed printing on `[NaI]` now it's displayed as `[NaI]`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,7 +1,11 @@
 name: CI
 on:
-  - push
-  - pull_request
+  push:
+    branches: "main"
+    tags: ["*"]
+  pull_request:
+  release:
+
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
--- a/src/decorations/decorations.jl
+++ b/src/decorations/decorations.jl
@@ -1,14 +1,14 @@
 include("intervals.jl")
 include("functions.jl")
 
-isnan(x::Interval) = false  # NaI is always decorated
+# isnan(x::Interval) = false  # NaI is always decorated
 
 """`NaI` not-an-interval: [NaN, NaN]."""
-nai(::Type{T}) where T = DecoratedInterval(convert(T, NaN), convert(T, NaN), ill)
+nai(::Type{T}) where T = DecoratedInterval(Interval(convert(T, NaN), convert(T, NaN)), ill)
 nai(::Type{F}) where {T, F<:Interval{T}} = nai(T)
 nai(::Interval{T}) where T<:Real = nai(T)
 nai(::DecoratedInterval{T}) where T<:Real = nai(T)
-nai() = nai(Interval{default_bound()})
+nai() = nai(default_bound())
 
-isnai(x::Interval) = isnan(x.lo) || isnan(x.hi) #|| x.lo > x.hi || (isinf(x.lo) && x.lo == x.hi)
-isnai(x::DecoratedInterval) = isnai(interval(x)) || x.decoration == ill
+isnai(x::Interval) = false  # NaI is always decorated
+isnai(x::DecoratedInterval) = x.decoration == ill
--- a/src/decorations/functions.jl
+++ b/src/decorations/functions.jl
@@ -23,7 +23,7 @@ const bool_binary_functions = (
 )
 
 for f in bool_functions
-    @eval $(f)(xx::DecoratedInterval) = $(f)(interval(xx))
+    @eval $(f)(xx::DecoratedInterval) = $(f)(xx.interval)
 end
 
 for f in bool_binary_functions
@@ -42,7 +42,7 @@ scalar_functions = (
 )
 
 for f in scalar_functions
-    @eval $(f)(xx::DecoratedInterval{T}) where T = $f(interval(xx))
+    @eval $(f)(xx::DecoratedInterval{T}) where T = $f(xx.interval)
 end
 
 dist(xx::DecoratedInterval, yy::DecoratedInterval) = dist(interval(xx), interval(yy))
--- a/src/decorations/intervals.jl
+++ b/src/decorations/intervals.jl
@@ -33,7 +33,7 @@ struct DecoratedInterval{T}
     function DecoratedInterval{T}(I::Interval, d::DECORATION) where T
         dd = min(d, decoration(I))
         dd == ill && return new{T}(Interval(T(NaN), T(NaN)), ill)
-        return new{T}(I, min(d, dd))
+        return new{T}(I, dd)
     end
 end
 
--- a/src/decorations/intervals.jl
+++ b/src/decorations/intervals.jl
@@ -29,44 +29,51 @@ of a previously executed sequence of functions acting on an initial interval.
 struct DecoratedInterval{T}
     interval::Interval{T}
     decoration::DECORATION
+
+    function DecoratedInterval{T}(I::Interval, d::DECORATION) where T
+        dd = min(d, decoration(I))
+        dd == ill && return new{T}(Interval(T(NaN), T(NaN)), ill)
+        return new{T}(I, min(d, dd))
+    end
 end
+
+DecoratedInterval(I::Interval{T}, d::DECORATION) where T = DecoratedInterval{T}(I, d)
 
 DecoratedInterval(I::DecoratedInterval, dec::DECORATION) = DecoratedInterval(I.interval, dec)
 
-function DecoratedInterval(a::Real, b::Real, d::DECORATION)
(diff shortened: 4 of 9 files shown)

### CODE REVIEW

**Comment by lbenet** [comment]
Let me add few thoughts about `Decoratedintervals`, and then this PR:

- There is some an implicit dependence on the flavor which should be made explicit here: Concretely, in what is a valid interval. For example, if we (would) have a flavor that considers as a valid interval `[2,1]`, then `decoration( DecoratedInterval(interval(2,1)) ) == def` should be `true`, and it should not return a `nai()`. (I think in this case the decoration should be at most `def`, but I may be wrong.)
- Regarding the asymmetry, I think we **must** check the validity of the interval for `DecoratedInterval`, and as mentioned, this is flavor dependent. This is then tricky and unconfortable, in the sense we should use `interval` instead of `Interval`...
- Regarding the interval part of `nai()`, if it is supposed to be the empty interval, then `interval(nai()) == emptyinterval()` should be `true`. Then, this imposes to also have that `[NaN,NaN]` corresponds to the empty interval, and I guess that `[Inf, Inf]` or [`-Inf, -Inf]` should behave similarly when the intervals are decorated. As a side note, let me note that e.g. `interval(Inf,Inf)` returns the empty interval.

File: src/decorations/intervals.jl [hunk 1]
```diff
@@ -29,44 +29,51 @@ of a previously executed sequence of functions acting on an initial interval.
 struct DecoratedInterval{T}
     interval::Interval{T}
     decoration::DECORATION
+
+    function DecoratedInterval{T}(I::Interval, d::DECORATION) where T
+        dd = min(d, decoration(I))
+        dd == ill && return new{T}(Interval(T(NaN), T(NaN)), ill)
+        return new{T}(I, min(d, dd))
```
lucaferranti, line 36:
```suggestion
        return new{T}(I, dd)
```

src/decorations/decorations.jl
```diff
@@ -1,14 +1,14 @@
 include("intervals.jl")
 include("functions.jl")
 
-isnan(x::Interval) = false  # NaI is always decorated
+# isnan(x::Interval) = false  # NaI is always decorated
```
Kolaru, line 4:
```suggestion

```

src/decorations/intervals.jl
```diff
@@ -29,44 +29,51 @@ of a previously executed sequence of functions acting on an initial interval.
 struct DecoratedInterval{T}
     interval::Interval{T}
     decoration::DECORATION
+
+    function DecoratedInterval{T}(I::Interval, d::DECORATION) where T
```
Kolaru, line 33: A small note in the docstring to tell that the decoration is always checked at creation would be good.

src/decorations/intervals.jl
```diff
@@ -29,44 +29,51 @@ of a previously executed sequence of functions acting on an initial interval.
...
-end
+DecoratedInterval(a::Real, b::Real) = DecoratedInterval(Interval(a,b))
 
 DecoratedInterval(a::Real) = DecoratedInterval(a, a)
 DecoratedInterval(a::Tuple) = DecoratedInterval(a...)
 
-interval(x::DecoratedInterval) = x.interval
+function interval(x::DecoratedInterval{T}) where {T}
+    if isnai(x)
+        @warn "trying to access interval part of [NaI], returning empty interval"
```
Kolaru, line 63:
```suggestion
        @warn "trying to access interval part of [NaI], returning empty interval. Use `x.interval` for unsafe and unchecked access."
```

Kolaru (approved):
Not my main area of expertise, so all I have are very minor doc changes.

.github/workflows/CI.yml
```diff
@@ -1,7 +1,11 @@
 name: CI
 on:
-  - push
-  - pull_request
+  push:
+    branches: "main"
+    tags: ["*"]
+  pull_request:
+  release:
```
lbenet, line 7: Simply asking if this is needed? I mean, we have in this repo the "master" branch (old scheme) rather than the "main" branch (new scheme).

src/decorations/functions.jl
```diff
@@ -23,7 +23,7 @@ const bool_binary_functions = (
 )
 
 for f in bool_functions
-    @eval $(f)(xx::DecoratedInterval) = $(f)(interval(xx))
+    @eval $(f)(xx::DecoratedInterval) = $(f)(xx.interval)
```
lbenet, line 26: Question: why do you prefer to use `xx.interval` instead of the helper function `interval(xx)`?

src/decorations/intervals.jl
lbenet: Maybe start the sentence with a capital letter... (Sorry for being too picky.)

test/decoration_tests/decoration_tests.jl
```diff
@@ -63,6 +67,12 @@ let b
     @test isnai(@decorated(NaN, 3))
     @test isnai(@decorated(3, NaN))
     @test isnai(@decorated(NaN, NaN))
+
+    @test !isnai(Interval(1, 2))
```
lbenet, line 71: Not relevant for this PR, but we should also test the warnings being emitted...

test/interval_tests/consistency.jl
```diff
@@ -165,7 +165,7 @@ using Test
         @test !(nai(a) ≛ nai(a))
         @test nai(a) === nai(a)
         @test nai(Float64) === DecoratedInterval(NaN)
-        @test isnan(interval(nai(BigFloat)).lo)
+        @test isnan(inf(nai(BigFloat)))
```
lbenet, line 168: Suggestion: `@test isnan(inf(nai(BigFloat))) && isnan(sup(nai(BigFloat)))`

lbenet (commented):
Thanks @lucaferranti for addressing this! 

I've left a couple of suggestions and questions. Aside from them, LGTM.

@lucaferranti Let me know if I you have time to address @Kolaru and my suggestions, or if I should go ahead and push the changes.

OlivierHnt:
Superseded by #590.

PR #529, opened 2022-05-27
fix rational power bug in 1.0

lucaferranti:
## Summary

This PR fixes two bugs in  `^(::Interval, ::Rational)`

- Due to the style change `Interval{T} where {T}` --> `F where F<:Interval`, 1.0-dev had introduced a small bug with rational powers of intervals, this went unnoticed because the corresponding test file wasn't run. I noticed this while updating `IntervalContractors.jl` for 1.0-dev. This PR fixes it
- Using `Int64` instead of `BigInt` errors because of overflow if very big rationals are used. This uses `BigInt` instead of `Int64` it's probably slower, but if you use rationals it means in the first place you prefer exactness over efficiency.


lbenet:
Thanks @lucaferranti for addressing this!

Rather than reviewing your PR, I simply pushed a commit which I think is in the spirit of many changes introduced by @Kolaru. The bottom line is: I don't think we need to go to `BigInt`s (tests pass in master!) 😄

lbenet:
Tests pass 🎉

Issue #528, opened 2022-05-27
Union with more than 2 arguments


petvana:
I've just found that `union` with more than 2 arguments fallbacks to the implemetation from Base (Julia v1.7.2), and returns an array. Is it intentional?
``` julia
julia> union(1..2, 3..4)
[1, 4]

julia> union(1..2, 3..4, 5..6)
3-element Vector{Interval{Float64}}:
 [1, 2]
 [3, 4]
 [5, 6]

julia> @which union(1..2, 3..4)
union(a::Interval, b::Interval) in IntervalArithmetic at /home/petr/repos/IntervalArithmetic.jl/src/intervals/set_operations.jl:137

julia> @which union(1..2, 3..4, 5..6)
union(s, sets...) in Base at abstractset.jl:50
```
This also happens for union with number
``` julia
julia> union(1..2, 1)
2-element Vector{Interval{Float64}}:
 [1, 2]
 [1, 1]

julia> union(1, 1..2)
2-element Vector{Interval{Float64}}:
 [1, 1]
 [1, 2]


```

lucaferranti:
probably [this line](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/src/intervals/set_operations.jl#L141)

should be changed to
```julia
union(a::Interval...) = reduce(∪, a)
```

or just remove that and use only `hull` (which is clearer and shorter to type)

```julia
julia> (1..2) ∪ (3..4) ∪ (5..6)
[1, 6]

julia> hull(1..2, 3..4, 5..6)
[1, 6]
```

lucaferranti:
technically the current definition `union(a...) = reduce(union, a)` is type pyracy

petvana:
Using `hull` seems better because it would be complicated to solve mixed cases, like `union(1..2, 3, 4, 5..6)`. However, `hull` seems to have its own issues (Julia 1.7.2, IntervalArithmetic v0.20.6)
``` julia
julia> hull(1, 2..3)
ERROR: StackOverflowError:
Stacktrace:
     [1] mapfoldl(f::Function, op::Function, itr::Tuple{Int64, Interval{Float64}}; init::Base._InitialValue)
       @ Base ./reduce.jl:162
     [2] mapfoldl
       @ ./reduce.jl:162 [inlined]
     [3] #mapreduce#248
       @ ./reduce.jl:289 [inlined]
     [4] mapreduce
[...]

PR #530, opened 2022-05-27
Quick and dirty fix for hull(1, 2..3)

petvana:
As proposed in #528 by @lucaferranti .

Unfortunately, it does not cover all the cases yet, like 
``` julia
julia> hull(1, 2, 3, 1..2)
ERROR: StackOverflowError:
...
julia> hull(1, 1)
ERROR: StackOverflowError:
...
```
So, it's only a draft now.

### DIFF
--- a/src/intervals/set_operations.jl
+++ b/src/intervals/set_operations.jl
@@ -120,6 +120,10 @@ Returns the "interval hull" of the intervals `a` and `b`, considered as
 all of `a` and `b`.
 """
 hull(a::Interval, b::Interval) = Interval(min(a.lo, b.lo), max(a.hi, b.hi))
+
+# Prevent using fallback from Base
+hull(a::Real, b::Interval) = hull(Interval(a), b)
+hull(a::Interval, b::Real) = hull(a, Interval(b))
 #
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
--- a/src/intervals/set_operations.jl
+++ b/src/intervals/set_operations.jl
@@ -128,7 +128,11 @@ hull(a::Interval, b::Real) = hull(a, Interval(b))
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
-hull(a...) = reduce(hull, a)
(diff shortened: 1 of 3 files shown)


lucaferranti:
using promotion should(?) work for master. Promotion might be removed when `1.0-dev` is merged, but that's not something you need to worry in this PR

Kolaru:
> Promotion might be removed when `1.0-dev` is merged, but that's not something you need to worry in this PR

That being said (and for the record, this indeed don't need to apply to this PR), that kind of edge case is one of the reason to remove promotion. Without promotion these function calls should error with a proper `MethodError` instead of falling back to base, avoiding incorrect results.

Changing
```julia
hull(a...) = reduce(hull, a)
```
to
```julia
hull(a...) = reduce(hull, Interval.(a))
```
should fix it for the other cases you mentionned.

petvana:
@Kolaru Thank you for the idea. I've extended it by new non-exported function `tointerval` to make it compatible with `IntervalBox`.

src/multidim/intervalbox.jl
```diff
@@ -45,7 +45,7 @@ numtype(x::IntervalBox{N, T}) where {N, T<:Real} = T
 
 length(X::IntervalBox{N,T}) where {N,T} = N
 
-
+tointerval(x::IntervalBox) = x
```
lucaferranti, line 48:
```suggestion
tointerval(x::IntervalBox) = x
tointerval(x::AbstractVector) = IntervalBox(x)
```

src/intervals/set_operations.jl
```diff
@@ -120,11 +120,19 @@ Returns the "interval hull" of the intervals `a` and `b`, considered as
...
+# Prevent using fallback from Base
+hull(a::Real, b::Interval) = hull(Interval(a), b)
+hull(a::Interval, b::Real) = hull(a, Interval(b))
 #
 # hull{T,S}(a::Interval{T}, b::Interval{S}) = hull(promote(a, b)...)
 hull(a::Complex{<:Interval},b::Complex{<:Interval}) =
     complex(hull(real(a),real(b)),hull(imag(a),imag(b)))
-hull(a...) = reduce(hull, a)
+
+tointerval(x) = Interval(x)
```
lucaferranti, line 132: here we could add a comment that e.g. `0.1` is not rounded outwards.

lucaferranti (approved):
after today discussion this is good.

A couple of minor comments

1) could you check if the suggestion I left things like
```julia
hull([1, 2, 3], IntervalBox(1..2, 3..4, 5..6))
```
work and add 1-2 tests for those mixed cases

PR #531, opened 2022-05-28
Use inf, sup (or bounds) instead of .lo and .hi

lbenet:


lbenet:
There are few instances left; by some reason, using `inf`/`sup` produces errors in the tests. So for the time being, I think this is ok...

lucaferranti (approved):
in general LGTM.

If I understand correctly the missing cases are

1. in trigonometric.jl, `@round(F, half_pi(F).lo, atan(ylo, xlo))` we cannot use `inf` inside the round macro (at least without adding `inf(a, RoundingMode)` but feels too pedantic

2. `bounds(x) = x.lo, sup(x)` the reason for this (which is I guess also the reason for always normalising `-0.0` to `0.0` when constructing the interval instead of directly "correctly" normalising lower and upper bound. Is that having `-0.0` around leads some ambiguities errors in `atan(y, x)` computation (returns `π` instead of `-π` and viceversa).

3. A couple of instances in the examples folder, but given those are probably outdated anyway I guess it doesn't matter

Given that the current thing works and given how idiotic the whole -0.0 vs 0.0 issue is in the first place, I would say this is fine, since it works

Kolaru:
> 1. in trigonometric.jl, `@round(F, half_pi(F).lo, atan(ylo, xlo))` we cannot use `inf` inside the round macro (at least without adding `inf(a, RoundingMode)` but feels too pedantic

`F(inf(half_pi(F)), atan(ylo, xlo, RoundUp))` should do the trick (I don't want to mess with the PR while you are rebasing :))

lbenet:
The last commit takes into account @Kolaru suggestions [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/531#issuecomment-1140272910).

lbenet:
(Forgot to push the last commit...)

lucaferranti:
@lbenet I suggest we merge this and leave the pi vs -pi nuisance for a separate PR

lucaferranti:
also because since #533 is built on top on this and hence it also incorporates all the `inf, sup` changes, which makes review a little harder

lbenet:
I think the last commit includes the nuisances related to pi you mean. 

I agree: let's merge this, and afterwards I'll rebase #533 to 1.0-dev

PR #532, opened 2022-05-28
Use rounding in extended div

lbenet:


lbenet:
There was a small bug in `extended-div` (1,0-dev branch), which was not rounding.

lucaferranti:
Indeed I just checked in `IntervalContractors.jl` and using `1.0-dev` there were a few new test failures. Some for rev power taken care of #529 and some with `mul_rev_to_pair` taken care of by this (that is with this branch those tests pass).

Side note, the only difference between `mul_rev_to_pair` in the standard and `extended_division` here is the order of the inputs, indeed in `IntervalContractors.jl` the definition is

```julia
mul_rev_to_pair(b::Interval, c::Interval) = extended_div(c, b)
```

does that make sense? Or can we just change the input order of `extended_div` here so that it matches the IEEE standard? The ITF1788 tests for `mulRevToPair` are all nicely in their own file, so picking the corresponding tests from `IntervalContractors` is trivial

lucaferranti (approved):

PR #548, opened 2022-10-11
Test Julia v1.6 explicitly in CI

lbenet:


.github/workflows/CI.yml
```diff
@@ -12,6 +12,7 @@ jobs:
       matrix:
         version:
           - '1'
```
schillic, line 14: This already uses the latest version with pattern `1.*`.

lucaferranti: yep: https://github.com/JuliaIntervals/IntervalArithmetic.jl/actions/runs/3228974047/jobs/5285738961#step:3:7

lbenet: Ups.... I thought that `1` stood for the LTS, which is 1.6, isn't it? Should I add 1.6 instead of 1.8? Or simply close this?

lucaferranti: `1` stands for latest `1.x.y` stable  release. I don't have strong opinions on whether or not we should test on LTS. I guess it could be good

lbenet: Currently [we require](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/9813b95655d4319439baba6833028b82ff84e993/Project.toml#L26) Julia 1.5; I think it makes sense to require 1.6 instead, and thus it is *healthy* to test it.... I'll change this PR. (Yet, I leave this adjustment to #388, which I am reviewing now.)

Thanks @schillic for spotting this!

PR #552, opened 2023-01-19
1.0-dev: remove superfluous `Real`-like methods for `Interval`

OlivierHnt:
Julia already defines the fallback method `real(x::Real) = x` so there is no need to add a new method.

Also, minor format changes in the docstrings and the header of the file.

### DIFF
--- a/src/intervals/real_interface.jl
+++ b/src/intervals/real_interface.jl
@@ -1,13 +1,16 @@
-#= This file contains the function that must be defined on Intervals
-    so that they behave like Real in julia.
+# This file is part of the IntervalArithmetic.jl package; MIT licensed
+
+#=
+Description:
+
+This file contains the functions that must be defined on `Interval`s so that
+they behave like `Real` in julia.
 =#
 
-real(a::Interval) = a
-
-zero(a::F) where {T<:Real, F<:Interval{T}} = F(zero(T))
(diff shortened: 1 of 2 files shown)


OlivierHnt:
I also removed:
- `size(::Interval) = (1,)` since it does not match the behaviour in Base, i.e. `size(::Number) = ()`. Note that this has no impact on `length` which is still `1`.
- `eltype(::F) where {F<:Interval} = F` since this is already defined in Base.

Kolaru:
LGTM thanks a lot.

Note : I checked and the removed methods are already checked in the test suite

OlivierHnt:
@Kolaru
Sorry I just noticed I made a mistake in the PR #553. I made a commit (cf. [f9c8bd1](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/553/commits/f9c8bd19b0b47777842339502692468244196fe5)), which explicitly reverts the commit [e04314b](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/552/commits/e04314b7c83f5329f420797f566623c5eccc94b0) from this PR... 🤦‍♂️ 
Can you cherry-pick this commit and push it again into 1.0-dev?
It is a very minor change so I can also quickly re-open a PR if that is easier.

Kolaru:
That's slightly outside my git comfort zone, but I just did it. Nice catch :)

PR #553, opened 2023-01-19
1.0-dev: fix `Interval(::Interval)` ambiguity error

OlivierHnt:
Before this PR
```Julia
julia> Interval(Interval(1))
ERROR: MethodError: Interval(::Interval{Float64}) is ambiguous.

Candidates:
  (::Type{T})(x::T) where T<:Number
    @ Core boot.jl:792
  Interval(a)
    @ IntervalArithmetic ~/.julia/dev/dev_IA/IntervalArithmetic/src/intervals/construction.jl:73

Possible fix, define
  Interval(::Interval)

Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```
With this PR:
```Julia
julia> Interval(Interval(1))
[1, 1]
```

The PR also contains some minor changes:
- simplified the logic in `_normalisezero` as I did not observe any impact on performance. But perhaps I am missing something?
- only keep the required inner constructor (the other one is now a regular outer constructor)

### DIFF
--- a/src/intervals/construction.jl
+++ b/src/intervals/construction.jl
@@ -34,7 +34,7 @@ Interval{Float64}
 """
 default_bound() = Float64
 
-@inline _normalisezero(a::Real) = ifelse(iszero(a) && signbit(a), copysign(a, 1), a)
+@inline _normalisezero(a::Real) = ifelse(iszero(a), zero(a), a)
 
 """
     Interval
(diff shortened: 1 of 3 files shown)


src/intervals/construction.jl
```diff
@@ -64,18 +64,11 @@ struct Interval{T} <: Real
     lo::T
     hi::T
 
-    function Interval{T}(a, b) where T
```
lucaferranti, line 67: I dont thinks this should be removed? I think this is needed to correctly construct an interval from rationals or irrational numbers?

OlivierHnt: Yes, in fact I did not delete it, I simply defined it as an outer constructor (cf. line 71) since it is more of a "feature" and not an "essential" constructor.

Kolaru: I think that's fair. And since the tests don't complain it looks fine to me.

Besides, it clarifies the logic which is nice.

PR #555, opened 2023-01-23
1.0-dev: improve printing and fix alignment issue

### LINKED ISSUE #316: BigFloat output incorrect for large exponents
```jl
julia> IntervalArithmetic.atomic(Interval{BigFloat}, -Inf)
[-∞, -5.87565e+1388]₂₅₆

julia> prevfloat(big(Inf))
5.875653789111587590936911998878442589938516392745498308333779606469323584389875e+1388255822130839282
```
The exponent in the interval display is wrong (truncated).

cc @yashrajgupta

### LINKED ISSUE #434: Fix printing of matrices of intervals
```jl
julia> M = [1.2..1.6 1.2..1.6; -1..1 -1..1]
2×2 Array{Interval{Float64},2}:
      [1.19999, 1.60001]       [1.19999, 1.60001]
 [-1, 1]                  [-1, 1]
```

OlivierHnt:
Over the week-end I experimented with the printing mechanism of the 1.0-dev branch. This is the resulting PR 🙂.

Fix #316, #434.

It seems that the alignment issue when printing arrays is linked to the fact that `Interval <: Real` and sometimes the bounds of the intervals are printed with no decimal points. The current behaviour (before this PR) relies on the format `"%.g"` which gives the shortest form for a given number (e.g. `1.0` would be returned as `1`).

So this would explain also why `IntervalBox` had no printing issues: it is not a subtype of `Real`.

In this PR I tried to stay as close as possible to the current behaviour, so the only **visual change** is that exact integers are displayed with a trailing decimal point and 0, for instance `1` is printed as `1.0`.
I have modified the tests to comply with this change.

Before this PR:

```julia
julia> M = [1.2..1.6 1.2..1.6; -1..1 -1..1]  # format === :standard
2×2 Matrix{Interval{Float64}}:
      [1.19999, 1.60001]       [1.19999, 1.60001]
 [-1, 1]                  [-1, 1]

julia> M = [1.2..1.6 1.2..1.6; -1..1 -1..1]  # format === :midpoint
2×2 Matrix{Interval{Float64}}:
     1.4 ± 0.200001      1.4 ± 0.200001
 0 ± 1               0 ± 1
```

After this PR:

```julia
julia> M = [1.2..1.6 1.2..1.6; -1..1 -1..1]  # format === :standard
2×2 Matrix{Interval{Float64}}:
  [1.19999, 1.60001]   [1.19999, 1.60001] # expected shift to the right to account for potential minus signs `-`
 [-1.0, 1.0]          [-1.0, 1.0]

julia> M = [1.2..1.6 1.2..1.6; -1..1 -1..1]  # format === :midpoint
2×2 Matrix{Interval{Float64}}:
 1.4 ± 0.200001  1.4 ± 0.200001
 0.0 ± 1.0       0.0 ± 1.0
```

### Other changes

- There was only one test which did not pass but I think it is wrong.
Before this PR:
```julia
string(2.1..2.2) == [2.1, 2.20001]   # seems wrong, no?
```
After this PR:
```julia
string(2.1..2.2) == [2.09999, 2.20001]
```

- I uncommented some tests for `Complex{<:Interval}`, they pass successfully. In the file changelog.md, it is mentioned that support for `Complex{<:Interval}` is dropped 🙁. However, there are still existing functionalities for `Complex{<:Interval}` on this branch and the display tests pass without extra work.

- Rename the keyword argument `sigfigs`  of `setformat` to `sigdigits` to match the naming convention of Julia (e.g. see the keywords of `round`).

- Fix all the inconsistencies where `==` was used instead of `===` (or `!=` instead of `!==`). In particular, use `===` to compare symbols.

- Improve performance of printing. For instance, faster `supscriptify` and `subscriptify` and `representation` is now type-stable.

For instance, before this PR:

```julia
julia> @btime IntervalArithmetic.subscriptify($8);
  135.643 ns (6 allocations: 352 bytes)

julia> @btime IntervalArithmetic.subscriptify($111);
  202.847 ns (7 allocations: 464 bytes)

julia> @btime IntervalArithmetic.supscriptify($8);
  148.988 ns (7 allocations: 448 bytes)

julia> @btime IntervalArithmetic.supscriptify($111);
  178.257 ns (7 allocations: 496 bytes)
```

After this PR:

```julia
julia> @btime IntervalArithmetic.supscriptify($8);  # Same timing for subscriptify
  3.666 ns (0 allocations: 0 bytes)

julia> @btime IntervalArithmetic.supscriptify($111);  # Same timing for subscriptify
  107.967 ns (4 allocations: 240 bytes)
```

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -11,7 +11,7 @@ function Base.show(io::IO, ::MIME"text/plain", params::DisplayParameters)
     print(io, "  - significant digits: ", params.sigdigits)
 end
 
-const display_params = DisplayParameters(:standard, false, 6)  # Default
+const display_params = DisplayParameters(:standard, true, 6)  # Default
 
 """
     setformat(format::Symbol; decorations::Bool, sigdigits::Int)
@@ -38,7 +38,7 @@ Possible options:
 
 See also: `@format`.
 
-# Example:
+# Example
 ```
 julia> x = interval(0.1, 0.3)  # Default display options
 [0.0999999, 0.300001]
@@ -46,7 +46,7 @@ julia> x = interval(0.1, 0.3)  # Default display options
 julia> setformat(:full)  # Equivalent to `@format full`
 Display parameters:
   - format: full
-  - decorations: false
+  - decorations: true
   - significant digits: 6
 
 julia> x
@@ -55,7 +55,7 @@ Interval(0.09999999999999999, 0.30000000000000004)
 julia> setformat(:standard; sigdigits = 3)  # Equivalent to `@format standard 3`
 Display parameters:
   - format: standard
-  - decorations: false
+  - decorations: true
   - significant digits: 3
 
 julia> x
@@ -100,7 +100,7 @@ julia> x = interval(0.1, 0.3)  # Default display options
 julia> @format full  # Equivalent to `setformat(:full)``
 Display parameters:
   - format: full
-  - decorations: false
+  - decorations: true
   - significant digits: 6
 
 julia> x
--- a/src/display.jl
+++ b/src/display.jl
@@ -19,7 +19,7 @@ const display_params = DisplayParameters(:standard, false, 6)  # Default
 
 Change the format used by `show` to display intervals.
 
-Initially, the display options are `format = : standard`, `decorations = false`
+Initially, the display options are `format = :standard`, `decorations = false`
 and `sigdigits = 6`.
 
 If any of the three argument `format`, `decorations` and `sigdigits` is omitted,
--- a/src/display.jl
+++ b/src/display.jl
@@ -253,36 +253,27 @@ end
 # Code inspired by `_string(x::BigFloat, k::Integer)` in base/mpfr.jl
 
 function round_string(x::AbstractFloat, sigdigits::Int, r::RoundingMode)
-    precision_sigdigits = ceil(Int, log10(2)*precision(x))
-    k = sigdigits#min(sigdigits, precision_sigdigits)  # Do not exceed the precision
-
     str_digits = replace(replace(string(x), '.' => ""), '-' => "")
-    if (isinteger(x) || ispow2(x)) && k ≥ length(str_digits) # `x` is exactly representable
-        return round_string(big(x), k, RoundNearest)
+    if (isinteger(x) || ispow2(x)) && sigdigits ≥ length(str_digits) # `x` is exactly representable
+        return round_string(big(x), sigdigits, RoundNearest)
     else
-        return round_string(big(x), k, r)
+        return round_string(big(x), sigdigits, r)
     end
 end
 
-function round_string(x::BigFloat, sigdigits::Int, ::RoundingMode{:Nearest})
-    precision_sigdigits = ceil(Int, log10(2)*precision(x))
-    k = sigdigits#min(sigdigits, precision_sigdigits)  # Do not exceed the precision
-
-    return Base.MPFR._string(x, k-1)  # `k-1` digits after the decimal, so `k` significant digits
-end
+round_string(x::BigFloat, sigdigits::Int, ::RoundingMode{:Nearest}) =
+    Base.MPFR._string(x, sigdigits-1)  # `sigdigits-1` digits after the decimal
 
 function round_string(x::BigFloat, sigdigits::Int, r::RoundingMode)
     if !isfinite(x)
         return string(Float64(x))
     else
-        precision_sigdigits = ceil(Int, log10(2)*precision(x))
-        k = sigdigits#min(sigdigits, precision_sigdigits)  # Do not exceed the precision
         str_digits = replace(replace(string(x), '.' => ""), '-' => "")
-        if (isinteger(x) || ispow2(x)) && k ≥ length(str_digits) # `x` is exactly representable
-            return round_string(x, k, RoundNearest)
+        if (isinteger(x) || ispow2(x)) && sigdigits ≥ length(str_digits) # `x` is exactly representable
+            return round_string(x, sigdigits, RoundNearest)
         else
-            # `k` digits after the decimal, so `k+1` significant digits
-            str = Base.MPFR.string_mpfr(x, "%.$(k)Re")
+            # `sigdigits` digits after the decimal
+            str = Base.MPFR.string_mpfr(x, "%.$(sigdigits)Re")
             rounded_str = round_string(str, r)
             return Base.MPFR._prettify_bigfloat(rounded_str)
         end
@@ -331,8 +322,6 @@ function round_string_down(s::String)
     len = length(mantissa)
     idx = findlast(d -> (d !== '0') & (d !== '.'), mantissa)
     if idx == len  # The extra significant digit is not 0
-        # d = parse(Int, mantissa[len]) - 1
-        # return string(view(mantissa, 1:len-1-1*(1)), d, 'e', exponent)
         return string(view(mantissa, 1:len-1), 'e', exponent)  # Truncate
 
     else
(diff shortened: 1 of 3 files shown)


src/display.jl
```diff
@@ -1,379 +1,405 @@
...
+# Printing mechanism for various types of intervals
 
-if VERSION < v"1.1.0-DEV.683"
-    to_mpfr(r) = Base.MPFR.to_mpfr(r)
-else
-    to_mpfr(r) = convert(Base.MPFR.MPFRRoundingMode, r)
-end
-
-## Output
+show(io::IO, a::Union{Interval,Complex{<:Interval},DecoratedInterval,IntervalBox}) =
```
Kolaru, line 142: Should this be general or only define for `MIME"text/plain"` This also overrides `repr` which is [kind of](https://docs.julialang.org/en/v1/base/strings/#Base.repr-Tuple%7BAny%7D) expected to return a string that can be copy pasted back into julia and work.

Then an even more specific method could be defined for the `:compact:` case, when intervals are printed inside an array.

@lucaferranti @lbenet 

(note that this is also not central to the PR, and can be left for another one)

test/display_tests/display.jl
```diff
@@ -81,58 +81,56 @@ let x, b
...
 
-        setformat(decorations=false)
+        setformat(; decorations = false)
         @test string(a) == "2.5 ± 0.5"
 
     end
 
     setprecision(BigFloat, 128)
 
     @testset "BigFloat intervals" begin
```
Kolaru, line 128: Unless I missed it, can you add the example from #316 to be sure it works now (and that it will continue to work).

Kolaru (changes requested):
Very nice job overall, I'm very thankful you decided to delve into this!

Also thanks very much for the informative comments and docstrings.

src/display.jl
```diff
@@ -1,379 +1,405 @@
...
+    i == 0 && return "₀"
+    i == 1 && return "₁"
+    i == 2 && return "₂"
+    i == 3 && return "₃"
+    i == 4 && return "₄"
+    i == 5 && return "₅"
+    i == 6 && return "₆"
+    i == 7 && return "₇"
+    i == 8 && return "₈"
+    return "₉"
```
lucaferranti, line 404: I would also include the

```julia
i == 9 && return "₉"
```

and throw an exception as fallback. Might be a little overcautious, but if somehow you get an invalid integer and the printing silently,identifying the bug might be a little tricky

lucaferranti (dismissed):

src/display.jl
```diff
@@ -1,379 +1,405 @@
 mutable struct DisplayParameters
```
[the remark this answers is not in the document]
OlivierHnt: Indeed, I removed the extra spacing.

src/display.jl
```diff
@@ -1,379 +1,405 @@
...
+  - decorations: true
+  - significant digits: 3
 
-julia> x
-[0.0999, 0.301]
+julia> y
+[0.0999, 0.301]_com
 ```
 """
 macro format(expr...)
```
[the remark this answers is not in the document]
OlivierHnt: I reverted this change. You are right that it may be better to have a specific PR targeting docstrings.

About the proposition to remove `@format`: I also fail to see why `@format` is more convenient than just calling `setformat` directly. I can make a PR suggesting to remove it.

test/display_tests/display.jl
OlivierHnt: I added the requested tests.

src/display.jl
OlivierHnt: That's a good point. Done.

src/display.jl
OlivierHnt: Interesting. I just tried this and I think it is a nice change. As an added benefit, this allows us to remove `showfull` entirely since `show(:Interval)` now has this behaviour.

OlivierHnt:
@Kolaru @lucaferranti
Thank you for the feedback on this PR! 🙂 

I have made the requested changes.

The commit [cbbd9ed](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/555/commits/cbbd9edab4c5b00d4265b6b0824620eb06142e0d) (which introduces tests for large exponents) revealed that the printing of large exponent changes depending on the OS, e.g. Windows. Since I work on macOS, I did not see this issue.
Does anyone know what could be the cause? An issue with how `BigFloat` is handled?

Moreover, the two last commits are worth discussing. The commit [f66abd1](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/555/commits/f66abd11007b4224b81d09656bcab7f134fe8a2e) defines `show` on `::MIME"text/plain"` as suggested by @Kolaru. The default `show` with no `MIME` checks for `:compact` in which case it will behave as `::MIME"text/plain"` (in particular, this ensures that we have nice printing for matrices).
Doing this makes `show(::Interval)` behave exactly like `showfull`. So the commit [d2418e3](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/555/commits/d2418e3e0a6282f58709e2dbfdafa4b5fec42523) removes `showfull`.
N.B.: now `string(::Interval)` returns a string which looks like the output of `show(::Interval)`, which feels like the right thing.

Kolaru:
> The commit [cbbd9ed](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/555/commits/cbbd9edab4c5b00d4265b6b0824620eb06142e0d) (which introduces tests for large exponents) revealed that the printing of large exponent changes depending on the OS, e.g. Windows. Since I work on macOS, I did not see this issue.
> Does anyone know what could be the cause? An issue with how `BigFloat` is handled?

Oh no. This is concerning, I will try to reproduce locally (I've access to linux and windows) and see if can understand where it comes from.

> N.B.: now `string(::Interval)` returns a string which looks like the output of `show(::Interval)`, which feels like the right thing.

I think it feels right as well, but we need to check if the IEEE standard as an opinion about representing intervals as string. Worst case we need an additional specific method to have the output prescribed by the standard.

OlivierHnt:
> Oh no. This is concerning, I will try to reproduce locally (I've access to linux and windows) and see if can understand where it comes from.

Thanks for looking into this!

> I think it feels right as well, but we need to check if the IEEE standard as an opinion about representing intervals as string. Worst case we need an additional specific method to have the output prescribed by the standard.

I just had a look at the standard and, as far as I can tell, they only specify "interval literals" for parsing purposing. So the meaning of the function `string` from Base does not seem to be conflicting (in fact, it is kind of better now because the output of `string` is now valid code to construct the exact interval).
Perhaps @lucaferranti could double-check this since he seems well versed with the standard? 🙂

lucaferranti:
sorry for the very slow follow-up. I am in a very busy couple of weeks, with several paper deadlines approaching.

I'll try to double check the discussion today

Kolaru:
I checked and the test are passing on my CentOS machine. So I'm very confused.

I will run the tests with the PR on my windows machine tonight.

Kolaru:
This is a platform depend behavior (bug ?) of BigFloat. I reported it there https://github.com/JuliaLang/julia/issues/48678.

On Windows:
```julia
julia> big"-5.87565e+1388255822130839282"
-Inf
```

On Ubuntu (WSL):
```julia
julia> big"-5.87565e+1388255822130839282"
-5.875649999999999999999999999999999999999999999999999999999999999999999999999983e+1388255822130839282
```

For now I think it would be good enough to test with a slightly less crazy bound like `1e123456789`.

OlivierHnt:
Thanks @Kolaru for looking into this! The CI now runs successfully with the smaller exponent that you suggested 🎉

src/display.jl
```diff
@@ -1,379 +1,405 @@
...
+    println(io, "Display options:")
+    println(io, "  - format: ", params.format)
+    println(io, "  - decorations: ", params.decorations)
+    print(io, "  - significant digits: ", params.sigdigits)
 end
 
-const display_params = DisplayParameters(:standard, false, 6)
-
-const display_options = (:standard, :full, :midpoint)
+const display_params = DisplayParameters(:standard, false, 6)  # Default
```
lbenet, line 14: What about displaying by default the decoration? If you care about `DecoratedIntervals`, you want to see it!

test/display_tests/display.jl
```diff
@@ -8,235 +8,238 @@ let x, b
...
+        @test sprint(show, MIME("text/plain"), a) == "[2.0, 3.0]₁₂₈"
 
-        setformat(:standard, decorations=true)
-        @test string(a) == "[2, 3]₁₂₈_com"
+        setformat(:standard; decorations = true)
+        @test sprint(show, MIME("text/plain"), a) == "[2.0, 3.0]₁₂₈_com"
 
         setformat(:full)
-        @test string(a) == "DecoratedInterval(Interval(2.0, 3.0), com)"
+        @test sprint(show, MIME("text/plain"), a) == "DecoratedInterval(Interval(2.0, 3.0), com)"
```
lbenet, line 162: This is a comment/suggestion, which I guess is small enough to be incorporated for intervals/decorated intervals with `BigFloat`s: I like that the `:standard` format shows the precision of the BigFloat end-points; this is not included in the `:full` format, as shown by this example. I think displaying the precision is informative; while we *usually* fix it (and forget it), MPFR allows to work with different precisions, which is a case where it is worth a reminder. 
So, what about including the precision for `Interval{BigFloat}` and `DecoratedInterval{BigFloat}` always?

OlivierHnt: It is not a bad idea. Note that we need a signature for `show` which returns parseable Julia code (cf. docstring for `show`). In this PR this behaviour is tied with the formatting mechanism via `setformat(:full)`.
I think the best option would be to add an other symbol, say `:raw` (or perhaps `:parseable` 🤷‍♂️ ), such that `setformat(:raw)` replaces the current`setformat(:full)`. Then, `setformat(:full)` can have the behaviour you suggest.

Also, do you have a display in mind? Something like `@biginterval(1)` is displayed as `"Interval(1.0, 1.0)₂₅₆"`?

src/display.jl
OlivierHnt: That is a good idea, I think that would be a nice improvement 🙂.

test/display_tests/display.jl
OlivierHnt: For consistency, what about doing this for `:midpoint` as well?

So it would look like
```julia
julia> setformat(:midpoint)
Display options:
  - format: midpoint
  - decorations: true
  - significant digits: 6

julia> big(1)..2
(1.5 ± 0.5)₂₅₆

julia> setformat(:full);  # full display of the bounds and show precision if BigFloat

julia> big(1)..2
Interval(1.0, 2.0)₂₅₆

julia> setformat(:parseable);  # parseable Julia code

julia> big(1)..2
Interval(1.0, 2.0)
```

lbenet: Great that the comment was useful! 😄 

I like all proposals:
- the `:raw` and `:full` formats, where the raw format is for copy-paste (or saving results to be reused), and the full format is to have all details displayed;
- I like the display you propose `Interval(1.0, 1.0)₂₅₆`;
- include the precision in the `:midpoint` format.

OlivierHnt: But now I am not unsure about the `:raw`/`:parseable` options 🤔. Mainly because the way Julia gives information of the precision for `BigFloat` is through the function `precision` and not the display.
So showing `Interval{BigFloat}` is sufficient and consistent with Base. A user would know that they have an interval with bounds of type `BigFloat` which has precision `precision(BigFloat)`.

But it is definitely useful to display the precision for `:midpoint`. So I will add a commit for this.

OlivierHnt:
In the last commit I have also changed the display of decorated intervals with the format `:midpoint`. Before `DecoratedInterval(1..2)` was displayed `1.5 ± 0.5_com`, now it is shown as `(1.5 ± 0.5)_com`.

Kolaru (approved):
Everything looks good to me, thanks a lot!

lbenet:
LGTM too, so let's go ahead and merge this!

lbenet:
Thanks a lot @OlivierHnt

src/display.jl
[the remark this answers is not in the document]
OlivierHnt: See PR #563.

PR #556, opened 2023-01-26
Use `AbstractIrrational` instead of `Irrational`

devmotion:
I was looking at usage of Irrational in packages (https://juliahub.com/ui/Search?q=Irrational&type=symbols&u=use&t=type), to estimate how breaking https://github.com/JuliaMath/IrrationalConstants.jl/pull/19 (motivated by https://github.com/JuliaLang/julia/issues/46531) will be. IntervalArithmetic showed up in the list of packages that define methods for `Irrational` but it seems the definitions could be generalized to support `AbstractIrrational`.

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -125,12 +125,12 @@ const Region{T} = Union{Interval{T}, IntervalBox{T}}
 
 # These definitions has been put there because generated functions must be
 # defined after all methods they use.
-@generated function Interval{T}(x::Irrational) where T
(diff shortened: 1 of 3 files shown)


lbenet (approved):
Thanks a lot for the initiative @devmotion ! 

LGTM. From my side, it can be merged. Do you have any comments @lucaferranti?

lucaferranti (approved):
LGTM

lbenet:
Merging. Thanks a lot @devmotion !

PR #559, opened 2023-04-16
1.0-dev: Enforce the bounds of `Interval` to be a `Rational` or  an `AbstractFloat`

OlivierHnt:
In this PR, the type of the bounds for an `Interval` are restricted to `Rational` or `AbstractFloat`.

@lbenet I believe this PR fixes the few remaining items that were discussed in PR #558, i.e.

```Julia
Interval{Irrational{:π}}(π, π) # throws an error
interval(1//1, 2//1) # has type Interval{Rational{Int}} instead of Interval{Float}
```

Or did I miss some?

One note: perhaps it is possible to do better in the construction of an `Interval{<:Rational}` when the inputs have a different type? See lines https://github.com/OlivierHnt/IntervalArithmetic.jl/blob/e381c3fa23895a5d77fe963810837ecc422c1b10/src/intervals/construction.jl#L79-L82

### DIFF
--- a/src/intervals/arithmetic/power.jl
+++ b/src/intervals/arithmetic/power.jl
@@ -120,11 +120,11 @@ end
 
 function ^(a::Interval{Rational{T}}, x::AbstractFloat) where {T<:Integer}
     a = Interval{Float64}(inf(a).num/inf(a).den, sup(a).num/sup(a).den)
-    return F(a^x)
+    return a^x
 end
 
 # Rational power
-function ^(a::F, x::Rational{R}) where {F<:Interval, R<:Integer}
+function ^(a::F, x::Rational{R}) where {F<:Interval,R<:Integer}
     p = x.num
     q = x.den
 
@@ -244,7 +244,7 @@ for f in (:exp2, :exp10, :cbrt)
 end
 
 for f in (:log, :log2, :log10)
-    @eval function ($f)(a::F) where {T, F<:Interval{T}}
+    @eval function ($f)(a::F) where {T<:NumTypes,F<:Interval{T}}
             domain = F(0, Inf)
             a = a ∩ domain
 
@@ -254,7 +254,7 @@ for f in (:log, :log2, :log10)
         end
 end
 
-function log1p(a::F) where {T, F<:Interval{T}}
+function log1p(a::F) where {T<:NumTypes,F<:Interval{T}}
     domain = F(-1, Inf)
     a = a ∩ domain
 
@@ -290,7 +290,7 @@ function nthroot(a::F, n::Integer) where {F<:Interval{BigFloat}}
     return interval(low , high)
 end
 
-function nthroot(a::F, n::Integer) where {T, F<:Interval{T}}
+function nthroot(a::F, n::Integer) where {T<:NumTypes,F<:Interval{T}}
     n == 1 && return a
     n == 2 && return sqrt(a)
 
--- a/src/intervals/complex.jl
+++ b/src/intervals/complex.jl
@@ -1,11 +1,10 @@
-
 for op in (:⊆, :⊂, :⪽)
-    @eval function $(op)(x::Complex{Interval{T}}, y::Complex{Interval{S}}) where {T, S}
+    @eval function $(op)(x::Complex{<:Interval}, y::Complex{<:Interval})
         return $(op)(real(x), real(y)) && $(op)(imag(x), imag(y))
     end
 end
 
-function ^(x::Complex{Interval{T}}, n::Integer) where {T}
+function ^(x::Complex{<:Interval}, n::Integer)
     if n < 0
         return inv(x)^n
     end
@@ -13,11 +12,11 @@ function ^(x::Complex{Interval{T}}, n::Integer) where {T}
     return Base.power_by_squaring(x, n)
 end
 
-function ^(x::Complex{Interval{T}}, y::Real) where {T}
+function ^(x::Complex{<:Interval}, y::Real)
     return exp(y*log(x))
 end
 
-function ^(x::Complex{Interval{T}}, y::Complex) where {T}
+function ^(x::Complex{<:Interval}, y::Complex)
     return exp(y*log(x))
 end
 
@@ -57,7 +56,7 @@ function sqrt_realpart(x::T, y::T,RND::RoundingMode) where T<:AbstractFloat
     ρ = ldexp(sqrt(ρ,RND),k) #sqrt((abs(z)+abs(x))/2) without over/underflow
 end
 
-function sqrt(z::Complex{Interval{T}}) where T<:AbstractFloat
+function sqrt(z::Complex{Interval{T}}) where {T<:AbstractFloat}
     x, y = reim(z)
 
     (inf(x) < 0 && inf(y) < 0 && sup(y) >= 0) && error("Interval lies across branch cut")
@@ -101,7 +100,7 @@ function sqrt(z::Complex{Interval{T}}) where T<:AbstractFloat
 end
 
 
-function log(z::Complex{T}) where T<:Interval
+function log(z::Complex{<:Interval})
     ρ = abs(z)
     θ = angle(z)
 
@@ -109,13 +108,10 @@ function log(z::Complex{T}) where T<:Interval
 end
 
 
-function abs2(z::Complex{T}) where T<:Interval
-    return real(z)^2 + imag(z)^2
-end
+abs2(z::Complex{<:Interval}) = real(z)^2 + imag(z)^2
 
-function abs(z::Complex{T}) where T<:Interval
-    return sqrt(abs2(z))
-end
+abs(z::Complex{<:Interval}) = sqrt(abs2(z))
+
 #
 # # \left( |x|^p \right)^{1/p}.
 # function norm(z::Complex{T}, p=2) where T<:Interval
@@ -123,6 +119,6 @@ end
 # end
 
 # real functions
-mid(z::Complex{T}) where {T <: Interval} = mid(real(z)) + mid(imag(z)) * im
-mag(z::Complex{T}) where {T <: Interval} = sup(abs(z))
-mig(z::Complex{T}) where {T <: Interval} = inf(abs(z))
+mid(z::Complex{<:Interval}) = complex(mid(real(z)), mid(imag(z)))
+mag(z::Complex{<:Interval}) = sup(abs(z))
+mig(z::Complex{<:Interval}) = inf(abs(z))
--- a/src/intervals/construction.jl
+++ b/src/intervals/construction.jl
@@ -34,11 +34,13 @@ Interval{Float64}
 """
 default_bound() = Float64
 
+const NumTypes = Union{Rational,AbstractFloat} # allowed types for the bounds of an interval
+
 # Produce the type of the bounds of an interval when not explicitly imposed
-promote_numtype(::Type{T}, ::Type{S}) where {T<:AbstractFloat, S<:AbstractFloat} = promote_type(T, S)
-promote_numtype(::Type{T}, ::Type{S}) where {T<:AbstractFloat, S} = promote_type(T, S)
-promote_numtype(::Type{T}, ::Type{S}) where {T, S<:AbstractFloat} = promote_type(T, S)
-promote_numtype(::Type{T}, ::Type{S}) where {T, S} = promote_type(default_bound(), T, S)
+promote_numtype(::Type{T}, ::Type{S}) where {T<:NumTypes,S<:NumTypes} = promote_type(T, S)
+promote_numtype(::Type{T}, ::Type{S}) where {T<:NumTypes,S} = promote_type(T, S)
+promote_numtype(::Type{T}, ::Type{S}) where {T,S<:NumTypes} = promote_type(T, S)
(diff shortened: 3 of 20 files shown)


OlivierHnt:
Ah there is also one thing about string literals:
```Julia
I"1//2"
``` 
returns an empty interval. But presumably we would want to return a `Interval{Rational{Int}}`, no?
From glancing at the code, it looks like it could be resolved by modifying `parse_num`.

src/intervals/arithmetic/power.jl
```diff
@@ -120,11 +120,11 @@ end
 
 function ^(a::Interval{Rational{T}}, x::AbstractFloat) where {T<:Integer}
     a = Interval{Float64}(inf(a).num/inf(a).den, sup(a).num/sup(a).den)
-    return F(a^x)
+    return a^x
```
lbenet, line 123: Suggestion on line 122: imposing `a` to be `Interval{Float64}` seems limiting (to me) here. For the construction of `a`, I think we should rather use the type obtained by promoting `T` and `IntervalArithmetic.default_bound()`, or the actual type of `x`. Another option, I learned in your code below, is to use `float(Rational{T})`.

src/intervals/complex.jl
```diff
@@ -57,7 +56,7 @@ function sqrt_realpart(x::T, y::T,RND::RoundingMode) where T<:AbstractFloat
     ρ = ldexp(sqrt(ρ,RND),k) #sqrt((abs(z)+abs(x))/2) without over/underflow
 end
 
-function sqrt(z::Complex{Interval{T}}) where T<:AbstractFloat
+function sqrt(z::Complex{Interval{T}}) where {T<:AbstractFloat}
```
lbenet, line 59: Is there any reason to *not* define the method here as `z::Complex{Interval}`, similarly to other functions ?

src/intervals/construction.jl
```diff
@@ -66,20 +68,26 @@ constructors:
...
+    Interval{T}(a::T, b::T) where {T<:AbstractFloat} = new{T}(_normalisezero(a), _normalisezero(b))
+end
+
+function Interval{T}(a, b) where {S<:Integer,T<:Rational{S}}
+    R = float(S)
+    return Interval{T}(rationalize(S, prevfloat(R(a, RoundDown))), rationalize(S, nextfloat(R(b, RoundUp))))
 end
 
-Interval{T}(a, b) where {T} = Interval{T}(T(a, RoundDown), T(b, RoundUp))
+Interval{T}(a, b) where {T<:AbstractFloat} = Interval{T}(T(a, RoundDown), T(b, RoundUp))
```
lbenet, line 84: Just a note that rounding, here, is not working:
```julia
julia> setformat(:full); 

julia> Interval{Float64}(0.1, 0.1)
Interval(0.1, 0.1)

julia> Interval{BigFloat}(big(0.1), big(0.1))
Interval(0.1000000000000000055511151231257827021181583404541015625, 0.1000000000000000055511151231257827021181583404541015625)
```

src/intervals/construction.jl
```diff
@@ -200,10 +208,11 @@ to contain the number that was typed in.
...
 Otherwise it is the same as using the `Interval` constructor directly.
 """
-atomic(::Type{Interval{T}}, x) where {T} = interval(T, x)
-atomic(::Type{Interval{T}}, x::AbstractString) where {T} = parse(Interval{T}, x)
+atomic(::Type{Interval{T}}, x) where {T<:NumTypes} = interval(T, x)
+atomic(::Type{Interval{T}}, x::AbstractString) where {T<:NumTypes} = parse(Interval{T}, x)
 
-function atomic(::Type{Interval{T}}, x::AbstractFloat) where {T}
+atomic(::Type{Interval{T}}, x::AbstractFloat) where {T<:Rational} = Interval{T}(lo, hi)
+function atomic(::Type{Interval{T}}, x::AbstractFloat) where {T<:AbstractFloat}
```
lbenet, line 215: The following is not the tightest result, though it's a *valid* result
```julia
julia> IntervalArithmetic.atomic(Interval{Float64}, 0.1)
Interval(0.09999999999999999, 0.10000000000000002)
```
I think this should be documented.

The tightest result is obtained with `IntervalArithmetic.atomic(Interval{Float64}, "0.1")`

src/intervals/interval_operations/constants.jl
```diff
@@ -16,15 +16,12 @@ argument given, the default interval bound type is used.
...
 Implement the `empty` function of the IEEE Std 1788-2015 (section 10.5.2).
 """
-emptyinterval(::Type{F}) where {F<:Interval} = F(Inf, -Inf)
-emptyinterval(::F) where {F<:Interval} = emptyinterval(F)
-emptyinterval(::Type{Complex{F}}) where {F<:Interval} = complex(emptyinterval(F), emptyinterval(F))
-
-# NOTE Here the restriction should be any allowed bound type. Using
-# AbstractFloat for simplicity for now
-emptyinterval(::Type{T}) where {T<:AbstractFloat} = emptyinterval(Interval{T})
+emptyinterval(::Type{F}) where {T<:NumTypes,F<:Interval{T}} = F(typemax(T), typemin(T))
```
lbenet, line 19: Note that
```julia
julia> emptyinterval(Interval{Rational{Int}})
ERROR: MethodError: no method matching Rational{Int64}(::Float64, ::RoundingMode{:Down})
...
```

I think we need to define a specialized method for rational intervals, perhaps
```julia
emptyinterval(::Type{F}) where {T<:Integer,F<:Interval{Rational{T}}} = F(typemax(T)//one(T), typemin(T)//one(T))
```

src/intervals/interval_operations/constants.jl
```diff
@@ -41,7 +38,9 @@ of `Interval` for more information about the default interval falvor.
 
 Implement the `entire` function of the IEEE Std 1788-2015 (section 10.5.2).
 """
-entireinterval(::Type{F}) where {T, F<:Interval{T}} = F(-Inf, Inf)
-entireinterval(::F) where {F<:Interval} = entireinterval(F)
-entireinterval(::Type{T}) where T = Interval{T}(-Inf, Inf)
+entireinterval(::Type{F}) where {T<:NumTypes,F<:Interval{T}} = F(typemin(T), typemax(T))
+entireinterval(::Type{T}) where {T<:NumTypes} = entireinterval(Interval{T})
```
lbenet, line 42: i think these methods need to be more constrained, separating rational intervals from those with floats.

lbenet (changes requested):
Thanks a lot @OlivierHnt for this PR. 

I've left a couple of comments which I should be easy to address, together with the case mentioned in [your comment](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/559#issuecomment-1510458748). Once that is complete, I am in favor of merging.

src/intervals/interval_operations/constants.jl
OlivierHnt: That is strange, I cannot reproduce this error (on Julia 1.9-rc2).

src/intervals/interval_operations/constants.jl
OlivierHnt: How come? Is this related to https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/559#discussion_r1168082867?
On my machine, I have
```julia
julia> entireinterval(Interval{Rational{Int}})
[-1//0, 1//0]
```

src/intervals/complex.jl
OlivierHnt: The only reason I saw was type the type parameter `T` was used in the body of this function. Of course we could use something in the form of `T = numtype(z)` instead.

src/intervals/construction.jl
OlivierHnt: Yes `interval` has become the only correctly rounded interval constructor. In my mind, any call to `Interval` is unsafe.

src/intervals/arithmetic/power.jl
OlivierHnt: That is a good point. I fixed `^`.

OlivierHnt:
@lbenet thank you for the review. In the last commit I implemented a quick fix so that `I"1//2"` no longer errors, however the output is a `Interval{Float64}`.
From what I understand of the code, I think that changing `I"1//2"` to return an `Interval{Rational{Int}}` warrants a new PR. There are quite a few changes I would like to make on the constructors macro (e.g. `@interval`, etc.).

src/intervals/complex.jl
lbenet: Fair enough... let's leave it as it is.

src/intervals/interval_operations/constants.jl
lbenet: You are right... I think I was checking a different branch... sorry.

src/intervals/construction.jl
lbenet: i think the same applies to `interval`:
```julia
julia> setformat(:full); 

julia> Interval{Float64}(0.1, 0.1)
Interval{Float64}(0.1, 0.1)

julia> interval(0.1, 0.1)
Interval{Float64}(0.1, 0.1)
```

lbenet: Anyway, this should be left for another PR....

src/parsing.jl
```diff
@@ -234,7 +234,7 @@ Same as `parse(T, s, rounding_mode)`, but also accept string representing ration
 """
 function parse_num(T, s, rounding_mode)
     if '/' in s
-        num, denum = parse.(BigInt, split(s, '/'))
+        num, denum = parse.(BigInt, split(s, '/'; keepempty = false))
         return T(num//denum, rounding_mode)
     end
     return T(parse(BigFloat, s), rounding_mode)
```
lbenet, line 240: As you mentioned, this is probably worth a new PR. Yet, here there are some comments: Since `T` is a type, shouldn't we explicit that? I mean, is it referring to an Interval, or a numberic (<: AbstractFloat) type? I think the case of rational numbers / rational intervals should be handled separately...

lbenet (approved):
Thanks again @OlivierHnt. It LGTM, so I'm in favor of merging. Still, I've left a couple of comments related to parsing and rationals...

src/intervals/construction.jl
OlivierHnt: Ah indeed. But is that something we want to change? I mean `0.1` is a float in its own right and calling `interval` creates a valid interval that encapsulates that float after all.

src/parsing.jl
OlivierHnt: As far as I can tell, in the current implementation, `T` will always be an `AbstractFloat` even for `I"1//2"`.
Eventually, yes we will need to have a different method for `Rational` bounds.

src/intervals/construction.jl
lbenet: I agree that `0.1` is a `Float64`, indeed, but it includes already rounding , and it is actually rounded up (the simplest way to see this is `big(0.1)`). Then, the interval created does not contain the real `0.1 == 1//10`; the interval is indeed tight, but it's definitely not valid. 
This is not the motivation for this PR, yet it is an important concern.
cc @Kolaru

OlivierHnt: Absolutely. Honestly, I do not know how to disambiguate the situation (besides the use of strings and macros that we already have). When `interval` is being called, Julia only has access to the (already parsed) float without knowing that the user actually asked for some nearby value (e.g. the float `0.1` vs the actual value 1/10); in this regard, I consider the current behaviour valid.
Unless perhaps it is possible to monitor Julia's parsing system?

OlivierHnt: I see. In fact why was it chosen that `atomic` does not return the tightest interval?
I may be overlooking something here, but replacing the lines https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/4d6bd155a2d92cc442113c3feba36a7b7b17844b/src/intervals/construction.jl#L209-L214 with
```Julia
if x == lo == hi
    lo = prevfloat(lo)
end
```
would give us the tightest valid interval, no?

lbenet: Not in general... It works for `0.1`, because the parsed `0.1` is rounded upwards. If you use `0.3`, which is rounded downwards, it wouldn't work.... In other words, the Float64 "0.1" represents a real slightly larger than 1//10, and the Float64 `0.3` is a real slightly smaller than 3//10.
The problem is how to know the actual rounding mode used when parsing, to compensate it...

OlivierHnt: Ah indeed, thanks for the clarification 🙂

OlivierHnt:
Are there any items left for us to discuss about this PR?

lbenet:
I am in favor this PR as it is; do you agree @Kolaru ?

src/intervals/construction.jl
Kolaru: Even with a macro, we can not intercept the input before parsing to Float64. So this issue can only be solved in few ways
1. Just accept that the Float64 we get is the one intended (current state)
2. Use strings and parse them ourselves
3. Widen all floats as `(prefloat(x)..nexfloat(x))` (as Luis explained we can not do better)
4. Try to guess what the user intended (that use to be the case, converting floats to the closest reasonnable rational number)

At some point it was decided to remove the 4th possibility, and let the user decide what they want. The documentation still needs some big warning about that somewhere though.

Kolaru (approved):
That's very nice, thanks a lot !

Issue #109, opened 2018-04-02
Add `rand`/`randn`?


tkoolen:
Would it make sense to have

```julia
Compat.Random.rand(rng::AbstractRNG, ::Type{Interval{T}}) where {T} = Interval(rand(rng, T))
Compat.Random.randn(rng::AbstractRNG, ::Type{Interval{T}}) where {T} = Interval(randn(rng, T))
```

?

dpsanders:
There is also an argument that is should instead return an interval with two different random endpoints.

It certainly makes sense to have a function to generate some kind of random interval. Since it's not so clear (to me) what a "canonical" random interval would look like, perhaps it does not make so much sense to extend `rand`?

tkoolen:
Yeah, that makes sense; I suppose it's not clear which one is "right". Probably better to require the user to be explicit about which one is desired (and thus not define methods for `rand`/`randn`).

PR #561, opened 2023-05-03
Add rand(Interval)

schillic:
This is a proposal of something I find useful. There was a related discussion in #109.

### DIFF
--- a/src/rand.jl
+++ b/src/rand.jl
@@ -1,4 +1,5 @@
 using Random
+using Random: GLOBAL_RNG
 
 Base.rand(X::Interval{T}) where {T} = X.lo + rand(T) * (X.hi - X.lo)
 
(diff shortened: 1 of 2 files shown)


src/rand.jl
```diff
@@ -8,6 +9,17 @@ Base.rand(X::IntervalBox) = rand.(X)
 Random.gentype(::Type{Interval{T}}) where {T} = T
 Random.rand(rng::AbstractRNG, X::Random.SamplerTrivial{Interval{T}}) where {T} = rand(X[])
 
+Base.rand(type::Type{<:Interval}) = rand(GLOBAL_RNG, type)
+Base.rand(::Type{Interval}) = rand(Interval{Float64})
+
+function Random.rand(rng::AbstractRNG, ::Type{Interval{T}}) where {T}
```
lucaferranti, line 15: should this be `Random.randn` if it uses randn internally?

schillic: I guess so, yes.

schillic: I changed the name.

schillic:
Any opinion on this PR?

Kolaru:
I personally don't really like it. As far as I know, there is no accepted definition of what a random interval should be, let alone a normally distributed one.

My expectation would be that anyone needing a random interval would need something slightly different. Since it is easy enough (`interval(sort(randn(2))...)` in the present case), and unless I am missing something, I am rather in favor of not merging it.

PR #565, opened 2023-06-14
add AD hook for abs() per DiffRules

### LINKED ISSUE #564: Derivative of abs(::Interval)
```julia
using IntervalArithmetic, ForwardDiff

ForwardDiff.derivative(abs, -2.0 .. 2.0)
```

```julia
Stacktrace:
 [1] _abs_deriv(x::Interval{Float64})
   @ DiffRules ~/.julia/packages/DiffRules/wKSai/src/rules.jl:73
 [2] abs
   @ ~/.julia/packages/ForwardDiff/vXysl/src/dual.jl:240 [inlined]
```

See https://github.com/JuliaDiff/DiffRules.jl/pull/98 for additional details and problems with implementing an upstream solution.

agerlach:
Closes #564


lbenet:
Looks ok, though i think we should also include the case for `DecoratedIntervals`, which at the moment is not too clear to me...

agerlach:
I have no experience with `DecoratedIntervals` and I'm not even clear what they actually are from the docs. I would need some guidance on that.

lbenet:
> I have no experience with `DecoratedIntervals` and I'm not even clear what they actually are from the docs. I would need some guidance on that.

I have also essentially no experience with decorations... Decorations are a label which tracks the kind of operations involved in the computation. So, e.g., you can know if during a computation you evaluated something like `sqrt(-1 .. 1)`.

Going back to `_abs_deriv`, I guess that if the interval has a definite sign including perhaps zero, the decoration would be unchanged; if the two signs are involved I guess the decoration would be reduced...

But maybe we should leave this for another PR.

lbenet:
One more suggestion/question: Could we have (either through `Requires` or by exploiting the new pkg extensions (weakdeps) for julia 1.9), that `DiffEqRules` is optional?

agerlach:
My personal opinion is that this does not warrant doing an extension. `DiffRules` is a light weight dependency. If you want to do an extension and support <1.9 you may need to add an additional deps to be backwards compatible with Requires.jl anyway.

The ultimate question is how far to take the extension capabilities when lightweight "Core" deps exist. Should things like PlotRecipes and StaticArrayCore be ext? It seems like a lot of developmental and organizational overhead for what you get in return.

agerlach:
Any thoughts on keeping the dep and merging vs doing an extension? I would really like to see this fix merged. As mentioned above, given the lightweight nature of `DiffRules` I don't think an extension is warranted.

src/ad.jl
```diff
@@ -0,0 +1,9 @@
+function _abs_deriv(x::Interval{T}) where T
```
Kolaru, line 1: I believe to support decorated interval, we just need the definition

```julia
_abs_deriv(x::DecoratedInterval) = DecoratedInterval(_abs_deriv(interval(x)), decoration(x))
```
since this operation can not change the decoration (i.e. the validity) of the operation.

test/interval_tests/ad.jl
```diff
@@ -0,0 +1,14 @@
+using IntervalArithmetic, ForwardDiff
+using Test
+
+@testset "AD" begin
+    @test ForwardDiff.derivative(abs, -2.0 .. 2.0) == -1.0 .. 1.0
```
Kolaru, line 5:
```suggestion
using Test

@testset "AD" begin
    for F in (Interval, DecoratedInterval)
        @test ForwardDiff.derivative(abs, F(-2.0 .. 2.0)) == F(-1.0 .. 1.0)
```
This adaptation to all tests  would allow to test both interval and decorated intervals in one go.

Kolaru (changes requested):
I am fine with having ChainRulesCore as a dependency. I think we are not yet at the point of trying to reduce TTFX as much as possible with this package.

I have few suggestions for the rest, mostly to support DecoratedInterval.

agerlach:
It is standard in AD to take reasonable sub-derivative in these cases. For abs,  0 is the most reasonable. See https://juliadiff.org/ChainRulesCore.jl/dev/maths/nondiff_points.html. This is what occurs in this PR. 

I don't really understand the role of `DecoratedInterval`, so I'm unclear how that changes anything.

agerlach:
@OlivierHnt For additionally context, I had this linked in #564, https://github.com/JuliaDiff/DiffRules.jl/pull/98

agerlach:
Also, the tests are now failing after merging from master. I'm not quite sure why though.

OlivierHnt:
I understand that it is convenient to set the derivative at zero to be zero, but I wonder about the mathematical soundness of this (we want to be able to prove mathematical theorems with IntervalArithmetic.jl).

The point of a decoration is to have a way of tracking what went on during the successive operations that resulted in the final output.
In our case here, the derivative of `abs` is undefined at zero. We can return a bounded, non-empty interval if that is meaningful in AD as you suggest; at the same time, the decoration allows us to indicate that something spooky (mathematically) went on, typically we would tag the output interval with "trv" (as in "trivial") to let someone interested in computer-assisted proofs that they should double (or triple) check everything.

We cannot do this for a bare interval (which is currently represented by the structure `Interval`). In this case it probably makes more sense to return an empty interval.

agerlach:
I see. It seems that the ultimate question is whether or not `derivative` should return the strict derivative, or a subderivative. In the later, I believe this meets the definition. https://en.wikipedia.org/wiki/Subderivative#Definition.

Either way, it seems reasonable that a user may want either as there are certainly many applications for IA outside of the context of theorem proving.

OlivierHnt:
Superseded by PR #593.

PR #569, opened 2023-07-22
fix typos

spaette:
NEWS.md
docs/src/input_output.md
examples/Range of 2-dimensional functions.ipynb
src/intervals/arithmetic.jl
src/intervals/functions.jl
src/intervals/trigonometric.jl
src/multidim/intervalbox.jl

English doesn't have a plural form for `information`.


Kolaru:
Thanks, this is very nice :)

PR #572, opened 2023-08-01
Renaming of functions

### LINKED ISSUE #550: Remove IntervalBox in favor of AbstractVector{Interval}
We discuss we Luca and it looks like we could get rid of `IntervalBox` alltogether, and just define its methods on `AbstractVector{Interval}`.

It would unify things, as a mutlidimensional function extended to intervals would always return a `Vector{Interval}`. This is currently a slight annoyance in IntervalRootFinding.

It doesn't feel like it is type piracy (no clash is possible with other modules), and it doesn't seem to cause any problem (we only have few methods on `IntervalBox` and many are to emulate a vector).

@lucaferranti @dpsanders @lbenet

OlivierHnt:
`IntervalBox` is a wrapper that allows defining (unambiguously) set operations (`union`, `intersect`, `isempty`, etc.) in the sense of interval arithmetic.

This PR removes the `IntervalBox` struct, in favour of just using `AbstractVector{<:Interval}`.
Note that set operations (`union`, `intersect`, `isempty`, etc.) are now confusing since they have a completely different meaning for `Interval` and `AbstractVector`; thus this PR also renames such operations following the standard.

~~Name changes:~~
- ~~`in` -> `ismember`~~
- ~~`isthin` -> `issingleton` (aesthetic change to match the standard)~~
- ~~`isdisjoint` -> `disjoint`~~
- ~~`union` completely removed, it was an alias of `hull`~~
- ~~`hull` -> `convexhull` (aesthetic change to match the standard)~~
- ~~`intersect` -> `intersection`~~
- ~~`issubset` -> `subset`~~
- ~~`isstrictsubset` -> `strictsubset`~~
- ~~`isweaklyless` -> `less` (aesthetic change to match the standard)~~
- ~~`setdiff` -> `setdiffinterval`~~
- ~~`isempty` -> `isemptyinterval`~~
- ~~`isentire` -> `isentireinterval` (aesthetic change for consistency)~~
- ~~`≛` (`\starequal<tab>`) -> `≐` (`\dotequal<tab>`, I changed the star for a dot, since in the standard I saw symbols with dots..) and is now in the `Symbols` module, defined to be the unicode alias of a new function `equal`~~

~~Of course `disjoint`, `intersection`, `subset`, `less` and `equal` are semantically very close to the `Base` functions `is disjoint`, `intersect`, `issubset`, `isless` and `isequal` respectively. Nevertheless, I think if we document this properly (using many admonitions 🙂) this should be find.~~

EDIT: see https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/572#issuecomment-1706426553 for the current name changes.

I did not define all these functions for `AbstractVector`. Only `bisect`, `mince` and `setdiffinterval` are defined since the others can be defined in one line, e.g. `intersection(x::AbstractVector) = intersection.(x)`. Yet, should we define such functions? One benefit could be that it allows for writing code that works seamlessly for an `Interval` or a multidimensional interval.

cc-ed @Kolaru

P.S.: unrelated to the rest, but while I was renaming things, I also changed `midpoint_radius` to `midradius` for consistency, since `midpoint` is the `mid` function.


OlivierHnt:
Two comments:

1. it turns out that `InteractiveUtils` exports the name `less` as well. So we need to rename at least that one function.

2. PR #571 removed comparisons (i.e. "pointwise boolean functions" such as `<`, etc.) because these operations are notorious for silent failures. For instance, someone may not expect `0..2 < 1..3` to be true. In particular, the operation `==` now returns an error message inviting the user to use `≛` (renamed `equal`, or `≐` in the present PR).
Yet, `==(::Interval, Interval)` is not ambiguous in any way: the intervals are equal if they are identical.
In fact, there is still `===`, from `Core`, that is defined (and must be kept). In this regard, it is a bit inconsistent to prevent `==` from working.
Was there any reasons, other than the one mentioned above, for introducing `≛`? If not, I think it makes sense to keep `==` to check equality between intervals.

OlivierHnt:
Yep ok that's fair; let's have as many alarm bells as possible!

OlivierHnt:
@Kolaru have you any feedback on this PR?

Kolaru:
The renaming need to be discussed separately from the removal of `IntervalBox`, they are orthogonal issues. During the last JuliaInterval call, it appears the name changes were very controversial. I don't quite see the need to avoid extending base methods on interval. An interval is both an arithmetic object[*] and a set, so implementing both methods seem fine.

I therefore suggest to open another PR with only the `IntervalBox` removed, and use this one to discuss the renaming.

[*] A concept I just invented to avoid saying it is a number ^^'

OlivierHnt:
Following the triage call, here are the only function with name changes:
- `in` -> `ismember`
- `isdisjoint` -> `isdisjointinterval`
- `union` completely removed
- `intersect` -> `intersectinterval`
- `issubset` -> `isweaklysubset`
- `setdiff` -> `setdiffinterval`
- `isempty` -> `isemptyinterval`
- `isentire` -> `isentireinterval`
- `≛` (`\starequal<tab>`) -> unicode alias of `isequalinterval`

OlivierHnt:
I have added explicit error messages for `==`, `<`, `isdisjoint`, `issubset`, `issetequal`, `in`, `isempty`, `isfinite`, `isnan`, `isinteger`. These are also triggered implicitly when calling `≤`, `≥`, `isequal`, `isless`, `>`, `isinf`.

We can do the same for set operations to disambiguate, namely `intersect(!)`, `union(!)`, `setdiff(!)` (I may be missing some). Yet I wonder how much this will interfere with other parts of Base, which may need these functions to be defined and behave a specific way for a `Real`.

OlivierHnt:
In the last commit I removed `isstrictsubset`/`isstrictsupset` since these are equivalent to `isinterior` (except technically for empty intervals but I think the `isinterior` one is better).
I felt that the new naming were a little too random. Hence for consistency I also did the renaming:
- `isweaksubset` -> `isweakinterior` (`isweaksupset` is removed)
- `interior` ->`istrictinterior`
- `precedes` -> `weakprecedes`

OlivierHnt:
So here are the name changes:
- `in` -> `in_interval`
- `isdisjoint` -> `isdisjoint_interval`
- `intersect` -> `intersect_interval`
- `isempty` -> `isempty_interval`
- `isentire` -> `isentire_interval`
- `≛` -> `isequal_interval` (with `≛` as an alias)
- `issubset` -> `issubset_interval`
- `isinterior` -> `isstrictsubset_interval`
- `precedes` -> `precedes_interval`
- `strictprecedes`  -> `strictprecedes_interval`
- `setdiffinterval` -> `setdiff_interval`
Also, `union` is removed, allowing us to define in the future a `union_interval` function to take the disjoint union of intervals.

Note 1: array operations `intersect`, `setdiff`, etc., do not raise an error; they behave as they do for a `Real`.

Note 2: I did not rename `emptyinterval` to `empty_interval` (nor `entireinterval` to `entire_interval`).

Do not hesitate if you have more suggestions.

cc @Kolaru, @lbenet

PR #574, opened 2023-08-27
Fix multiplication with zero

schillic:
```julia
julia> 0.0 * interval(1)
[1.0, 1.0]  # master
[0.0, 0.0]  # after fix
```


OlivierHnt:
Oops thank you for catching this!

PR #583, opened 2023-09-10
Merge commit fix for v0.21.1

Kolaru:
Some commits are missing from the v0.21.1 branch, that should not be totally breaking.

@OlivierHnt Could you check if this is right ? I cherry picked the fix, as we discussed last week. I think the most serious changes should only be on master and not released yet.


OlivierHnt:
LGTM 👍

PR #586, opened 2023-09-28
Add an extension to support ForwardDiff

timholy:
Because of the extensive changes on `master`, ForwardDiff no longer works with this package. This add an extension (Julia 1.9+) providing at least basic support.

Note that:
- this depends on https://github.com/JuliaDiff/ForwardDiff.jl/pull/667. I am not sure when there will be a new release of ForwardDiff, but the master branch is now `0.11-DEV`.
- you have to `dev ForwardDiff in the `test` folder
- because of https://github.com/JuliaLang/Pkg.jl/issues/3636, you either have to run these tests on `nightly` or we have to abandon `test/Project.toml` and copy everything to the `[extras]` section of the main `Project.toml`.

So obviously this can't be merged yet, but I thought it would be better to submit it now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.8'
+          - '1.9'
           - '1'
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -12,10 +12,17 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 RoundingEmulator = "5eaf0fd0-dfba-4ccb-bf02-d820a40db705"
 SetRounding = "3cc68bcd-71a2-5612-b932-767ffbe40ab0"
 
+[weakdeps]
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+
+[extensions]
+IntervalArithmeticForwardDiffExt = "ForwardDiff"
+
 [compat]
 CRlibm = "0.7, 0.8, 1"
 EnumX = "1"
 FastRounding = "0.2, 0.3"
+ForwardDiff = "0.11"
 RoundingEmulator = "0.2"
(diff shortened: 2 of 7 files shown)


OlivierHnt:
Thanks for opening this PR!
I wonder though if it should be the other way around. ForwardDiff.jl adds IntervalArithmetic.jl as an extension; also this gives a stronger guarantee that their differentiation mechanism is compatible with the specifications of interval arithmetic. 🤔

timholy:
That's actually where I started. But I grew to think it might be better this way: ForwardDiff is widely used by a lot of different things, it seems more scalable to have extensions in packages that may have fewer "interop" demands. As a practical matter, ForwardDiff's tests also take a long time to run!

But in the end I'm not sure it matters a lot; the extension mechanism ensures the code will be loaded, one way or another.

OlivierHnt:
Good point about scalability.

So the main concern here is the reliability. Do we have a guarantee, with your PR, that `iszero_tuple` is used in ForwardDiff.jl in compliance with interval arithmetic specifications?

Unfortunately I am too unfamiliar with ForwardDiff.jl to know the answer. I see that it is used to define `iszero(::Partials)` and `iszero(::Dual)` which are themselves called at several places in the code.

OlivierHnt:
I tried the tests you included in your PR and they seem to pass (except the last one, where the right-hand side is a larger enclosure of the left-hand side) using the PR #593; I will add them to the test suite to keep track of FowardDiff.

OlivierHnt:
@timholy how is the interaction between IntervalArithmetic (v0.22.14) and ForwardDiff for you?
I think the issues raised at the time in this PR have been resolved now. But I am also wondering if you are aware of more issues?

timholy:
I haven't returned to this research topic in a while. I'll get back to it eventually, but for the time being...no idea! 🙂

OlivierHnt:
Closing this for the time being 🙂
Feel free to reopen in the future if something comes up!

Issue #431, opened 2020-12-24
Mince should return an iterator 


dpsanders:
For efficiency

dpsanders:
I think it's just a question of replacing the array comprehension by a generator expression:

```jl
julia> function mince(x::Interval, n)
             nodes = range(x.lo, x.hi, length = n+1)
             return (Interval(nodes[i], nodes[i+1]) for i in 1:length(nodes)-1)
         end
```

lbenet:
I've open #467, which improves performance (broadcasting!), but returns a vector...

OlivierHnt:
Is this still a wanted feature?

edtrelo:
@dpsanders what about using a new structure that relies on ranges? @lbenet and I have been discussing this issue and we think it would be useful to have an implementation of the `getindex` method, that proposed structure would enable us to achive such task.

edtrelo:
`@btime mince($`

PR #587, opened 2023-09-29
Mince as iterable 

edtrelo:
Solves #431: it enables to iterate over a minced interval.


OlivierHnt:
Thanks for opening this PR. I still do not see a need for introducing a new iterator structure just for `mince`; do you think it can have more usage?
Otherwise, since allocating a vector is only really problematic in loops, I think the fix for #431 is to make a `mince!` method, ie an in-place version of `mince`.

OlivierHnt:
Superseded by PR #593. We can re-open this if the in-place version is not good enough.

Issue #408, opened 2020-08-21
Empty interval as [NaN, NaN]?


dpsanders:
Is there anything to stop us using `Interval(NaN, NaN)` as the representation of the empty interval? I think this would remove the need for a lot of `isempty` checks, and hence may improve performance (although this would need to be checked).

PR #590, opened 2023-10-30
Make decorated interval the default

OlivierHnt:
This PR renames `Interval` to `BareInterval`, and `DecoratedInterval` to `Interval`. In effect, decorated intervals are now the default. While I was adapting the tests suite, I changed a few things and did a bit of cleaning up. Hopefully I did not miss anything.

Closes 462; closes 419; closes 394; closes 219.

Note: no warning is emitted when a bad `Interval` is given as an input, this avoid flooding the IO and is arguably not needed with decorations.

Other small changes:
- the `interval` constructor has a new keyword argument `format` which is either ":standard" (default) or ":midpoint". This allows us to put `±` in Symbols as a mere alias of `interval(m, r; format = :midpoint)`
- powers are more conservative with `BareInterval`. Integer powers are still permitted via `nthpow`, otherwise only `^(::BareInterval, ::BareInterval)` is allowed.
- `setformat` is renamed to `setdisplay` for clarity (since now we have the format keyword for interval which is not impacted by this `setformat` method)
- display changed to show open interval if bound is infinite
- parsing macro `@I_str` always return a decorated interval

TODO: some functions are still missing to have complete parity between `BareInterval` and `Interval`, e.g. hyperbolic functions.


OlivierHnt:
I found some interesting things discussed in the PR #527 from @lucaferranti.

According to the Standard, the bare interval part of a NaI (Not an Interval) is the empty interval, i.e. `isempty_interval(bareinterval(interval(2, 1)))` should be true. However, it currently returns `false` for floating-points bounds since both are `NaN`. On the other hand, since there is no `NaN` equivalent for rationals in Julia, `isempty_interval(bareinterval(interval(2//1, 1//1))) == true` as required by the Standard.

I think it may be good to fix this issue in this PR as well.

One approach would be to also address #408. That is, we internally represent an empty interval `x` with **floating-point bounds** as $[NaN, NaN]$ but keep the fact that `inf(x) == Inf` and `sup(x) == -Inf`. One benefit of doing this is that `x.lo`/`x.hi` will return `NaN` which will propagate throughout the computations.
For intervals with **rational bounds**, this means just keeping the current situation since there is no `NaN` equivalent in Julia.

Also from PR #527, I kind of like the idea of displaying a NaI as `NaI`, or `NaI_ill`, or `∅_ill`; similarly to how we currently display the empty interval via `∅`.

OlivierHnt:
Closes #408 and PR #527.

src/intervals/construction.jl
```diff
@@ -259,31 +246,249 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
+"""
+    Decoration
+
+Enumeration constant for the types of interval decorations described in
+Section 11.2 of the IEEE Standard 1788-2015:
+- `com -> 4`: non-empty, continuous and bounded (common)
+- `dac -> 3`: non-empty and continuous (defined and continuous)
+- `def -> 2`: non-empty (defined)
+- `trv -> 1`: always true (trivial)
+- `bad -> 0`: may not be an interval (badly-formed)
```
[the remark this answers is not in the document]
lucaferranti: I am not sure, `def` adn `trv` are still intervals.


The current description of `bad` sounds a lot like the definition of `ill` in the standard, would be good to clarify the difference (and note that `bad` is not a "standard" decoration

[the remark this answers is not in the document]
OlivierHnt: I am resolving this since now we have a separate field `guarantee` instead of the decoration `bad`.

src/intervals/construction.jl
```diff
@@ -88,142 +40,177 @@ default_numtype() = Float64
...
+- `lo :: T`
+- `hi :: T`
 
-julia> interval(Rational{Int32}, 1//1, π)
-Interval{Rational{Int32}}(1//1, 85563208//27235615)
+Constructor compliant with the IEEE Standard 1788-2015: [`bareinterval`](@ref).
 
-julia> interval(1, π)
-Interval{Float64}(1.0, 3.1415926535897936)
+!!! warning
```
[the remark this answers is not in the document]
OlivierHnt: Indeed, I wrote down a docstring for `_unsafe_bareinterval`. Let me know if that's enough.

src/display.jl
```diff
@@ -78,202 +78,255 @@ function setformat(format::Symbol = display_params.format;
...
+    return print(io, _str_repr(a, :full))
 end
 
-# `String` representation of various types of intervals
+# `String` representation
 
-representation(a::Interval, format::Symbol) = basic_representation(a, format)
+function _str_repr(a::BareInterval{T}, format::Symbol) where {T<:NumTypes}
+    # `format` is either `:standard`, `:midpoint` or `:full`
+    var_interval = _str_basic_repr(a, format)
```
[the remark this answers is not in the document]
OlivierHnt: Not sure.
Indeed it is the representation of the bounds, or `∅`, and depending on the format, the string can contain the symbols `[`, `]`, `(` and `)`. Do you have a naming preference? In the meantime I renamed it `str_interval`.

src/intervals/construction.jl
```diff
@@ -88,142 +40,177 @@ default_numtype() = Float64
...
-    macro [`@I_str`](@ref) to ensure tight enclosure around the typed numbers.
-
-See also: [`±`](@ref), [`..`](@ref) and [`@I_str`](@ref).
+"""
+    BareInterval{T<:NumTypes}
 
-# Examples
-```jldoctest
-julia> setformat(:full);
+Interval type for guaranteed computation with interval arithmetic according to
```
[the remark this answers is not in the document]
OlivierHnt: Good idea, done.

OlivierHnt:
In triage, @lucaferranti pointed that the flag for mixing `Interval` and `Number` can not really be a decoration. According to the Standard, the decoration is really there to monitor functions domain, while our decoration `bad` serves a completely different purpose.
For instance, ideally `sqrt(interval(-1, 1) + 0.5)` should be both `bad` and `trv`.

I have added a new field to `Interval`, called `guarantee :: Bool` (as suggested by @lucaferranti, but I have no personal preference on the name) which serves the purpose of the decoration `bad`. This field is always `true`, unless the interval was produced via conversion/promotion.

```Julia
julia> sqrt(interval(-1, 1) + 0.5)
[0.0, 1.22475]_trv

julia> guarantee(sqrt(interval(-1, 1) + 0.5))
false
```

OlivierHnt:
How do we want to display an interval `x` when `guarantee(x) == false` ?

lbenet:
> How do we want to display an interval `x` when `guarantee(x) == false` ? There should be some sort of visual clue for the user.

Perhaps, and only if `guarantee(x) == false`, we could add some sort of second subscript (`_bad` or `_notguaranteed`) to make the point clear, e.g., `[0,1]_trv_notguaranteed` (or some abreviation like `_NG`).

Also, I would suggest to emit a warning or something like that.

OlivierHnt:
After experimenting a bit, I am not sure we should show a warning when mixing `Interval` and `Number`. The problem is that a same function call can induce multiple conversions, each one will produce a warning. This tends to flood the REPL, while not being very informative.

Perhaps the fact that we have added the symbol "NG" (displayed on intervals which are no longer guaranteed) is sufficient?

src/intervals/construction.jl
```diff
@@ -259,31 +253,275 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
+    Interval{T<:NumTypes} <: Real
+
+Interval type for guaranteed computation with interval arithmetic according to
+the IEEE Standard 1788-2015. This structure combines a [`BareInterval`](@ref)
+together with a [`Decoration`](@ref). The decoration records the validity of the
+operations that produced this interval.
+
+Fields:
+- `bareinterval :: BareInterval{T}`
+- `decoration   :: Decoration`
```
Kolaru, line 300: Missing doc for `guarantee` field.

src/intervals/construction.jl
```diff
@@ -259,31 +253,275 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
+bareinterval(::Type{T}, x::Interval) where {T<:NumTypes} = bareinterval(T, bareinterval(x))
+decoration(x::Interval) = x.decoration
+
+"""
+    guarantee(::Interval)
+    guarantee(::BareInterval)
+    guarantee(::Complex{<:Interval})
+
+Return `false` if the `Interval` was constructed using
+`convert(::Type{<:Interval}, ::Number)`, and `true` otherwise.
```
Kolaru, line 340:
```suggestion
Return `false` when the interval **is not guaranteed** to encompass all possible numerical errors.

This happens whenever an `Interval` is constructed using `convert(::Type{<:Interval}, ::Number)` or when a normal number is mixed with an interval.
```

src/intervals/construction.jl
```diff
@@ -259,31 +253,275 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
-function bigequiv(x::Interval{T}) where {T<:NumTypes}
-    lock(precision_lock) do
-        setprecision(precision(float(T))) do
-            return Interval{BigFloat}(x)
-        end
-    end
+struct Interval{T<:NumTypes} <: Real
+    bareinterval :: BareInterval{T}
+    decoration   :: Decoration
+    guarantee    :: Bool
```
Kolaru, line 313: [bikeshedding] I would call that `guaranteed`.

Kolaru (changes requested):
Except from the docstring of the new `Interval` (and few minor things), everything looks fine to me.

docs/src/manual/construction.md
```diff
@@ -14,31 +12,52 @@ interval(Inf) # not valid since infinity is not part of an interval
...
+- `def` (defined): $X$ is a non-empty subset of $\mathrm{Dom}(f)$, i.e. $f$ is defined at each point of $X$.
+- `trv` (trivial): always true; gives no information
+- `ill` (ill-formed): NaI (Not an Interval) (an error occurred), e.g. $\mathrm{Dom}(f) = \emptyset$.
+Decorations follow the ordering `com > dac > def > trv > ill`.
+
+When constructing intervals, if no decoration is explicitly specified, then it is initialised with a decoration according to the underlying bare interval:
+- `com`: if the interval is non-empty and bounded
+- `dac`: if the interval is unbounded
+- `trv`: if the interval is empty
+- `ill`: if the interval is a NaI
```
lbenet, line 29: Isn't `def` missing here? (I think the subtlety wrt `dac` is that continuity is not guaranted)

docs/src/manual/construction.md
```diff
@@ -14,31 +12,52 @@ interval(Inf) # not valid since infinity is not part of an interval
 interval(3, 2) # not valid since the lower bound is strictly greater than the upper bound
 ```
 
-The submodule `IntervalArithmetic.Symbols` exports the infix operator `..` as an alias for `interval`.
+The above intervals came with a decoration. Decorations are flags, or labels, attached to intervals to indicate the status of a given interval. Decorated intervals provide valuable information on the result of evaluating a function on an initial interval.
```
lbenet, line 15: We should mention here the existence of `bareinterval` as non-guaranteed intervals, with a link to the relevant docs.

src/intervals/arithmetic/absmax.jl
```diff
@@ -3,39 +3,48 @@
...
+            isempty_interval(a) && return a
+            isempty_interval(b) && return b
+            return _unsafe_bareinterval(T, $f(inf(a), inf(b)), $f(sup(a), sup(b)))
+        end
+        $f(a::BareInterval, b::BareInterval) = $f(promote(a, b)...)
+
+        function $f(a::Interval, b::Interval)
+            r = $f(bareinterval(a), bareinterval(b))
+            d = min(decoration(a), decoration(b))
+            t = guarantee(a) & guarantee(b)
```
lbenet, line 46: Is there a region to use `&` instead of `&&`?

lbenet: This comment also applies to similar lines elsewhere; I have not added comments on them...

src/intervals/arithmetic/basic.jl
```diff
@@ -2,120 +2,95 @@
...
+-(a::BareInterval, b::BareInterval) = -(promote(a, b)...)
 
--(a::Interval, b::Interval) = -(promote(a, b)...)
-
-function -(a::Interval{T}, b::T) where {T<:NumTypes}
-    isempty_interval(a) && return a
-    return @round(T, inf(a) - b, sup(a) - b)
+function -(a::Interval, b::Interval)
+    r = bareinterval(a) - bareinterval(b)
+    d = min(decoration(a), decoration(b), decoration(r))
```
lbenet, line 54: Is there a decoration associated to `r`? Or is it `com` by default? I think we can spare `decoration(r)` in this line.

src/intervals/arithmetic/absmax.jl
OlivierHnt: That being said, I do not know if it has any impact at all... just a coding habit I guess...

docs/src/manual/construction.md
OlivierHnt: When constructing intervals, I believe we cannot have `def`. This decoration can only be produced by some function.

src/intervals/construction.jl
OlivierHnt: [More bikeshedding] How about `isguaranteed`?

docs/src/manual/decorations.md
```diff
@@ -3,136 +3,3 @@ DocTestSetup = quote
     using IntervalArithmetic
 end
```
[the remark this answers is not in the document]
OlivierHnt: Indeed, done.

docs/src/manual/construction.md
```diff
@@ -62,3 +81,72 @@ The various string formats are as follows:
     ```@repl construction
     I"0.1"
     ```
+
+
+
+## More on decorations
```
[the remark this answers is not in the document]
OlivierHnt: Oops, sorry, done.

src/intervals/construction.jl
OlivierHnt: Indeed, done.

src/intervals/construction.jl
OlivierHnt: Yes that's better, I changed it a little more.

docs/src/manual/construction.md
```diff
@@ -14,31 +12,52 @@ interval(Inf) # not valid since infinity is not part of an interval
...
 
 ```@repl construction
-using IntervalArithmetic.Symbols
-0.1..0.2 # interval(0.1, 0.2)
+x = interval(1, 2, dac)
+interval(x, def)
 ```
 
-The [`±`](@ref) (`\pm<tab>`) infix operator creates the interval from the midpoint and the radius.
+The submodule `IntervalArithmetic.Symbols` exports the infix operator `..` and `±` as an alias for `interval`.
```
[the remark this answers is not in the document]
OlivierHnt: Indeed, done.

OlivierHnt:
Thx for the feedback. I did some changes.

I changed the docs a bit, but they do need much more work in general (thx for opening #591 btw!). I am not opposed to keep adding more docs pertaining to decorations / the new guaranteed field before merging this PR. Let me know if you have further suggestions.

docs/src/manual/construction.md
```diff
@@ -14,6 +14,10 @@ interval(3, 2) # not valid since the lower bound is strictly greater than the up
 
 The above intervals came with a decoration. Decorations are flags, or labels, attached to intervals to indicate the status of a given interval. Decorated intervals provide valuable information on the result of evaluating a function on an initial interval.
 
+Moreover, an `Interval` comes with a boolean `isguaranteed` field which asserts the reliability of the interval. This field allows for `Interval` to interact, in generic code, with other number types by being set to `false` whenever a call to `convert(::{<:Interval}, ::Number)` occurs. A user interested in validated numerics should **always** have a resulting interval for which `isguaranteed` is true.
```
lucaferranti:
```suggestion
Moreover, an `Interval` comes with a boolean `isguaranteed` field which asserts the reliability of the interval. This field allows for `Interval` to interact, in generic code, with other number types by being set to `false` whenever a call to `convert(::Type{<:Interval}, ::Number)` occurs. A user interested in validated numerics should **always** have a resulting interval for which `isguaranteed` is true.
```

docs/src/manual/construction.md
```diff
@@ -14,31 +12,52 @@ interval(Inf) # not valid since infinity is not part of an interval
...
+The above intervals came with a decoration. Decorations are flags, or labels, attached to intervals to indicate the status of a given interval. Decorated intervals provide valuable information on the result of evaluating a function on an initial interval.
+
+Suppose that a decorated interval $(X, d)$ is the result of evaluating a function $f$, or the composition of a sequence of functions, on an initial decorated interval $(X_0, d_0)$. The meaning of the resulting decoration $d$ is as follows:
+- `com` (common): $X$ is a closed, bounded, non-empty subset of the domain of $f$; $f$ is continuous on the interval $X$; and the resulting interval $f(X)$ is bounded.
+- `dac` (defined and continuous): $X$ is a non-empty subset of $\mathrm{Dom}(f)$, and $f$ is continuous on $X$.
+- `def` (defined): $X$ is a non-empty subset of $\mathrm{Dom}(f)$, i.e. $f$ is defined at each point of $X$.
+- `trv` (trivial): always true; gives no information
+- `ill` (ill-formed): NaI (Not an Interval) (an error occurred), e.g. $\mathrm{Dom}(f) = \emptyset$.
+Decorations follow the ordering `com > dac > def > trv > ill`.
```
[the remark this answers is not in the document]
Kolaru: You are right, it is the standard that is confusing, not our doc.

Kolaru (approved):
For me everything looks fine. Thanks a lot!

OlivierHnt:
From my understanding, the flag `isguaranteed` is less about whether the rounding was correct, but more that Julia's promotion system can lead to silent errors.

In other words, we do not know how the `2` in the function `x -> 2^x` was produced, and since it is not an interval we have no choice but considering the operation "not guaranteed".

My attempt at contriving an example:
```Julia
julia> my_two = 2cos(floatmin(Float64)) # yields 1.0 due to rounding errors
2.0

julia> my_two ^ interval(1, 2) # something spooky happened, and we flag the promotion
[2.0, 4.0]_com_NG
```

lbenet:
I agree that your contrived example is good, and as you later noted, mine was emphasizing that integer (literals) are properly rounded always, and therefore the operation is guaranteed. The subtle difference is that, while the result of `2cos(floatmin(Float64))` is 0.0, it still is a `Float64`; perhaps simply allowing "promotion" for `Integer`s could be ok... Let's keep simply in mind.

Kolaru:
I don't think we will be able to prove that integer are safe (they could come from rounding a float).

Regardless, unless we can get a definitive answer I think the current choice is the most reasonable in view of our v1.0 objective. Some operation gaining the guaranteed trait is fine, but some operation losing it would be breaking, we should therefore rather not give it.

PS: Technically both are breaking, but in practice I don't think it makes sense to rely on something not being guaranteed, so it can be seen as an improvement.

docs/src/manual/construction.md
```diff
@@ -1,46 +1,188 @@
...
+These are all examples of ill-formed intervals, resulting in the decoration `ill`.
+
+!!! danger
+    The decoration `ill` is an indicator that an error has occured. Therefore, any interval marked by this decoration cannot be trusted and the code needs to be debugged.
+
+
+
+### Guarantee
+
+A *guarantee* is yet an other label, independent of decorations, and not described by the IEEE Standard 1788-2015 specifications. Its purpose is to accomodate for Julia's extensive conversion and promotion system, while retaining reliability in computations. Specifically, an interval `x` constructed via [`interval`](@ref) satisfies `isguaranteed(x) == true`. However, if a call to `convert(::Type{<:Interval}, ::Real)` occurs, then the resulting interval `x` satisfies `isguaranteed(x) == false`.
```
lbenet, line 136: an other -> another

src/intervals/arithmetic/hyperbolic.jl
```diff
@@ -2,58 +2,93 @@
 # Section 9.1 of the IEEE Standard 1788-2015 and required for set-based flavor
 # in Section 10.5.3
 
-for f in (:sinh, :tanh, :asinh)
+
```
lbenet, line 5: There is a TODO (in the first comment of this PR) which mentions that there are still things to be done to have complete parity between `BareInterval` and `Interval`, "e.g, hyperbolic functions". Are hyperbolic functions now complete? (My guess is that they are indeed ready, but otherwise, we should write somewhere what needs to be done.)

src/intervals/construction.jl
```diff
@@ -259,31 +253,275 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
+
+# some useful extra constructor
+interval(::Type{T}, a::Tuple, d::Decoration; format::Symbol = :standard) where {T<:NumTypes} = interval(T, a..., d; format = format)
+interval(a::Tuple, d::Decoration; format::Symbol = :standard) = interval(a..., d; format = format)
+interval(::Type{T}, a::Tuple; format::Symbol = :standard) where {T<:NumTypes} = interval(T, a...; format = format)
+interval(a::Tuple; format::Symbol = :standard) = interval(a...; format = format)
+
+# standard format
+
+function _interval_standard(::Type{T}, a, b, d::Decoration) where {T<:NumTypes}
```
lbenet, line 404: Docstrings for `_interval_standard` seem missing.

src/intervals/construction.jl
```diff
@@ -259,31 +253,275 @@ function atomic(::Type{T}, a::AbstractFloat) where {T<:AbstractFloat}
...
+_interval_standard(::Type{T}, a::Complex, b::Complex) where {T<:NumTypes} =
+    complex(_interval_standard(T, real(a), real(b)), _interval_standard(T, imag(a), imag(b)))
+_interval_standard(::Type{T}, a::Complex, b) where {T<:NumTypes} =
+    complex(_interval_standard(T, real(a), b), _interval_standard(T, imag(a)))
+_interval_standard(::Type{T}, a, b::Complex) where {T<:NumTypes} =
+    complex(_interval_standard(T, a, real(b)), _interval_standard(T, imag(b)))
+
+# midpoint constructors
+
+function _interval_midpoint(::Type{T}, m, r, d::Decoration) where {T<:NumTypes}
```
lbenet, line 435: Docstrings?

src/intervals/interval_operations/boolean.jl
```diff
@@ -4,200 +4,331 @@
...
+or when the upper bound equals the lower bound or the bounds are consecutive
+floating-point numbers.
 """
-isatomic(x::Interval) = isempty_interval(x) || (inf(x) == sup(x)) || (sup(x) == nextfloat(inf(x)))
+isatomic(x::BareInterval) = isempty_interval(x) | (inf(x) == sup(x)) | (sup(x) == nextfloat(inf(x)))
+
+function isatomic(x::Interval)
+    isnai(x) && return false
+    return isatomic(bareinterval(x))
+end
```
lbenet, line 264: Note the following:

```julia
julia> julia> IntervalArithmetic.atomic(Float64, 0.5)
BareInterval{Float64}(0.49999999999999994, 0.5000000000000001)

julia> isatomic(ans)
false
```

I understand the complications to make `atomic` return thiner intervals in a performant way (for `Float64`s it retuns two-eps wide intervals), and I agree with the result of `isatomic` for the interval used. Yet, it seems awkward that an interval (or bareinterval) created by `atomic` returns `false` by `isatomic`.

There is no need to make anything about this; I simply want to share this.

src/intervals/interval_operations/boolean.jl
```diff
@@ -4,200 +4,331 @@
...
+    isthinzero(x::Interval)
+
+Test whether `x` contains only zero.
+"""
+isthinzero(x::BareInterval) = iszero(inf(x)) & iszero(sup(x))
+
+function isthinzero(x::Interval)
+    isnai(x) && return false
+    return isthinzero(bareinterval(x))
+end
```
lbenet, line 305: This function remainded me about `iszero`, which seems to work for `BareInterval`s (!), though I don't see how:
```julia
julia> iszero(bareinterval(0.1))
false

julia> @which iszero(bareinterval(0.1))
iszero(x)
     @ Base number.jl:42

julia> iszero(interval(0.1))
ERROR: ArgumentError: `==` is purposely not supported for intervals. See instead `isequal_interval`.
Stacktrace:
 [1] ==(#unused#::Interval{Float64}, #unused#::Interval{Float64})
   @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/real_interface.jl:70
 [2] iszero(x::Interval{Float64})
   @ Base ./number.jl:42
 [3] top-level scope
   @ REPL[86]:1
```

This comment is therefore has two points:
- Either `iszero` shouldn't work with `BareInterval` (at least not if it involves promotion) or at all, and thus favor using `isthinzero`, or we should rely on it completely (and remove `isthinzero`).
- The error thrown by `iszero(::Interval)` shows the use of `==`, instead of `isequal_interval`, which should be corrected.

src/intervals/interval_operations/boolean.jl
```diff
@@ -4,200 +4,331 @@
...
+
+# extension
+
+isequal_interval(x, y, z, w...) = isequal_interval(x, y) & isequal_interval(y, z, w...)
+isequal_interval(x::Complex, y::Complex) = isequal_interval(real(x), real(y)) & isequal_interval(imag(x), imag(y))
+isequal_interval(x::Complex, y::Real) = isequal_interval(real(x), y) & isthinzero(imag(x))
+isequal_interval(x::Real, y::Complex) = isequal_interval(x, real(y)) & isthinzero(imag(y))
+
+isdisjoint_interval(x::Complex, y::Complex) =
+    isdisjoint_interval(real(x), real(y)) | isdisjoint_interval(imag(x), imag(y))
```
lbenet, line 330: Clearly, this function makes only sense for intervals/bareintervals. Yet, the signature (`::Complex`) seems like type-piracy because there is no reference to any struct defined by IntervalArithmetic. I think we should have `Complex{<:Interval}` or `Complex{<:BareInterval}`, or something alike?

src/intervals/interval_operations/boolean.jl
```diff
@@ -4,200 +4,331 @@
...
+
+in_interval(x::Complex, y::Complex) = in_interval(real(x), real(y)) & in_interval(imag(x), imag(y))
+in_interval(x::Complex, y::Real) = in_interval(real(x), y) & in_interval(imag(x), 0)
+in_interval(x::Real, y::Complex) = in_interval(x, real(y)) & in_interval(0, imag(y))
```
lbenet, line 334: Same comment from above...

src/intervals/interval_operations/constants.jl
```diff
@@ -4,40 +4,43 @@
...
 """
-emptyinterval(::Type{Interval{T}}) where {T<:NumTypes} = unsafe_interval(T, typemax(T), typemin(T))
-emptyinterval(::Type{T}) where {T<:NumTypes} = emptyinterval(Interval{T})
-emptyinterval(::Type{<:Real}) = emptyinterval(default_numtype())
-emptyinterval() = emptyinterval(default_numtype())
+emptyinterval(::Type{BareInterval{T}}) where {T<:AbstractFloat} = _unsafe_bareinterval(T, convert(T, NaN), convert(T, NaN))
+emptyinterval(::Type{BareInterval{T}}) where {T<:Rational} = _unsafe_bareinterval(T, typemax(T), typemin(T))
+# note: `using Base.unsafe_rational(Int, 0, 0)` as an equivalent to `NaN` for `Rational`
+# does not work well since most codes for `Rational` assume that the denominator cannot be zero
+# e.g. `iszero(Base.unsafe_rational(Int, 0, 0)) == true`
```
lbenet, line 16: I think this is worth reporting to Julia!

src/intervals/interval_operations/numeric.jl
```diff
@@ -127,39 +139,94 @@ Return the midpoint of an interval `a` together with its radius.
...
 """
-function mig(a::Interval{T}) where {T<:NumTypes}
+function mig(a::BareInterval{T}) where {T<:NumTypes}
     isempty_interval(a) && return convert(T, NaN)
     in_interval(0, a) && return zero(T)
     return min(abs(inf(a)), abs(sup(a)))
 end
 
-mig(a::Real) = abs(a)
+dist(a::BareInterval, b::BareInterval) = max(abs(inf(a)-inf(b)), abs(sup(a)-sup(b)))
```
lbenet, line 172: Docstrings?

src/intervals/arithmetic/hyperbolic.jl
OlivierHnt: Oops I will edit this, all the hyperbolic functions are defined.

src/intervals/interval_operations/boolean.jl
OlivierHnt: Cf. https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/590#discussion_r1411947601.

src/intervals/interval_operations/constants.jl
OlivierHnt: Good idea! I will do this over the week-end.

src/intervals/interval_operations/numeric.jl
OlivierHnt: Not sure where this function came from, I do not see it in the standard (maybe I missed it 🤷‍♂️). But I added a small docstring regardlesss.

src/intervals/construction.jl
OlivierHnt: Done.

src/intervals/construction.jl
OlivierHnt: Done.

docs/src/manual/construction.md
OlivierHnt: Done.

src/intervals/interval_operations/numeric.jl
lbenet: I think it comes from the book of Tucker, though I should need to have a look. If I recall correctly, it is the Housdorff distance

lbenet:
I finished all files in `src/`. I left few more comments, as said, nothing essential, though for the `Symbols` submodule adding some hints on how to get the actual UTF symbol would be very helpful. 

Give me the weekend to go through the modified test files...

OlivierHnt:
Sounds good, but do not waste time reviewing the ITF1788 test suite. Indeed, it will be overwritten in the next PR, since the test suite is now automatically generated (and the entire test suite is passing 🥳!).

So only the files in `test/interval_tests` are worth spending time on (but this should not be long).

src/symbols.jl
```diff
@@ -17,38 +17,70 @@ julia> using IntervalArithmetic
...
+"""
+±(m, r) = interval(m, r; format = :midpoint)
 
 """
     ≛(a, b)
 
 Unicode alias of [`isequal_interval`](@ref).
 """
-const ≛ = isequal_interval
+≛ = isequal_interval
```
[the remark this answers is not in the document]
lbenet: I didn't know the trick.... So no need for any change

src/intervals/arithmetic/basic.jl
lbenet: Thanks for the explanation! Forget my comment...

lbenet (approved):
Let me first thank you @OlivierHnt for all the enormous effort put in this PR. I agree with all your responses to my comments, perhaps just keeping in mind (i) operating with integers, and (ii) `atomic`. I am in favor of aproving this PR, and merge it! Thanks again!

Issue #543, opened 2022-09-26
Fractional powers of nonpositive intervals containing zero


cbilz:
`^` returns the zero interval while `pow` returns the empty interval:

```julia
julia> using IntervalArithmetic

julia> a = interval(-1.0, 0.0)
[-1, 0]

julia> a^1.7
[0, 0]

julia> pow(a,1.7)
∅

julia> b = interval(0.0)
[0, 0]

julia> b^1.7
[0, 0]

julia> pow(b,1.7)
∅
```

This seems inconsistent. Does ITF1788 say what the correct result should be?

lucaferranti:
sorry I forgot to answer, as you figured out the `^` is currently the standard conforming behavior. ITF1788 has a (similar) test for what you're doing [here](https://github.com/oheim/ITF1788/blob/master/itl/libieeep1788_elem.itl#L2865).

Issue #592, opened 2023-12-01
Replacing CRlibm with CORE-MATH


OlivierHnt:
In a recent triage call, we discussed replacing CRlibm with [CORE-MATH](https://gitlab.inria.fr/core-math/core-math/), which provides a more complete set of correctly rounded functions.

On particular appeal is that they give a correctly rounded version of `pow`, which should fix long standing performance issues for powers in IntervalArithmetic.jl.

@giordano I noticed that you created the CRlibm_jll binaries with Yggdrasil.jl. I tried to reproduce this on my computer today for core-math, but I am out of my depth here and could not make it work. The repository on GitLab is not a library per se, but it contains all the C functions. Do you think we can compile CORE-MATH into some CoreMath_jll binaries as easily?

lucaferranti:
havent forgotten, working hard hardly working

dpsanders:
The other approach with this is performance. Changing the rounding mode clears out the processor pipeline (or at least used to?) so will be much slower (?)

OlivierHnt:
Interesting, thanks for pointing this out.
Using Core-Math does not solve all the performance issues, but it does (or seems to) avoid using MPFR.
I will have to take a closer look at `fesetround`, I do not know much on the inner workings of these things. I do hope that a modern and maintained library for correctly rounded functions can be incorporated into Julia 🙂.

Issue #591, opened 2023-11-27
Make sure that the documentation is correct


Kolaru:
A lot of change have happened, and the documentation did not fully follow.

For example, some of the examples still use `IntervalBox`.

As discussed during today's call, once #584 is merge, this is the last issue to be fix before we move on to release 1.0 (first with a transition period with v0.22 that will give us time to update the downstream packages of the JuliaIntervals organization and make sure IntervalArithmetic in its new form is okay).

It doesn't need to be pretty for v1.0, just to be correct.

OlivierHnt:
Also, we should keep an eye on #524.

Issue #524, opened 2022-05-23
Docs: how to handle branching?


petvana:
This is more like a question. I haven't found any information in the docs about handling branching while using `InrervalArithmetic`. For example

```julia
julia> using IntervalArithmetic

julia> f(x) = x > 1 ? x-1 : x
f (generic function with 1 method)

julia> f(0.9..1.1)
[0.899999, 1.10001] # expecting [0.0, 1.0]

julia> f.([0.9, 1.1])
2-element Vector{Float64}:
 0.9
 0.10000000000000009
```

The code produces an incorrect interval because `0.9..1.1 > 1 == false` is valid only for part of the interval. Is there any way to handle such situations? 

Btw, what about producing a warning?

lucaferranti:
Hehe this is an oldie but goldie.

Currently, IA uses the definition recommended by the IEEE standard for interval arithmetic, that is 

$$
x < y \leftrightarrow \underline{x} < \underline{y} \land \overline{x} < \overline{y}
$$

this definition stems from the work of Ulrich Kulisch about axiomatization of interval arithmetic, but in practice is not that useful and indeed it seems to unify developers in disappointment.

The `1.0-dev` branch which, as the name suggests, is for the upcoming 1.0 version of IntervalArithmetic.jl, has a configuration mechanism that allows the user to choose how to handle boolean operations. Incidentally, I had a PR related to that yesterday, so you can have a sneak preview of that [here](#523).

As a final small comment, in the specific example you posted I wouldn't say the result in *incorrect* because it still holds $[0.9, 1] \subseteq [0.899999, 1.100001]$, that is the function returned a valid enclosure of the true range, which is the only promise of interval arithmetic. Especially for more complex functions, you should expect some overestimation (this is known as dependecy problem).

To play devil's advocate, here is a function with branching where a valid enclosure is not returned, so your point stands :)

```julia
julia> f(x)= x > 0 ? x + 3 : x - 3
f (generic function with 1 method)

julia> f(-1..1)
[-4, -2] # the true range is [-4, 4]
```

lucaferranti:
alternatively, you may want to have a look at [NumberIntervals.jl](https://github.com/gwater/NumberIntervals.jl) which uses a ternary logic for boolean intervals, returning `missing` for ambiguous cases (this will also be an option in 1.0-dev). Using that, however, your code would simply error for ambiguous cases, so depending on your application that may or may not be useful

[...]

PR #593, opened 2023-12-01
Beta 0.22

### LINKED ISSUE #84: Bisection should ignore atomic intervals

### LINKED ISSUE #323: Implement reverse mode elementary functions
Unless I've missed a file somewhere, the reverse mode elementary functions required for the set based flavor by the IEEE Std 1788-2015 (section 10.5.4) are not implemented.

### LINKED ISSUE #431: Mince should return an iterator

### LINKED ISSUE #465: ITF1788 compliance

### LINKED ISSUE #476: What is required to use custom types in `Interval`s?
I have a custom numeric type, and I'd like to use it in conjunction (composition?) with IntervalAnalysis.jl. What function implementations are necessary in order to make this happen? Is there an API of sorts for what types can be used as the lower/upper bounds?

I'm currently facing the current error:
```
ERROR: LoadError: MethodError: no method matching atomic(::Type{IntervalArithmetic.Interval{TracedLeaf}}, ::IntervalArithmetic.Interval{TracedLeaf})
Closest candidates are:
  atomic(::Type{IntervalArithmetic.Interval{T}}, ::IntervalArithmetic.Interval) where T<:AbstractFloat at /Users/skainswo/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/conversion.jl:122
  atomic(::Type{IntervalArithmetic.Interval{T}}, ::S) where {T<:AbstractFloat, S<:Real} at /Users/skainswo/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/conversion.jl:95
Stacktrace:
 [1] convert(#unused#::Type{IntervalArithmetic.Interval{TracedLeaf}}, x::IntervalArithmetic.Interval{TracedLeaf})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/conversion.jl:20
 [2] Pair
   @ ./pair.jl:12 [inlined]
 [3] Pair(a::String, b::IntervalArithmetic.Interval{TracedLeaf})
   @ Base ./pair.jl:15
 [4] top-level scope
   @ ~/dev/research/julia/boxes/src/test.jl:203
in expression starting at /Users/skainswo/dev/research/julia/boxes/src/test.jl:199
```
But I'm not able to find the `atomic` function in the public API: https://juliaintervals.github.io/pages/apiDocs/apiIntervalArithmetic/index.html.

### LINKED ISSUE #486: Float16 setrounding error
Does anyone know why this is happening?

```julia
using IntervalArithmetic

a = interval(Float16(1.0), Float16(3.0))
a + a
```

gives:
```Julia

ERROR: MethodError: no method matching rounding_raw(::Type{Float16})
Closest candidates are:
  rounding_raw(::Type{BigFloat}) at mpfr.jl:121
  rounding_raw(::Type{var"#s79"} where var"#s79"<:Union{Float32, Float64}) at rounding.jl:155
Stacktrace:
 [1] setrounding(f::IntervalArithmetic.var"#11#12"{Float16, Float16}, #unused#::Type{Float16}, rounding::RoundingMode{:Down})
   @ Base.Rounding ./rounding.jl:173
 [2] +(#unused#::IntervalArithmetic.IntervalRounding{:tight}, a::Float16, b::Float16, #unused#::RoundingMode{:Down})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/rounding.jl:177
 [3] +(a::Float16, b::Float16, r::RoundingMode{:Down})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/rounding.jl:278
 [4] +(a::Interval{Float16}, b::Interval{Float16})
   @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/UR6Qe/src/intervals/arithmetic.jl:86
 [5] top-level scope
   @ REPL[5]:1
```

I get the same error when:

```julia
lo = Float16(1)
hi = Float16(3)

out = setrounding(Float16, RoundDown) do
	collect(range(lo, hi, length = 10))
end
```

giving:

```julia
ERROR: MethodError: no method matching rounding_raw(::Type{Float16})
Closest candidates are:
  rounding_raw(::Type{BigFloat}) at mpfr.jl:121
  rounding_raw(::Type{var"#s79"} where var"#s79"<:Union{Float32, Float64}) at rounding.jl:155
Stacktrace:
 [1] setrounding(f::var"#1#2", #unused#::Type{Float16}, rounding::RoundingMode{:Down})
   @ Base.Rounding ./rounding.jl:173
 [2] top-level scope
   @ REPL[8]:1

```

### LINKED ISSUE #490: How should strict enclosure of interval boxes be defined?
Currently,

```julia
julia> a = (1..2) × (3..4)
[1, 2] × [3, 4]

julia> b = (1.5..1.9) × (3..4)
[1.5, 1.90001] × [3, 4]

julia> b ⊂ a
false
```

That is currently it returns true if **all** intervals in the first box are strictly contained in the corresponding intervals in the second box.

I wonder whether this is correct, as generally `b ⊂ a` means that `b` is a subset of `a` but is not equal to it, so according to this the example above should return true. 

**edit**: swapped `a` and `b`, as they were in the wrong order in the original example

### LINKED ISSUE #540: IntervalArithmetic.jl testing errors / broken found.
@v1.7) pkg> add IntervalArithmetic
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
   Installed SetRounding ───────── v0.2.1
   Installed FastRounding ──────── v0.3.1
   Installed CRlibm_jll ────────── v1.0.1+0
   Installed RoundingEmulator ──── v0.2.1
   Installed ErrorfreeArithmetic ─ v0.5.2
   Installed CRlibm ────────────── v1.0.1
   Installed IntervalArithmetic ── v0.20.6
  Downloaded artifact: CRlibm
    Updating `~/.julia/environments/v1.7/Project.toml`
  [d1acc4aa] + IntervalArithmetic v0.20.6
    Updating `~/.julia/environments/v1.7/Manifest.toml`
  [96374032] + CRlibm v1.0.1
  [90fa49ef] + ErrorfreeArithmetic v0.5.2
  [fa42c844] + FastRounding v0.3.1
  [d1acc4aa] + IntervalArithmetic v0.20.6
  [5eaf0fd0] + RoundingEmulator v0.2.1
  [3cc68bcd] + SetRounding v0.2.1
  [4e9b3aee] + CRlibm_jll v1.0.1+0
Precompiling project...
  7 dependencies successfully precompiled in 30 seconds (417 already precompiled, 1 skipped during auto due to previous errors)

(@v1.7) pkg> test IntervalArithmetic
     Testing IntervalArithmetic
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.7/Pkg/src/Operations.jl:1488
      Status `/tmp/jl_UGQbQJ/Project.toml`
  [96374032] CRlibm v1.0.1
  [fa42c844] FastRounding v0.3.1
  [d1acc4aa] IntervalArithmetic v0.20.6
  [15111844] IntervalContractors v0.4.7
  [f27b6e38] Polynomials v0.7.0
  [3cdcf5f2] RecipesBase v1.2.1
  [5eaf0fd0] RoundingEmulator v0.2.1
  [3cc68bcd] SetRounding v0.2.1
  [90137ffa] StaticArrays v1.5.2
  [37e2e46d] LinearAlgebra `@stdlib/LinearAlgebra`
  [d6f4376e] Markdown `@stdlib/Markdown`
  [9a3f8284] Random `@stdlib/Random`
  [8dfed614] Test `@stdlib/Test`
      Status `/tmp/jl_UGQbQJ/Manifest.toml`
  [96374032] CRlibm v1.0.1
  [34da2185] Compat v3.45.0
  [90fa49ef] ErrorfreeArithmetic v0.5.2
  [e2ba6199] ExprTools v0.1.8
  [fa42c844] FastRounding v0.3.1
  [842dd82b] InlineStrings v1.1.4
  [d1acc4aa] IntervalArithmetic v0.20.6
  [15111844] IntervalContractors v0.4.7
  [d8418881] Intervals v1.8.0
  [692b3bcd] JLLWrappers v1.4.1
  [78c3b35d] Mocking v0.7.3
  [69de0a69] Parsers v2.3.2
  [f27b6e38] Polynomials v0.7.0
  [21216c6a] Preferences v1.3.0
  [3cdcf5f2] RecipesBase v1.2.1
  [5eaf0fd0] RoundingEmulator v0.2.1
  [6c6a2e73] Scratch v1.1.1
  [3cc68bcd] SetRounding v0.2.1
  [90137ffa] StaticArrays v1.5.2
  [1e83bf80] StaticArraysCore v1.0.1
  [f269a46b] TimeZones v1.9.0
  [4e9b3aee] CRlibm_jll v1.0.1+0
  [0dad84c5] ArgTools `@stdlib/ArgTools`
  [56f22d72] Artifacts `@stdlib/Artifacts`
  [2a0f44e3] Base64 `@stdlib/Base64`
  [ade2ca70] Dates `@stdlib/Dates`
  [8bb1440f] DelimitedFiles `@stdlib/DelimitedFiles`
  [8ba89e20] Distributed `@stdlib/Distributed`
  [f43a241f] Downloads `@stdlib/Downloads`
  [7b1f6079] FileWatching `@stdlib/FileWatching`
  [b77e0a4c] InteractiveUtils `@stdlib/InteractiveUtils`
  [4af54fe1] LazyArtifacts `@stdlib/LazyArtifacts`
  [b27032c2] LibCURL `@stdlib/LibCURL`
  [76f85450] LibGit2 `@stdlib/LibGit2`
  [8f399da3] Libdl `@stdlib/Libdl`
  [37e2e46d] LinearAlgebra `@stdlib/LinearAlgebra`
  [56ddb016] Logging `@stdlib/Logging`
  [d6f4376e] Markdown `@stdlib/Markdown`
  [a63ad114] Mmap `@stdlib/Mmap`
  [ca575930] NetworkOptions `@stdlib/NetworkOptions`
  [44cfe95a] Pkg `@stdlib/Pkg`
  [de0858da] Printf `@stdlib/Printf`
  [3fa0cd96] REPL `@stdlib/REPL`
  [9a3f8284] Random `@stdlib/Random`
  [ea8e919c] SHA `@stdlib/SHA`
  [9e88b42a] Serialization `@stdlib/Serialization`
  [1a1011a3] SharedArrays `@stdlib/SharedArrays`
  [6462fe0b] Sockets `@stdlib/Sockets`
  [2f01184e] SparseArrays `@stdlib/SparseArrays`
  [10745b16] Statistics `@stdlib/Statistics`
  [fa267f1f] TOML `@stdlib/TOML`
  [a4e569a6] Tar `@stdlib/Tar`
  [8dfed614] Test `@stdlib/Test`
  [cf7118a7] UUIDs `@stdlib/UUIDs`
  [4ec0a83e] Unicode `@stdlib/Unicode`
  [e66e0078] CompilerSupportLibraries_jll `@stdlib/CompilerSupportLibraries_jll`
  [deac9b47] LibCURL_jll `@stdlib/LibCURL_jll`
  [29816b5a] LibSSH2_jll `@stdlib/LibSSH2_jll`
  [c8ffd9c3] MbedTLS_jll `@stdlib/MbedTLS_jll`
  [14a3606d] MozillaCACerts_jll `@stdlib/MozillaCACerts_jll`
  [4536629a] OpenBLAS_jll `@stdlib/OpenBLAS_jll`
  [83775a58] Zlib_jll `@stdlib/Zlib_jll`
  [8e850b90] libblastrampoline_jll `@stdlib/libblastrampoline_jll`
  [8e850ede] nghttp2_jll `@stdlib/nghttp2_jll`
  [3f19e933] p7zip_jll `@stdlib/p7zip_jll`
Precompiling project...
  5 dependencies successfully precompiled in 57 seconds (23 already precompiled)
     Testing Running tests...
Test Summary:          | Pass  Broken  Total
Constructing intervals |  120       1    121
Test Summary: | Pass  Broken  Total
Big intervals |    6       1      7
Test Summary:     | Pass  Total
Complex intervals |    2      2
Test Summary:   | Pass  Total
Typed intervals |    9      9
Test Summary: | Pass  Total
.. tests      |   17     17
Test Summary: | Pass  Total
± tests       |    5      5
Test Summary:                       | Pass  Total
Conversion to interval of same type |    2      2
Test Summary:                                   | Pass  Total
Conversions between different types of interval |    2      2
Test Summary:          | Pass  Total
Conversion to Interval |    3      3
Test Summary:           | Pass  Total
Interval{T} constructor |    4      4
Test Summary:               | Pass  Total
Disallow NaN in an interval |    6      6
Test Summary:    | Pass  Total
Interval strings |    4      4
Test Summary: | Pass  Total
setdiff tests |    6      6
Test Summary:            | Pass  Total
Interval{T}(x::Interval) |    2      2
Test Summary:         | Pass  Total
@interval with fields |    1      1
Test Summary:                        | Pass  Total
@interval with user-defined function |    1      1
Test Summary:   | Pass  Total
a..b with a > b |    2      2
Test Summary:        | Pass  Total
Hashing of Intervals |    8      8
Test Summary:  | Pass  Total
force_interval |    6      6
Test Summary: | Pass  Total
Zero interval |    2      2
Test Summary:     | Pass  Total
Consistency tests |  250    250
Test Summary:     | Pass  Total
brodcasting tests |   81     81
Test Summary: | Pass  Total
Numeric tests |   35     35
Test Summary:             | Pass  Total
Arithmetic with constants |    5      5
Test Summary: | Pass  Total
Power tests   |   41     41
Test Summary:     | Pass  Total
Exp and log tests |   20     20
Test Summary:    | Pass  Total
Comparison tests |    8      8
Test Summary:  | Pass  Total
Rational tests |    8      8
Test Summary:    | Pass  Total
Floor etc. tests |   53     53
Test Summary: | Pass  Total
Fast power    |   37     37
Test Summary:         | Pass  Total
Mince for `Interval`s |    6      6
Test Summary: | Pass  Total
nthroot test  |   36     36
Test Summary: | Pass  Total
sin           |   14     14
Test Summary: | Pass  Total
cos           |   12     12
Test Summary: | Pass  Total
sinpi         |    5      5
Test Summary: | Pass  Total
cospi         |    5      5
Test Summary: | Pass  Total
tan           |   12     12
Test Summary: | Pass  Total
Inverse trig  |   15     15
Test Summary: | Pass  Total
atan          |   55     55
Test Summary: | Pass  Total
Trig          |    4      4
Test Summary:             | Pass  Total
Trig with large arguments |    7      7
Test Summary: | Pass  Total
Hyperb tests  |   48     48
Test Summary:                 | Pass  Total
Tests with rational intervals |    3      3
Test Summary:               | Pass  Total
Rounding rational intervals |    1      1
Test Summary:              | Pass  Total
Tests with float intervals |    4      4
Test Summary:                  | Pass  Total
Testing functions of intervals |    2      2
Test Summary:       | Pass  Total
Testing conversions |    2      2
Test Summary:                       | Pass  Total
Linear algebra with intervals tests |    2      2
Test Summary:       | Pass  Total
Interval loop tests |   12     12
Test Summary: | Pass  Total
Pi tests      |    7      7
Test Summary:            | Pass  Total
Parse string to Interval |   31     31
Test Summary:                     | Pass  Total
Parse string to DecoratedInterval |   20     20
Test Summary:                                  | Pass  Total
Tight, slow rounding by changing rounding mode |    2      2
Test Summary:                                         | Pass  Total
Accurate, fast rounding using prevfloat and nextfloat |    2      2
Test Summary: | Pass  Total
No rounding   |    3      3
Test Summary:                              | Pass  Total
Tight, fast rounding using FastRounding.jl |    2      2
Test Summary:                            | Pass  Total
Tight rounding using RoundingEmulator.jl |    6      6
Test Summary:     | Pass  Total
`bisect` function |   14     14
Test Summary:               | Pass  Total
Complex interval operations |   16     16
Test Summary:     | Pass  Total
Complex functions |    8      8
Test Summary:  | Pass  Total
Complex powers |    4      4
Test Summary: | Pass  Total
abs2 and abs  |    4      4
Test Summary:  | Pass  Total
real functions |    3      3
Test Summary:           | Pass  Total
DecoratedInterval tests |   37     37
Test Summary:                       | Pass  Total
Convert string to DecoratedInterval |    2      2
Test Summary: | Pass  Total
rand tests    |  330    330
Test Summary:   | Pass  Total
setformat tests |   58     58
Test Summary: | Pass  Total
showfull      |   12     12
Test Summary: | Pass  Total
@format tests |    3      3
Test Summary:       | Pass  Total
Operations on boxes |  144    144
Test Summary:      | Pass  Total
Functions on boxes |    5      5
Test Summary:           | Pass  Total
setdiff for IntervalBox |   10     10
Test Summary:                | Pass  Total
mid, diam, × for IntervalBox |   21     21
Test Summary:                               | Pass  Total
Constructing multidimensional IntervalBoxes |   16     16
Test Summary:         | Pass  Total
getindex and setindex |    5      5
Test Summary: | Pass  Total
Iteration     |    2      2
Test Summary: | Pass  Total
Broadcasting  |   23     23
Test Summary: | Pass  Total
∈             |    6      6
Test Summary:              | Pass  Total
Multiplication by a matrix |    2      2
Test Summary:                      | Pass  Total
Mince and hull for `IntervalBox`es |   18     18
Test Summary:            | Pass  Total
Special box constructors |    3      3
Test Summary:      | Pass  Total
minimal.atan2_test |   38     38
Test Summary:       | Pass  Total
cxsc.intervaladdsub |    6      6
Test Summary:       | Pass  Total
cxsc.intervalmuldiv |   31     31
Test Summary:       | Pass  Total
cxsc.intervalsetops |   24     24
Test Summary:          | Pass  Total
cxsc.intervalmixsetops |   12     12
Test Summary:        | Pass  Total
cxsc.scalarmixsetops |    6      6
Test Summary:           | Pass  Total
cxsc.intervalsetcompops |   34     34
Test Summary:                 | Pass  Total
cxsc.intervalscalarsetcompops |   35     35
Test Summary:        | Pass  Total
cxsc.intervalstdfunc |   12     12
Test Summary: | Pass  Total
FI_LIB.addii  |   19     19
Test Summary: | Pass  Total
FI_LIB.subii  |   19     19
Test Summary: | Pass  Total
FI_LIB.mulii  |   46     46
Test Summary: | Pass  Total
FI_LIB.divii  |   21     21
Test Summary:          | Pass  Broken  Total
FI_LIB.unary_functions |  708      50    758
Test Summary: | Pass  Total
IEEE1788.a    |    1      1
Test Summary: | Pass  Total
IEEE1788.b    |    4      4
Test Summary: | Pass  Total
IEEE1788.c    |   11     11
Test Summary: | Pass  Total
IEEE1788.d    |    3      3
Test Summary: | Pass  Total
IEEE1788.e    |    1      1
Test Summary: | Pass  Total
IEEE1788.e    |   18     18
Test Summary: | Pass  Total
IEEE1788.f    |    5      5
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/LgwfS/src/intervals/intervals.jl:106
Test Summary: | Pass  Broken  Total
exceptions    |    1       3      4
Test Summary:         | Pass  Total
minimal_is_empty_test |   14     14
Test Summary:             | Pass  Total
minimal_is_empty_dec_test |   15     15
Test Summary:          | Pass  Total
minimal_is_entire_test |   14     14
Test Summary:              | Pass  Total
minimal_is_entire_dec_test |   17     17
Test Summary:           | Pass  Total
minimal_is_nai_dec_test |   16     16
Test Summary:      | Pass  Total
minimal_equal_test |   15     15
Test Summary:          | Pass  Total
minimal_equal_dec_test |   19     19
Test Summary:       | Pass  Total
minimal_subset_test |   27     27
Test Summary:           | Pass  Total
minimal_subset_dec_test |   29     29
Test Summary:     | Pass  Total
minimal_less_test |   26     26
Test Summary:         | Pass  Total
minimal_less_dec_test |   30     30
Test Summary:         | Pass  Total
minimal_precedes_test |   21     21
Test Summary:             | Pass  Total
minimal_precedes_dec_test |   25     25
Test Summary:         | Pass  Total
minimal_interior_test |   16     16
Test Summary:             | Pass  Total
minimal_interior_dec_test |   20     20
Test Summary:              | Pass  Total
minimal_strictly_less_test |   14     14
Test Summary:                  | Pass  Total
minimal_strictly_less_dec_test |   18     18
Test Summary:                  | Pass  Total
minimal_strictly_precedes_test |   14     14
Test Summary:                      | Pass  Total
minimal_strictly_precedes_dec_test |   18     18
Test Summary:         | Pass  Total
minimal_disjoint_test |   10     10
Test Summary:             | Pass  Total
minimal_disjoint_dec_test |   14     14
Test Summary:            | Pass  Total
minimal_cancel_plus_test |   58     58
Test Summary:                | Pass  Total
minimal_cancel_plus_dec_test |   58     58
Test Summary:             | Pass  Total
minimal_cancel_minus_test |   63     63
Test Summary:                 | Pass  Total
minimal_cancel_minus_dec_test |   63     63
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/LgwfS/src/intervals/intervals.jl:106
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/LgwfS/src/intervals/intervals.jl:106
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/LgwfS/src/intervals/intervals.jl:106
┌ Warning: Invalid input, empty interval is returned
└ @ IntervalArithmetic ~/.julia/packages/IntervalArithmetic/LgwfS/src/intervals/intervals.jl:106
Test Summary:                 | Pass  Total
minimal_nums_to_interval_test |    8      8
Test Summary:                           | Pass  Total
minimal_nums_to_decorated_interval_test |    8      8
Test Summary:                 | Pass  Broken  Total
minimal_text_to_interval_test |   38      30     68
Test Summary:                           | Pass  Broken  Total
minimal_text_to_decorated_interval_test |   36      35     71
Test Summary:              | Pass  Broken  Total
minimal_interval_part_test |   13       1     14
Test Summary:        | Pass  Total
minimal_new_dec_test |   13     13
Test Summary:        | Pass  Broken  Total
minimal_set_dec_test |   16       6     22
Test Summary:                | Pass  Total
minimal_decoration_part_test |    6      6
Test Summary:    | Pass  Total
minimal_pos_test |   11     11
Test Summary:        | Pass  Total
minimal_pos_dec_test |    4      4
Test Summary:    | Pass  Total
minimal_neg_test |   11     11
Test Summary:        | Pass  Total
minimal_neg_dec_test |    4      4
Test Summary:    | Pass  Total
minimal_add_test |   31     31
Test Summary:        | Pass  Total
minimal_add_dec_test |    6      6
Test Summary:    | Pass  Total
minimal_sub_test |   31     31
Test Summary:        | Pass  Total
minimal_sub_dec_test |    6      6
Test Summary:    | Pass  Total
minimal_mul_test |  116    116
Test Summary:        | Pass  Total
minimal_mul_dec_test |    6      6
Test Summary:    | Pass  Total
minimal_div_test |  341    341
Test Summary:        | Pass  Total
minimal_div_dec_test |    6      6
Test Summary:      | Pass  Total
minimal_recip_test |   18     18
Test Summary:          | Pass  Total
minimal_recip_dec_test |    8      8
Test Summary:    | Pass  Total
minimal_sqr_test |   12     12
Test Summary:        | Pass  Total
minimal_sqr_dec_test |    4      4
Test Summary:     | Pass  Total
minimal_sqrt_test |   13     13
Test Summary:         | Pass  Total
minimal_sqrt_dec_test |    4      4
Test Summary:    | Pass  Total
minimal_fma_test |  564    564
Test Summary:        | Pass  Total
minimal_fma_dec_test |    3      3
Test Summary:     | Pass  Total
minimal_pown_test |  163    163
Test Summary:         | Pass  Total
minimal_pown_dec_test |   11     11
Test Summary:    | Pass  Total
minimal_pow_test | 1344   1344
Test Summary:        | Pass  Total
minimal_pow_dec_test |   84     84
Test Summary:    | Pass  Total
minimal_exp_test |   19     19
Test Summary:        | Pass  Total
minimal_exp_dec_test |    2      2
Test Summary:     | Pass  Total
minimal_exp2_test |   18     18
Test Summary:         | Pass  Total
minimal_exp2_dec_test |    2      2
Test Summary:      | Pass  Total
minimal_exp10_test |   19     19
Test Summary:          | Pass  Total
minimal_exp10_dec_test |    2      2
Test Summary:    | Pass  Total
minimal_log_test |   21     21
Test Summary:        | Pass  Total
minimal_log_dec_test |    3      3
Test Summary:     | Pass  Total
minimal_log2_test |   19     19
Test Summary:         | Pass  Total
minimal_log2_dec_test |    4      4
Test Summary:      | Pass  Total
minimal_log10_test |   20     20
Test Summary:          | Pass  Total
minimal_log10_dec_test |    2      2
Test Summary:    | Pass  Total
minimal_sin_test |   52     52
Test Summary:        | Pass  Total
minimal_sin_dec_test |    3      3
Test Summary:    | Pass  Total
minimal_cos_test |   52     52
Test Summary:        | Pass  Total
minimal_cos_dec_test |    3      3
Test Summary:    | Pass  Total
minimal_tan_test |   33     33
Test Summary:        | Pass  Total
minimal_tan_dec_test |   33     33
Test Summary:     | Pass  Total
minimal_asin_test |   18     18
Test Summary:         | Pass  Total
minimal_asin_dec_test |    5      5
Test Summary:     | Pass  Total
minimal_acos_test |   18     18
Test Summary:         | Pass  Total
minimal_acos_dec_test |    5      5
Test Summary:     | Pass  Total
minimal_atan_test |   10     10
Test Summary:         | Pass  Total
minimal_atan_dec_test |    5      5
Test Summary:      | Pass  Total
minimal_atan2_test |  169    169
Test Summary:          | Pass  Total
minimal_atan2_dec_test |  169    169
Test Summary:     | Pass  Total
minimal_sinh_test |   11     11
Test Summary:         | Pass  Total
minimal_sinh_dec_test |    5      5
Test Summary:     | Pass  Total
minimal_cosh_test |   11     11
Test Summary:         | Pass  Total
minimal_cosh_dec_test |    5      5
Test Summary:     | Pass  Total
minimal_tanh_test |   11     11
Test Summary:         | Pass  Total
minimal_tanh_dec_test |    5      5
Test Summary:      | Pass  Total
minimal_asinh_test |   11     11
Test Summary:          | Pass  Total
minimal_asinh_dec_test |    5      5
Test Summary:      | Pass  Total
minimal_acosh_test |   11     11
Test Summary:          | Pass  Total
minimal_acosh_dec_test |    8      8
Test Summary:      | Pass  Total
minimal_atanh_test |   15     15
Test Summary:          | Pass  Total
minimal_atanh_dec_test |    9      9
Test Summary:     | Pass  Total
minimal_sign_test |   11     11
Test Summary:         | Pass  Total
minimal_sign_dec_test |    7      7
Test Summary:     | Pass  Total
minimal_ceil_test |   15     15
Test Summary:         | Pass  Total
minimal_ceil_dec_test |   14     14
Test Summary:      | Pass  Total
minimal_floor_test |   13     13
Test Summary:          | Pass  Total
minimal_floor_dec_test |   12     12
Test Summary:      | Pass  Total
minimal_trunc_test |   13     13
Test Summary:          | Pass  Total
minimal_trunc_dec_test |   12     12
Test Summary:                   | Pass  Total
minimal_round_ties_to_even_test |   18     18
Test Summary:                       | Pass  Total
minimal_round_ties_to_even_dec_test |    6      6
Test Summary:                   | Pass  Total
minimal_round_ties_to_away_test |   18     18
Test Summary:                       | Pass  Total
minimal_round_ties_to_away_dec_test |    7      7
Test Summary:    | Pass  Total
minimal_abs_test |   12     12
Test Summary:        | Pass  Total
minimal_abs_dec_test |    8      8
Test Summary:    | Pass  Total
minimal_min_test |   15     15
Test Summary:        | Pass  Total
minimal_min_dec_test |    4      4
Test Summary:    | Pass  Total
minimal_max_test |   15     15
Test Summary:        | Pass  Total
minimal_max_dec_test |    4      4
Test Summary:    | Pass  Total
minimal_inf_test |   14     14
Test Summary:        | Pass  Total
minimal_inf_dec_test |   15     15
Test Summary:    | Pass  Total
minimal_sup_test |   14     14
Test Summary:        | Pass  Total
minimal_sup_dec_test |   15     15
Test Summary:    | Pass  Broken  Total
minimal_mid_test |   11       1     12
Test Summary:        | Pass  Broken  Total
minimal_mid_dec_test |   12       1     13
Test Summary:    | Pass  Total
minimal_rad_test |    9      9
Test Summary:        | Pass  Total
minimal_rad_dec_test |   10     10
Test Summary:        | Pass  Broken  Total
minimal_mid_rad_test |   11       1     12
Test Summary:            | Pass  Broken  Total
minimal_mid_rad_dec_test |   12       1     13
Test Summary:    | Pass  Total
minimal_wid_test |    8      8
Test Summary:        | Pass  Total
minimal_wid_dec_test |    9      9
Test Summary:    | Pass  Total
minimal_mag_test |    8      8
Test Summary:        | Pass  Total
minimal_mag_dec_test |    9      9
Test Summary:    | Pass  Total
minimal_mig_test |   11     11
Test Summary:        | Pass  Total
minimal_mig_dec_test |   12     12
Test Summary:        | Broken  Total
minimal_overlap_test |     48     48
Test Summary:            | Broken  Total
minimal_overlap_dec_test |     29     29
Test Summary:                   | Pass  Total
minimal_is_common_interval_test |   12     12
Test Summary:                       | Pass  Total
minimal_is_common_interval_dec_test |   21     21
Test Summary:             | Pass  Total
minimal_is_singleton_test |   15     15
Test Summary:                 | Pass  Total
minimal_is_singleton_dec_test |   16     16
Test Summary:          | Pass  Total
minimal_is_member_test |   35     35
Test Summary:              | Pass  Total
minimal_is_member_dec_test |   40     40
Test Summary:    | Pass  Total
minimal_sum_test |    3      3
Test Summary:        | Pass  Total
minimal_sum_abs_test |    3      3
Test Summary:        | Pass  Total
minimal_sum_sqr_test |    3      3
Test Summary:    | Pass  Broken  Total
minimal_dot_test |    5       1      6
Test Summary:             | Pass  Total
minimal_intersection_test |    5      5
Test Summary:                 | Pass  Total
minimal_intersection_dec_test |    5      5
Test Summary:            | Pass  Total
minimal_convex_hull_test |    5      5
Test Summary:                | Pass  Total
minimal_convex_hull_dec_test |    5      5
Test Summary: | Pass  Total
mpfi_abs      |   12     12
Test Summary: | Pass  Total
mpfi_acos     |    8      8
Test Summary: | Pass  Total
mpfi_acosh    |    5      5
Test Summary: | Pass  Total
mpfi_add      |   19     19
Test Summary: | Pass  Total
mpfi_add_d    |   32     32
Test Summary: | Pass  Total
mpfi_asin     |    8      8
Test Summary: | Pass  Total
mpfi_asinh    |   19     19
Test Summary: | Pass  Total
mpfi_atan     |   19     19
Test Summary: | Pass  Total
mpfi_atan2    |   18     18
Test Summary: | Pass  Total
mpfi_atanh    |    9      9
Test Summary:  | Pass  Total
mpfi_bounded_p |   16     16
Test Summary: | Pass  Total
mpfi_cbrt     |   10     10
Test Summary: | Pass  Total
mpfi_cos      |   46     46
Test Summary: | Pass  Total
mpfi_cosh     |   14     14
Test Summary: | Pass  Total
mpfi_cot      |   19     19
Test Summary: | Pass  Total
mpfi_coth     |   16     16
Test Summary: | Pass  Total
mpfi_csc      |  109    109
Test Summary: | Pass  Total
mpfi_csch     |   16     16
Test Summary: | Pass  Total
mpfi_d_div    |   30     30
Test Summary: | Pass  Total
mpfi_diam_abs |   10     10
Test Summary: | Pass  Total
mpfi_div      |   62     62
Test Summary: | Pass  Total
mpfi_div_d    |   25     25
Test Summary: | Pass  Total
mpfi_d_sub    |   32     32
Test Summary: | Pass  Total
mpfi_exp      |   12     12
Test Summary: | Pass  Total
mpfi_exp2     |   13     13
Test Summary: | Pass  Total
mpfi_expm1    |   12     12
Test Summary: | Pass  Total
mpfi_hypot    |   17     17
Test Summary:  | Pass  Total
mpfi_intersect |   14     14
Test Summary: | Pass  Total
mpfi_inv      |   11     11
Test Summary: | Pass  Total
mpfi_is_neg   |   16     16
Test Summary:  | Pass  Total
mpfi_is_nonneg |   16     16
Test Summary:  | Pass  Total
mpfi_is_nonpos |   16     16
Test Summary: | Pass  Total
mpfi_is_pos   |   16     16
Test Summary:        | Pass  Total
mpfi_is_strictly_neg |   16     16
Test Summary:        | Pass  Total
mpfi_is_strictly_pos |   16     16
Test Summary: | Pass  Total
mpfi_log      |    7      7
Test Summary: | Pass  Total
mpfi_log1p    |    7      7
Test Summary: | Pass  Total
mpfi_log2     |    6      6
Test Summary: | Pass  Total
mpfi_log10    |    7      7
Test Summary: | Pass  Total
mpfi_mag      |   10     10
Test Summary: | Pass  Total
mpfi_mid      |   11     11
Test Summary: | Pass  Total
mpfi_mig      |   10     10
Test Summary: | Pass  Total
mpfi_mul      |   50     50
Test Summary: | Pass  Total
mpfi_mul_d    |   45     45
Test Summary: | Pass  Total
mpfi_neg      |    8      8
Test Summary: | Pass  Total
mpfi_put_d    |    3      3
Test Summary: | Pass  Total
mpfi_sec      |  109    109
Test Summary: | Pass  Total
mpfi_sech     |   14     14
Test Summary: | Pass  Total
mpfi_sin      |  128    128
Test Summary: | Pass  Total
mpfi_sinh     |   13     13
Test Summary: | Pass  Total
mpfi_sqr      |   11     11
Test Summary: | Pass  Total
mpfi_sqrt     |    7      7
Test Summary: | Pass  Total
mpfi_sub      |   19     19
Test Summary: | Pass  Total
mpfi_sub_d    |   32     32
Test Summary: | Pass  Total
mpfi_tan      |  128    128
Test Summary: | Pass  Total
mpfi_tanh     |   14     14
Test Summary: | Pass  Total
mpfi_union    |   14     14
Test Summary:          | Pass  Total
minimal.absRevBin_test |   24     24
Test Summary:        | Pass  Total
minimal_sqr_rev_test |   10     10
Test Summary:            | Pass  Total
minimal_sqr_rev_bin_test |   11     11
Test Summary:            | Pass  Total
minimal_sqr_rev_dec_test |   10     10
Test Summary:                | Pass  Total
minimal_sqr_rev_dec_bin_test |   11     11
Test Summary:        | Pass  Total
minimal_abs_rev_test |    9      9
Test Summary:            | Pass  Total
minimal_abs_rev_bin_test |    7      7
Test Summary:            | Pass  Total
minimal_abs_rev_dec_test |    9      9
Test Summary:                | Pass  Total
minimal_abs_rev_dec_bin_test |    7      7
Test Summary:         | Pass  Broken  Total
minimal_pown_rev_test |  135       8    143
Test Summary:             | Pass  Total
minimal_pown_rev_bin_test |   37     37
Test Summary:             | Pass  Broken  Total
minimal_pown_rev_dec_test |  134       8    142
Test Summary:                 | Pass  Total
minimal_pown_rev_dec_bin_test |   36     36
Test Summary:        | Pass  Total
minimal_sin_rev_test |    6      6
Test Summary:            | Pass  Total
minimal_sin_rev_bin_test |   20     20
Test Summary:            | Pass  Total
minimal_sin_rev_dec_test |    6      6
Test Summary:                | Pass  Total
minimal_sin_rev_dec_bin_test |   20     20
Test Summary:        | Pass  Total
minimal_cos_rev_test |    6      6
Test Summary:            | Pass  Total
minimal_cos_rev_bin_test |   21     21
Test Summary:            | Pass  Total
minimal_cos_rev_dec_test |    6      6
Test Summary:                | Pass  Total
minimal_cos_rev_dec_bin_test |   21     21
Test Summary:        | Pass  Total
minimal_tan_rev_test |    5      5
Test Summary:            | Pass  Total
minimal_tan_rev_bin_test |   10     10
Test Summary:            | Pass  Total
minimal_tan_rev_dec_test |    5      5
Test Summary:                | Pass  Total
minimal_tan_rev_dec_bin_test |   10     10
Test Summary:         | Pass  Total
minimal_cosh_rev_test |    5      5
Test Summary:             | Pass  Total
minimal_cosh_rev_bin_test |    5      5
Test Summary:             | Pass  Total
minimal_cosh_rev_dec_test |    5      5
Test Summary:                 | Pass  Total
minimal_cosh_rev_dec_bin_test |    5      5
Test Summary:        | Pass  Total
minimal_mul_rev_test |  172    172
Test Summary:            | Pass  Total
minimal_mul_rev_ten_test |    5      5
Test Summary:            | Pass  Total
minimal_mul_rev_dec_test |   10     10
Test Summary:                | Pass  Total
minimal_mul_rev_dec_ten_test |    5      5
Test Summary:             | Pass  Total
minimal_mulRevToPair_test |  172    172
Test Summary:                 | Pass  Total
minimal_mulRevToPair_dec_test |  175    175
Test Summary:        | Pass  Broken  Total
minimal.powRev1_test |  338      91    429
Test Summary:        | Pass  Broken  Total
minimal.powRev2_test |  233     142    375
     Testing IntervalArithmetic tests passed 

(@v1.7) pkg>

### LINKED ISSUE #564: Derivative of abs(::Interval)

OlivierHnt:
With all the progress made recently, we should be a few commits away from releasing 0.22! If everything goes well, this should be the last version before the 1.0 release 🥳!

### Changes

- changed behaviour of functions in numeric.jl for numbers; e.g. `sup(Inf) = NaN` (`Inf` is first converted into an `Interval`), previous behaviour `sup(Inf) == Inf`
- add `DiffRules` as a weak dependency and hook into `DiffRules._abs_deriv` for decorated intervals (closes #564)
- delete changelog.md (seemed abandonned); update NEWS.md (merged elements of the TODO file into it)
- change default bisection to be at 0.5 by default
- new `mince!` function (in-place `mince`) (closes #431, closes PR#587)
- rename `nthroot` to `rootn`, and `nthpow` to `pown`. It reads better and is consistent with the standard.
- define `isinterior` method (closes #490)
- rename format `:standard` to `:infsup`, feels more consistent with the other format name `:midpoint`
- remove `atomic` since it is an unused internal constructor (see also https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/590#discussion_r1411560983)
- fix `bisect` for atomic and empty intervals (closes #84)
- rename `setdiff_interval` to `interiordiff`, it was a misleading name since the function is not quite the "set difference", rather it removes the interior of the second argument from the first. For instance `x = interval(1, 3)`, `y = interval(2, 4)`, the true set difference should not contain `2`; due to the fact that all intervals are closed intervals (unless the inf or sup is unbounded) we can never truly represent the set difference
- ITF1788 test suite generated and parsed automatically when running tests, ITF1788 compliance for "non rev" functions except one broken test on "dot" which seems unrelated to the library (closes #323, closes #465, closes #540, closes PR#399, closes PR#398)
- add a small section for custom bound type in docs (closes #476)
- after some investigation, it seemed appropriate to drop support for Random. As we discussed in triage, the `rand` method is overloaded in IntervalArithmetic.jl, but it is not doing the right thing. Fixing it appears more tricky than initially thought, see e.g. https://discourse.julialang.org/t/output-distribution-of-rand-float32-and-rand-float64-thread-2/105184/12. Fortunately, I believe the package Distributions.jl already looks into such features

To do:
- [x] rework rounding to stop doing type piracy (currently we overload sqrt, *, /, etc from Base) (closes #486)

To do list (perhaps not necessary for 0.22):
- support `Complex{<:Interval}` (cf. #41)
- improve rounding (cf. #592)
- improve documentation (cf. #524, #591)

### Some thoughts on naming convention

I would like to take this opportunity to review once more the name changes that have occurred since 0.20. One big change is that we decided to not overload Base boolean functions (e.g. `==`, `issubset`, etc.) to prevent silent errors. We have appended the suffix `_interval` to distinguish the functions (e.g. `issubset_interval` vs `issubset` from Base).

Tim Holy's naming convention in [ThickNumbers.jl](https://github.com/timholy/ThickNumbers.jl), is to use the suffix `tn` as a short form of "thick number". Perhaps we want to use something similar to have shorter function names, e.g. the suffix `_itl` (short for "interval") to replace `_interval`.

#### Post-scriptum

Should bisect at any other point than 0.5 work for unbounded intervals?


OlivierHnt:
It seems that there is still more work to be done to have better ForwardDiff.jl compatibility. I have added a few broken tests to keep track of this a little.

For instance, `ForwardDiff.derivative(x -> x^interval(2), interval(-1, 1))` throws a `MethodError` due method ambiguity; which is probably not so bad to fix.

On the other hand, ForwardDiff seems to rely on `T(::Real)`, where `T<:Real`, for conversion. This is something we have discontinued in favour of explicit `convert` call; but perhaps we can re-allow it (with the extra `isguarranteed == false` of course). Perhaps simply

```Julia
Interval{T}(x) where {T<:NumTypes} = convert(Interval{T}, x)
```

lbenet:
Looks again like a giant PR is coming 😄 (or 🥲 ? )... Thanks again for the enourmous effort.

Few comments (note that apriori I do not opose to the changes):

- > changed behaviour of functions in numeric.jl for numbers; e.g. sup(Inf) = NaN (Inf is first converted into an Interval), previous behaviour sup(Inf) == Inf

    Can you explain the motivation? It seems that this is some default method, to be appied to `Real`s. Note in pass that `sup(5.0)` returns `5.0` (not an interval), so I think returning `Inf` is consistent...

- > change default bisection to be at 0.5 by default

    If I recall correctly, the value (127//256) was considered to "break slightly the symmetry" (for symmetric intervals) which in somecases complicates applying IntervalRootFinding methods; in other words, breaking such symmetry permitted faster convergence. (I think this was inspired by a comment in the book by Tucker.)

- > remove atomic since it is an unused internal constructor (see also https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/590#discussion_r1411560983)

    We could also consider keeping `atomic` if it indeed returns an atomic interval (either a thin interval of a one-eps wide interval, except perhaps for unbounded intervals). I'm thinking about the possibility to use it to construct tight intervals, that do contain the actual real number, as `atomic(Float64, "0.1")` does. Issues about performance can be stated in the docs. This is, at the end, also related to rounding.

- A side comment, related to the ITF1788 test suite: I am in favor of having the `rev` functions within IntervalArithmetic; I truly do not see any scenario of using such functions separated from IA...

- I don't understand the to-do:
    > rework rounding to stop doing type piracy (currently we overload sqrt, *, /, etc from Base)

    My point is, rather than type-piracy, we do method-overloading exploiting multiple dispatch. I am in favor of keeping those functions.

- Perhaps the to-do and to-do list should be addressed in a separate PR... makes life easier to review. The question is if the compatibility with ForwardDiff should be also addressed in another PR. Once said this, if `T(x)` is needed by ForwardDiff, your solution seems ok with me. The question is if we can use ForwardDiff in a guaranteed form.

.github/workflows/CI.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.8'
+          - '1.9'
```
lbenet, line 14: We should add 1.10 for testing; current `nightly` is 1.11.0-DEV, I think.

Project.toml
```diff
@@ -5,16 +5,20 @@ version = "0.21.2"
...
+
 [compat]
 CRlibm = "0.7, 0.8, 1"
-EnumX = "1"
+DiffRules = "1"
 FastRounding = "0.2, 0.3"
 RoundingEmulator = "0.2"
 SetRounding = "0.2"
-julia = "1.8"
+julia = "1.9"
```
lbenet, line 24: Do we require Julia >= 1.9? (It's fine, just want to be sure.)

src/intervals/arithmetic/integer.jl
```diff
@@ -2,56 +2,112 @@
...
 
-        function $f(a::Interval)
-            r = $f(bareinterval(a))
-            d = decoration(a)
-            d = min(d, ifelse(isthin(r), d, def))
-            return _unsafe_interval(r, d, isguaranteed(a))
-        end
-    end
+Implement the `sign` function of the IEEE Standard 1788-2015 (Table 9.1).
+"""
```
lbenet, line 10: Is this change motivated by the docstrings? If so, the docs can be included afterwords using `@docs """...""" fn_name`. Seems to me that the code exploiting metaprogramming is more concise.

src/intervals/construction.jl
```diff
@@ -313,186 +294,244 @@ struct Interval{T<:NumTypes} <: Real
...
-        This constructor is *not* compliant with the IEEE Standard 1788-2015.
-        Since misuse of this function can deeply corrupt code, its usage is
-        *strongly discouraged* in favour of [`interval`](@ref).
-    """
     global _unsafe_interval(bareinterval::BareInterval{T}, decoration::Decoration, isguaranteed::Bool) where {T<:NumTypes} =
         new{T}(bareinterval, decoration, isguaranteed)
 end
 
-bareinterval(x::Interval) = x.bareinterval
+"""
```
lbenet, line 301: Maybe you need `@doc` here... (not sure).

src/intervals/construction.jl
```diff
@@ -103,7 +94,25 @@ struct BareInterval{T<:NumTypes}
...
+    **strongly discouraged** in favour of [`bareinterval`](@ref).
+"""
+_unsafe_bareinterval
+
 _normalisezero(a) = ifelse(iszero(a), zero(a), a)
+# used only to construct intervals
+_inf(a::BareInterval) = a.lo
+_sup(a::BareInterval) = a.hi
+_inf(a::Real) = a
+_sup(a::Real) = a
```
lbenet, line 114: Thanks a lot for renaming these functions! I sort of dislike `inf(a)`, because it makes me think on `Inf`...

src/intervals/interval_operations/bisect.jl
```diff
@@ -1,29 +1,43 @@
...
-    r1, r2 = bisect(bx, α)
-    d1, d2 = min(decoration(x), decoration(r1), trv), min(decoration(x), decoration(r2), trv)
+    r₁, r₂ = bisect(bx, α)
+    d₁, d₂ = min(decoration(x), decoration(r₁), trv), min(decoration(x), decoration(r₂), trv)
     t = isguaranteed(x)
-    return (_unsafe_interval(r1, d1, t), _unsafe_interval(r2, d2, t))
+    return (_unsafe_interval(r₁, d₁, t), _unsafe_interval(r₂, d₂, t))
+end
+
+function bisect(x::AbstractVector, i::Integer, α::Real=0.5)
```
lbenet, line 26: Don't we need to specify that the elements of `x` are `Interval`s or `BareInterval`s?

src/intervals/interval_operations/numeric.jl
```diff
@@ -26,7 +26,7 @@ function inf(x::Interval{<:Rational})
     return inf(bareinterval(x))
 end
 
-inf(x::Real) = x
+inf(x::Real) = inf(interval(x))
```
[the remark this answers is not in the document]
lbenet: Same questions for `inf` and `bounds`...

.github/workflows/CI.yml
OlivierHnt: Julia v1.9 is needed to use weak dependencies. We will have to wait a bit longer for 1.10 to be out; at least having nightly working is a good indicator (due to backward compatibility) that it should also pass for 1.10.

LICENSE.md
```diff
@@ -1,5 +1,6 @@
-Copyright (c) David P. Sanders, Luis Benet, and other contributors:
+MIT License
 
+Copyright (c) 2014 David P. Sanders, Luis Benet, and other contributors:
```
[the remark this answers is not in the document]
OlivierHnt: I am bad with copyrights as well but I thought only the initial date was needed, the current year does not need to be (perhaps only if the license changes all the sudden).
Oh I am not sure how to add names; perhaps we can add a line:
```
Copyright (c) 2014 David P. Sanders, Luis Benet
Copyright (c) 2023 Luis Benet, Luca Ferranti, Olivier Hénot and Benoît Richard
[.... license text....]
```

Project.toml
OlivierHnt: Julia v1.9 is needed to use weak dependencies; although not necessary, I think it's ok to add this constraint (especially since 1.10 is around the corner).

docs/src/manual/usage.md
```diff
@@ -103,3 +103,9 @@ issubset_interval(interval(1, 2), interval(2))
 intersect(interval(1, 2), interval(2))
 intersect_interval(interval(1, 2), interval(2))
 ```
+
+
+
+## Custom interval bounds type
+
+A `BareInterval{T}` or `Interval{T}` has the restriction `T <: Union{Rational,AbstractFloat}` which is type for the bounds of the interval. Supposing one wishes to use their own numeric type `MyNumType <: Union{Rational,AbstractFloat}`, they must provide their own arithmetic operations (with correct rounding!).
```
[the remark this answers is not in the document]
OlivierHnt: Thx; done.

src/intervals/construction.jl
```diff
@@ -219,42 +230,7 @@ Base.promote_rule(::Type{BareInterval{T}}, ::Type{BareInterval{S}}) where {T<:Nu
...
-    lo = T(a, RoundDown)
-    hi = T(a, RoundUp)
-    if a == lo
-        lo = prevfloat(lo)
-    end
-    if a == hi
-        hi = nextfloat(hi)
-    end
-    return _unsafe_bareinterval(T, lo, hi)
-end
```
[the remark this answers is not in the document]
OlivierHnt: Sounds good! Done.

src/intervals/interval_operations/bisect.jl
OlivierHnt: No not really, this code will fail properly if that is not the case.

src/intervals/interval_operations/bisect.jl
```diff
@@ -42,3 +56,29 @@ function mince(x::Interval{T}, n::Integer) where {T<:NumTypes}
...
+mince(x::AbstractVector, n::NTuple{N,Integer}) where {N} = mince!(Vector{typeof(x)}(undef, prod(n)), x, n)
+
+"""
+    mince!(v, x, n)
+
+In-place version of [`mince`](@ref).
+"""
+mince!(v::AbstractVector, x::AbstractVector, n::Integer) = mince!(v, x, ntuple(_ -> n, length(x)))
+
+function mince!(v::AbstractVector, x::AbstractVector, n::NTuple{N,Integer}) where {N}
```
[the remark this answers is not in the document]
OlivierHnt: Probably not, it will fail when calling `resize!` probably. Since it is an in-place functionality, one should not use non-mutable array types; this type of signature occurs also in LinearAlgebra.

src/intervals/interval_operations/numeric.jl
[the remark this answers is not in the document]
OlivierHnt: Well I figured that one using `inf`/`sup` on a `Real` would be writing generic code that normally accepts an `Interval`; that is, the `Real` number is though of as a thin number.
So by constructing the interval associated to `x::Real` it is a `conversion` that ensure the output complies, notably with the current flavour. For instance
```Julia
julia> sup(Inf)
┌ Warning: invalid interval, NaI is returned
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/construction.jl:439
NaN
```

OlivierHnt:
Yes 😅, I wanted to push the release to 0.22 as soon as possible. Especially opening this PR since it allows people to check our there code against most of the new features.

Thank for all the feedback!

> Can you explain the motivation? It seems that this is some default method, to be appied to Reals. Note in pass that sup(5.0) returns 5.0 (not an interval), so I think returning Inf is consistent...

Yes it always return a number and not an interval. The conversion to an interval is there to, eg, be consistent with the current flavour. I though that `inf`, and `sup`, acting on a `Real` is the same as thinking about this `Real` as a thin interval.

> If I recall correctly, the value (127//256) was considered to "break slightly the symmetry" (for symmetric intervals) which in somecases complicates applying IntervalRootFinding methods; in other words, breaking such symmetry permitted faster convergence. (I think this was inspired by a comment in the book by Tucker.)

Interesting. I had one brief email exchange about this with @Kolaru. In my opinion, from the perspective of IntervalArithmetic.jl, it is an unexpected choice. Within IntervalRootFinding.jl, the bisection point can be set to be something else that `0.5` (which is probably already the case?).

> A side comment, related to the ITF1788 test suite: I am in favor of having the rev functions within IntervalArithmetic; I truly do not see any scenario of using such functions separated from IA...

I am ok with this; I have no idea what "rev" functions are but we could probably deprecate and archive IntervalContractor.jl and merge it inside IntervalArithmetic.jl.

> My point is, rather than type-piracy, we do method-overloading exploiting multiple dispatch. I am in favor of keeping those functions.

But this is type piracy: we do not own the functions (they come from Base), **nor** the type of the arguments. I agree though that no on else uses these functions this way and that it is unlikely to be an issue, but still... we should avoid type-piracy when possible. Here it is just a matter of defining an interval function, eg `_sin` instead of `Base.sin`.

> Perhaps the to-do and to-do list should be addressed in a separate PR... makes life easier to review. The question is if the compatibility with ForwardDiff should be also addressed in another PR. Once said this, if T(x) is needed by ForwardDiff, your solution seems ok with me. The question is if we can use ForwardDiff in a guaranteed form.

Agreed. There is already a lot of changes; but I do think we should at least tackle the type-piracy issue mentioned above in this PR.
For the other to-do items, they can be added later on as feature enhancements.

I do not know if ForwardDiff.jl satisfy interval arithmetic specifications (and in the mean time, let us not add this conversion. method `T(x)` where `T<:Interval`), but I intend to look deeper into this at some point. This seems to warrant its own PR.

src/intervals/construction.jl
```diff
@@ -209,52 +220,22 @@ bareinterval(a::Tuple) = bareinterval(T, a...)
     return :($res) # set body of the function to return the precomputed result
 end
 
+# atomic constructor
+
+atomic(::Type{T}, x::AbstractString) where {T<:NumTypes} = parse(BareInterval{T}, x)
```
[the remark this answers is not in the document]
lbenet: i mean, have them here

src/intervals/interval_operations/numeric.jl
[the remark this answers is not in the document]
lbenet: Thanks for the clarification.

lbenet:
> Yes 😅, I wanted to push the release to 0.22 as soon as possible. Especially opening this PR since it allows people to check our there code against most of the new features.

I fully agree. Among other, it will make it easier to make the changes in the other packages; I already begun with the transition in TaylorSeries; having the new version tagged would make testing lot easier! 

> > If I recall correctly, the value (127//256) was considered to "break slightly the symmetry" (for symmetric intervals) which in somecases complicates applying IntervalRootFinding methods; in other words, breaking such symmetry permitted faster convergence. (I think this was inspired by a comment in the book by Tucker.)
> 
> Interesting. I had one brief email exchange about this with @Kolaru. In my opinion, from the perspective of IntervalArithmetic.jl, it is an unexpected choice. Within IntervalRootFinding.jl, the bisection point can be set to be something else that `0.5` (which is probably already the case?).

`0.5` is ok with me; I just added the note to give some perspective where it came from. 

> > My point is, rather than type-piracy, we do method-overloading exploiting multiple dispatch. I am in favor of keeping those functions.
> 
> But this is type piracy: we do not own the functions (they come from Base), **nor** the type of the arguments. I agree though that no on else uses these functions this way and that it is unlikely to be an issue, but still... we should avoid type-piracy when possible. Here it is just a matter of defining an interval function, eg `_sin` instead of `Base.sin`.

Perhaps you should be more concrete on this. I thought you were referring to the functions which include some rounding direction, in which case we do own the type, e.g. `Base.sqrt(::IntervalRounding{:fast}, a, ::$RoundingDirection)`. 

> I do not know if ForwardDiff.jl satisfy interval arithmetic specifications (and in the mean time, let us not add this conversion. method `T(x)` where `T<:Interval`), but I intend to look deeper into this at some point. This seems to warrant its own PR.

I agree.

lbenet (commented):
I've sent a bunch of comments that you have already addressed, though i did not check the test-files. LGTM, and I agree to merge it if this is ready (if it is not, I can re-review it later). I do suggest to bump to v0.22.0, merge and tag, so we can begin migrating other packages.

lbenet:
I just found the following:
```julia
julia> b = interval(-1,1)
[-1.0, 1.0]_com

julia> b^3
[0.0, 1.0]_com_NG
```

The result is wrong, perhaps related to `_positive_power_by_squaring`, but it should also be guaranteed!

OlivierHnt:
> I've sent a bunch of comments that you have already addressed, though i did not check the test-files. LGTM, and I agree to merge it if this is ready (if it is not, I can re-review it later). I do suggest to bump to v0.22.0, merge and tag, so we can begin migrating other packages.

Thanks for all this! The only thing I think is important before merging, is to fix the type-piracy issue.
I'll try to find some time today.

lbenet:
> This is compliant with the specifications of the Standard I believe; you may have been looking for the function `pown` (which is distinct from the `pow` function, i.e. `^`)?

You are right, I forgot about `pown` (`nthpow` before this PR). Perhaps it is worth adding a note in the docstrings of `^` about `pown` and/or include docstrings for it.

OlivierHnt:
That's a great idea! It is awfully confusing that `pow` and `pown` give such disparate results 😅.

I will try to find time tonight to add this.

OlivierHnt:
I revived `@interval` and made conversion from `AbstractIrrational` to `Interval` guaranteed.
Closes #595.

mforets:
> With all the progress made recently, we should be a few commits away from releasing 0.22! If everything goes well, this should be the last version before the 1.0 release 🥳!

Nice! Do you have any comments about when/how to approach updating upstream packages in the JuliaIntervals ecosystem to work with < 1.0 versions, or would at the current stage suggest to hold off a bit until the IA 1.0 version is out for such revamp?

lbenet:
I am playing with `TaylorSeries`, `TaylorIntegration` and `IntervalRootFinding` to check any issues, and there are a bunch of them. As @OlivierHnt notes, 0.22 should be the last version before 1.0, so that would be the time to check everything.

OlivierHnt:
Also, some pointers to update the packages from 0.20 to the forthcoming 0.22 version (the 0.21 version of IntervalArithmetic should be skipped):
- `Interval{T}(...)` should be replaced by `interval(T, ...)`
- to use `±`, `..`, `∅`, etc. one must call `using IntervalArithmetic.Symbols`
- this is a list of renamed functions (I hope I did not forget any...):
   - `==` -> `isequal_interval`
   - `issubset`, `⊆` -> `issubset_interval`
   - `isdisjoint` -> `isdisjoint_interval`
   - `isempty` -> `isempty_interval`
   - `isentire` -> `isentire_interval`
   - `intersect`, `∩` -> `intersect_interval`
   - `union` -> removed in favour of `hull`
   - `setdiff` -> `setdiff_interval`
   - `in`, `∈` -> `in_interval`
   - `<`, `>` -> removed in favour of `precedes`, `isweakless` and `isstrictless`
- `IntervalBox` has been entirely removed in favour of using vectors of intervals directly
- `@floatinterval`, `@biginterval` have been removed; `@interval` requires a type, eg `@interval Float64 sin(0.1)` or `@interval Float64 sin(0.1) exp(2)`

Hopefully this helps!

OlivierHnt:
Closes #543 (supersedes PR #544)

lbenet:
Just noticed the following with the `@interval` macro:
```julia
julia> setdisplay(:full); 

julia> @interval Float64 0.1  # this is ok
Interval{Float64}(0.09999999999999999, 0.1, com)

julia> @interval BigFloat 0.1 # rounding did not worked here!
Interval{BigFloat}(0.1000000000000000000000000000000000000000000000000000000000000000000000000000002, 0.1000000000000000000000000000000000000000000000000000000000000000000000000000002, com)
```

OlivierHnt:
As alias: I defined `const ComplexI{T} = Complex{Interval{T}}` and `const RealOrComplexI{T} = Union{Interval{T},Complex{Interval{T}}}`.

If you have better names in mind do not hesitate to suggest them!

OlivierHnt:
@lbenet I have changed the behaviour of `^(x, y)` which dispatches to `pow(x, y)`, unless `y` is a thin integer which then calls `pown(x, sup(y))`.

This function is not included in the IEEE Std for interval arithmetic, but I agree that it is more intuitive.

OlivierHnt:
I have added a configuration mechanism for powers (supersedes PR #388) which changes the behaviour of `^`:

```Julia
julia> using BenchmarkTools

julia> a = interval(1, 2)
[1.0, 2.0]_com

julia> b = interval(0.25, 1)
[0.25, 1.0]_com

julia> IntervalArithmetic.power_mode() = IntervalArithmetic.PowerMode{:slow}() # default, use `pow` and `pown`

julia> @btime $a^$b
  5.056 μs (36 allocations: 1.27 KiB)
Interval{Float64}(1.0, 2.0, com)

julia> IntervalArithmetic.power_mode() = IntervalArithmetic.PowerMode{:fast}() # use `fastpow`

julia> @btime $a^$b;
  109.808 ns (0 allocations: 0 bytes)
Interval{Float64}(1.0, 2.0000000000000004, com)
```

I would be inclined to set `IntervalArithmetic.PowerMode{:fast}` as the default since the "slow" option is embarrassingly slow; but note that this setting is not compliant with the ITF1788 test suite.

lbenet:
What should we do about the following:
```julia
julia> a = interval(-1, 1)
[-1.0, 1.0]_com

julia> b = interval(-1, 1)*1.0
[-1.0, 1.0]_com_NG

julia> isequal_interval(a, b)
true
```

OlivierHnt:
It's a good question.

I decided to neglect the `isguaranteed` field, since we also neglect the decoration (this is asked by the standard).

```Julia
julia> isequal_interval(interval(1, 2, trv), interval(1, 2, com))
true
```

Maybe keeping it this way is not so bad, since `isguaranteed` being false will contaminate every intervals it touches. Also perhaps returning false might lead to "unexepected" errors? Eg used in `if ... else ...`, it will end up picking up a different branch, distracting the user from focusing on "debugging" `isguaranteed == false`.

P.S.: I checked the third box in https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/593#issuecomment-1850452669; but uncheck it if you have further concerns/comments on the `^` function.

OlivierHnt:
@lbenet I took a quick Look at TaylorSeries.jl. It does seem (but I could have missed some subtleties) that the code can be refactored by defining your own comparison operators: eg `_ts_isequal(x, y) = ==(x, y)`, `_ts_isless(x, y) = <(x, y)`, etc.
Then in TaylorSeriesIAExt.jl you can have the first lines read `_ts_isequal(x::Interval, y::Interval) = isequal_interval(x, y)`, `_ts_isless(x, y) = strictprecedes(x, y)` (or `isweakless`, or something else), etc.

Hopefully this also help reduce the size of TaylorSeriesIAExt.jl.

OlivierHnt:
Ouch I stumbled across #374 ... this means that we do not have rigorous numerics for the :accurate mode (renamed :fast in this PR)...
Since the number of ulp also depends on the architecture, this is not a "quick" fix. So for the time being I will just comment out all the `:accurate` rounding functions.

lbenet:
> @lbenet I have changed the behaviour of `^(x, y)` which dispatches to `pow(x, y)`, unless `y` is a thin integer which then calls `pown(x, sup(y))`.
> 
> This function is not included in the IEEE Std for interval arithmetic, but I agree that it is more intuitive.

Thanks a lot! I truly enjoyed it (after a rather painful experience)...

> I would be inclined to set IntervalArithmetic.PowerMode{:fast} as the default since the "slow" option is embarrassingly slow; but note that this setting is not compliant with the ITF1788 test suite.

Regarding this, I guess we can run those tests with `PowerMode{:slow}`, to show that they pass.

lbenet:
> I decided to neglect the `isguaranteed` field, since we also neglect the decoration unless it is `ill` (this is asked by the standard).

I agree with your solution. I didn't remember that equality didn't involve the decorations, so your solution is consistent with that.

lbenet:
> @lbenet I took a quick look at TaylorSeries.jl. It does seem (but I could have missed some subtleties) that the code can be refactored by defining your own comparison operators: eg _ts_isequal(x, y) = ==(x, y), _ts_isless(x, y) = <(x, y), etc.
Then in TaylorSeriesIAExt.jl you can have the first lines read _ts_isequal(x::Interval, y::Interval) = isequal_interval(x, y), _ts_isless(x::Interval, y::Interval) = strictprecedes(x, y) (or isweakless, or something else), etc.
>
>Hopefully this also help reduce the size of TaylorSeriesIAExt.jl.

It indeed can be nicely refactored! I haven't pushed the branch where i am playing with all that, but it does work. So far, I have only used `_isthinzero(x) = iszero(x)` for some functions, which when extended to intervals use the method `_isthinzero(x::Interval) = isthinzero(x)`. Also, I love how`ComplexI` allows to use metaprogramming which reduces the code a lot!

incidentally, do you know what is better, to have separate methods for `Interval` and ComplexI`, or a single method for `RealOrComplexI`?

Regarding TaylorSeries, I am a bit now checking the internal methods for the functions, to try to avoid the propagation of `NG`s... and also increasing the coverages of the tests. Complex intervals are an issue, but this should be treated separately.

OlivierHnt:
> incidentally, do you know what is better, to have separate methods for Interval and ComplexI, or a single method for RealOrComplexI?

Well my hope is that most of the time one can just use `RealOrComplexI` since a bunch of functionalities works for both `Interval` and `Complex{<:Interval}` seamlessly.

> Regarding TaylorSeries, I am a bit now checking the internal methods for the functions, to try to avoid the propagation of NGs... and also increasing the coverages of the tests. Complex intervals are an issue, but this should be treated separately.

Is it because of things like `x + im*y` where `x = y = interval(1)` for instance? The fix is to use `complex(x, y)`... but it is a little annoying. Unfortunately (and unlike `Irrational`) conversion from `Number` to `Bool` are allowed (eg `Bool(1.)`) so in principle one can produce a `Bool` from non-rigorous arithmetic...

src/intervals/arithmetic/power.jl
```diff
@@ -1,412 +1,440 @@
...
-function _pow(a::BareInterval{BigFloat}, x::BigFloat)
-    domain = _unsafe_bareinterval(BigFloat, zero(BigFloat), typemax(BigFloat))
+# helper functions for power
 
-    if isthinzero(a)
-        x > 0 && return zero(BareInterval{BigFloat})
-        return emptyinterval(BareInterval{BigFloat})
-    end
+_select_pow(::PowerMode{:fast}, x, y) = fastpow(x, y)
+_select_pown(::PowerMode{:fast}, x, y) = fastpown(x, y)
```
lbenet, line 71: According to the docstrings (above), this should correspond to `pown(x, y)`, isn't it?

OlivierHnt: Ah the docstring is wrong; I figured we should have a `fastpown` function as well. I will fix it.

OlivierHnt:
Mhm `@interval` has very poor performance; this is due to `parse`, e.g.:

```Julia
julia> @time IntervalArithmetic.atomic(Float64, 1)
  0.000029 seconds (14 allocations: 644 bytes)
[1.0, 1.0]_com
```

test/test_ITF1788/LICENSE.md
lbenet, line 1: Not that I disagree deleting this file, but don't we need it?
cc @lucaferranti

OlivierHnt: Oh good catch; it was in the folder with the generated tests. I should add it back to the non-generated test suite folder called "itl".

lbenet:
Just noticed the following (which relies on Base methods!):
```julia
julia> a = interval(-1, 1)
Interval{Float64}(-1.0, 1.0, com)

julia> b = complex(a, a)
Interval{Float64}(-1.0, 1.0, com) + Interval{Float64}(-1.0, 1.0, com)im

julia> zero(a)  # OK
Interval{Float64}(0.0, 0.0, com)

julia> zero(b)  # Result is ok, except perhaps for the NG (due to using a method from Base)
Interval{Float64}(0.0, 0.0, com, NG) + Interval{Float64}(0.0, 0.0, com, NG)im

julia> @which zero(b)
zero(x::Number)
     @ Base number.jl:308
```

Maybe we should open a new issue with the most basic operations we want to have for `ComplexI`... (Truly, I adore that alias!)

OlivierHnt:
I added a specialised method for `zero` and `one`.

We should perhaps merge this PR and release 0.22? This will allow us getting more feedback from downstream packages during December.

lbenet:
Please go ahead! I fully agree merging this PR and releasing v0.22 will allow us to check everything more reliably. Great work!

PR #594, opened 2023-12-06
Arithmetic operations with integers and intervals

lbenet:
Operations with integers should be promoted, since there are no issues with rounding,. I've hand-coded this operations to avoiding promotion.

### DIFF
--- a/src/intervals/arithmetic/basic.jl
+++ b/src/intervals/arithmetic/basic.jl
@@ -193,6 +193,16 @@ Implement the rational division; this is semantically equivalent to `a / b`.
 
 //(a::Interval, b::Interval) = /(a, b)
 
+# Arithmetic operations with intervals yield guaranteed results
+for (T, fc) in ((:BareInterval, :(bareinterval)), (:Interval, :(interval)))
+    for op in (:+, :-, :*, :/, ://)
+        @eval begin
+            $(op)(n::Integer, x::$T) = $(op)($fc(n), x)
+            $(op)(x::$T, n::Integer) = $(op)(x, $fc(n))
+        end
+    end
+end
+
 """
     muladd(a::BareInterval, b::BareInterval c::BareInterval)
     muladd(a::Interval, b::Interval c::Interval)
--- a/test/interval_tests/display.jl
+++ b/test/interval_tests/display.jl
@@ -54,7 +54,7 @@ setprecision(BigFloat, 256) do
 
     @testset "Interval" begin
         a = interval(1, 2)
-        a_NG = a/1
+        a_G = a/1
         b = interval(-floatmin(Float64), 1.3)
         b32 = interval(-floatmin(Float32), parse(Float32, "1.3"))
         b16 = interval(-floatmin(Float16), parse(Float16, "1.3"))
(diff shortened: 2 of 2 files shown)


lbenet:
More tests are perhaps needed...

lbenet:
> Aren't there issues with rounding if the integer is too big to represent exactly as a float, ie larger than 2^52 (?)

Good point! Indeed, that creates a problem with rounding....

OlivierHnt:
I am very reluctant about this change; it is not a guaranteed operation. Integers can be produced in various ways which do not comply with interval arithmetic specifications.

Again, the `isguarantee` field is not about correct rounding, but about the origin of the operands.

Hence, the only thing that is debatable are for **literal** numbers (whether they are floats or integers or any other types). The problem is that it is not clear how to identify them. In fact, I am not even sure it can be distinguished from constant folded (from Julia's compiler) numbers...

lbenet:
> I am very reluctant about this change; it is not a guaranteed operation. Integers can be produced in various ways which do not comply with interval arithmetic specifications.

I agree with you that the operation is not guaranteed in general, certainly not for for large integers (`Int`s) as David notes, but for small ones I think it should work. The following is a non-performant way to check that the "promotion" of `Int`s to intervals is guaranteed (at least for literal integers):

```julia
julia> setdisplay(:full);

julia> IntervalArithmetic.atomic(Float64,string(2.0^54)) # no need to round
BareInterval{Float64}(1.8014398509481984e16, 1.8014398509481984e16)

julia> isthin(ans)
true

julia> IntervalArithmetic.atomic(Float64,string(2.0^54+6)) # rounding is needed!
BareInterval{Float64}(1.8014398509481988e16, 1.801439850948199e16)

julia> isthin(ans)
false
```

(I think 2^54+6 is a the first `Int` that needs rounding when promoted to `Float64`, but that should be properly computed.)

Why do I care about this: essentially, many algorithms (in TaylorSeries) involve things like `d = 2*c`, which would lead to the "NG" tag, though it should definitely work. One could certainly write specialized methods with `interval(2)` instead, but that seems to involve repeating too much code for such a "tiny" change.

> Again, the `isguarantee` field is not about correct rounding, but about the origin of the operands.

I agree, though I think that "origin of operands" at the end refers to the (finite) representation in some float format, which may involve rounding: `interval(2,4)*0.5` is exacly `interval(1,2)`, but replacing 0.5 by 0.1 requires rounding or the NG flag; dividing by `interval(2)` or `interval(10)` should yield a rigorous result. For these I do agree to have the NG flag when the "literals" are `Float64`. But for small enough integers, I think there should be no issue about rounding.

> Hence, the only thing that is debatable are for **literal** numbers (whether they are floats or integers or any other types). The problem is that it is not clear how to identify them. In fact, I am not even sure they can be distinguished from constant folded (from Julia's compiler) numbers...

I think the code above shows the point of David: inspite of passing a literal `Int`, once the `(Bare)Interval{Float64}` is created, there may be issues with rounding. 

The point to debate is whether we allow the user to write code like `d = 2*c` and expect no `NG` flag (due to the `2`, but get it if the literal is too large), or we force him to write `interval(2)`; in my opinion, we should opt for the second one.

PR #598, opened 2023-12-16
Update docs link in README.md

vtjeng:
I'm not sure why the release process isn't producing any docs at `stable` (or, in fact, for any previous versions). 

In the meantime, it seems better to link users to the dev docs rather than a dead link.

FWIW, I had a similar issue in my package some time ago (https://github.com/vtjeng/MIPVerify.jl/pull/116) but it seems to have to do with me using an old version of `Documenter`, which is not the issue you are having here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 [![DOI](https://zenodo.org/badge/87007945.svg)](https://zenodo.org/badge/latestdoi/87007945)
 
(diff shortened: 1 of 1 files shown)


OlivierHnt:
Indeed I noticed this (I realised it had been broken for years in fact) and I did some things that I hope will produce docs for the next stable release (some v0.22.1). But I did not want to tag the release just for the docs, since I am not sure if what I did will work 😅.

Anyways, you are right about pointing to the dev branch of the docs! Thx!

PR #600, opened 2023-12-17
Improve accuracy of `sinpi` and `cospi`

### LINKED ISSUE #381: Use version of sin with quadrant for Float32 too

### LINKED ISSUE #412: `sinpi` returns unnecessarily wide intervals

OlivierHnt:
This PR is essentially a revival of the PR #421 which got forgotten unfortunately.
Closes #381; closes #412.

This PR also refactors the code in trigonometric.jl; the code is about 200 lines shorter.

### DIFF
--- a/src/intervals/arithmetic/trigonometric.jl
+++ b/src/intervals/arithmetic/trigonometric.jl
@@ -185,7 +185,7 @@ function Base.cospi(x::BareInterval{T}) where {T<:NumTypes}
 
     lo, hi = bounds(x)
 
-    isthin(x) & isinteger(2lo) && return zero(BareInterval{T}) # by-pass rounding to improve accuracy for 32 bit systems
+    isthin(x) & !isinteger(lo) & isinteger(2lo) && return zero(BareInterval{T}) # by-pass rounding to improve accuracy for 32 bit systems
 
     lo_quadrant = _quadrantpi(lo)
     hi_quadrant = _quadrantpi(hi)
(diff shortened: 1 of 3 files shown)


OlivierHnt:
So it seems that the stack overflow comes from the fact that `sinpi`, and `cospi` are not properly defined for 32 bit system in [CRlibm.jl](https://github.com/JuliaIntervals/CRlibm.jl/tree/master).

Specifically, https://github.com/JuliaIntervals/CRlibm.jl/blob/d212ba9154ade995bb6c330c51092c685a8ec432/src/CRlibm.jl#L152-L155 do not include the functions `sinpi`, `cospi`,  `tanpi` and `atanpi`.

I may be missing something here, but why were these functions omitted? At first glance, I would  have defined [`function_names`](https://github.com/JuliaIntervals/CRlibm.jl/blob/d212ba9154ade995bb6c330c51092c685a8ec432/src/CRlibm.jl#L137-L140) and [`MPFR_function_names `](https://github.com/JuliaIntervals/CRlibm.jl/blob/d212ba9154ade995bb6c330c51092c685a8ec432/src/CRlibm.jl#L152-L155) to be the same.

cc @dpsanders

OlivierHnt:
I fixed the StackOverflow issue for `cospi` and `sinpi` in 32 bit systems; I guess I was mislead into thinking that `cospi` and `sinpi` were defined in MPFR.

PR #608, opened 2024-01-04
Add RecipesBase as an extension

### LINKED ISSUE #605: Plotting intervals after removal of `IntervalBox`
I have often made use of `IntervalBox` for plotting enclosures of functions. With the removal of `IntervalBox` in 0.22 this is no longer possible. Is there some other convenient way of plotting interval enclosures? In the [documentation](https://juliaintervals.github.io/pages/tutorials/tutorialArithmetic/) at juliaintervals.github.io is still makes use of `IntervalBox`, but I believe that is outdated?

The removal of `IntervalBox` seems to be missing from the [NEWS.md](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/master/NEWS.md) file as well. As a side note it would be great to link to the NEWS.md file in the releases! It took me some time before I found it.

OlivierHnt:
Closes #605.

In the future we should add recipes like this for Makie.jl.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 # News
 
-## 0.22.2
+## 0.22.4
+  - no longer generate the ITF1788 test suite automatically (cf. [#606](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/606))
+  - no longer write a log for the warnings of the tests
+
+## 0.22.3
 
 - add `sind`, `cosd` and `sincosd`
 - pointwise boolean functions are re-introduced (cf. PR [#604](https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/604)), namely: `iszero`, `isone`, `isinteger`, and `==(::Interval, ::Number)`/`==(::Number, ::Interval)`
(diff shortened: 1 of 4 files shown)


lorenzgillner:
Please note that there is a typo in `ext/IntervalArithmeticRecipesBaseExt.jl`: in lines 5 and 19 `AbstractVector` is spelled incorrectly, resulting in an error when plotting boxes.

OlivierHnt:
🤦‍♂️ my bad

Joel-Dahne:
Looks good! Only comment I have is that it would maybe make sense to also support `BareInterval`. But I haven't really understood the new role of `BareInterval`, so it might not make sense.

OlivierHnt:
Indeed, that crossed my mind. But I figured, it is better to be a bit conservative, at least for the time being.

An important aspect of the 0.22 versions is that one should always use `Interval` and never `BareInterval`; except perhaps in performance critical code, where the user has also a clear understanding of the operations taking place to make sense of the output.
E.g.
```Julia
julia> sqrt(bareinterval(-1, 1))
[0.0, 1.0]

julia> sqrt(interval(-1, 1))
[0.0, 1.0]_trv
```

PR #609, opened 2024-01-05
Properly closed interval boxes for other backends than GR

lorenzgillner:
When plotting an interval box with `Plots.jl` and the `GR` backend, the produced rectangle has borders on all four sides. However, when using other backends like `UnicodePlots`, `PGFPlotsX` or `PythonPlots`, the left border is missing (left side in the image below). This patch resolves the issue (right side):

unicodeplots_interval


OlivierHnt:
Oh nice, thx!

PR #610, opened 2024-01-05
(re)Introduce scaling factor for mid

Kolaru:
I am looking into updating IntervalRootFinding.jl for v0.22.

One think that disappeared that we need is `scaled_mid`.

This PR reintroduces it by giving `mid` an optional argument.

Technically it makes the default `mid` a bit slower (4 flops and 1 division, instead of 2 flops). I can split the default if it is somehow very critical for someone.

The new formula looks weird, but that's what we need to pass the test suite.

### DIFF
--- a/src/intervals/interval_operations/numeric.jl
+++ b/src/intervals/interval_operations/numeric.jl
@@ -84,43 +84,58 @@ Implement the `mid` function of the IEEE Standard 1788-2015 (Table 9.2).
 
 See also: [`inf`](@ref), [`sup`](@ref), [`bounds`](@ref), [`diam`](@ref),
 [`radius`](@ref) and [`midradius`](@ref).
+
+
+    mid(x, α)
+
+Relative midpoint of `x`, for `α` between 0 and 1.
+
+`mid(x, 0)` is the lower bound of the interval, `mid(x, 1)` the upper bound,
+and `mid(x, 0.5)` the midpoint.
 """
-function mid(x::BareInterval{T}) where {T<:AbstractFloat}
+function mid(x::BareInterval{T}, α = 0.5) where {T<:AbstractFloat}
     isempty_interval(x) && return convert(T, NaN)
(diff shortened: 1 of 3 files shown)


src/intervals/interval_operations/numeric.jl
```diff
@@ -84,43 +84,58 @@ Implement the `mid` function of the IEEE Standard 1788-2015 (Table 9.2).
...
+    if isentire_interval(x)
+        α == 0.5 && return zero(T)
+        α > 0.5 && return prevfloat(typemax(T))
+        return nextfloat(typemin(T))
+    else
+        lo, hi = bounds(x)
+        lo == typemin(T) && return nextfloat(lo) # cf. Section 12.12.8
+        hi == typemax(T) && return prevfloat(hi) # cf. Section 12.12.8
+        β = convert(T, α)
+        midpoint = β * (hi + lo * (1/β - 1)) # Exactly 0.5 * (hi + lo) for β = 0.5
```
OlivierHnt, line 107: Why not `midpoint = β * (hi - lo) + lo` (which is what is used for the case of rational interval bounds)?

OlivierHnt:
That's a good idea.

There probably were tests for `scaled_mid` that we can revive.

src/intervals/interval_operations/numeric.jl
Kolaru: For some small floats you get a slightly different results, and the ITF suite doesn't pass. Namely:

```julia
minimal_mid_test: Test Failed at C:\Users\Kolaru\.julia\dev\IntervalArithmetic\test\ITF1788_tests\libieeep1788_num.jl:153
  Expression: mid(bareinterval(-5.0e-324, 1.0e-323)) === 0.0
   Evaluated: 5.0e-324 === 0.0
```

OlivierHnt: I see

Issue #295, opened 2019-06-25
Implement better algorithms for complex interval division (and multiplication?)


dpsanders:
https://dl.acm.org/citation.cfm?id=362563&dl=ACM&coll=DL

dpsanders:
cc @yashrajgupta

dpsanders:
https://ieeexplore.ieee.org/document/6158963

Issue #41, opened 2017-05-27
Broken functions for complex intervals


dpsanders:
The following functions are currently broken for complex intervals; see https://github.com/JuliaLang/julia/issues/22095.
They give a stack overflow due to incorrect fallbacks in base Julia:
- [x] `exp`, `sin`, `cos`, `cosh`, `sinh` -- due to `sincos`
- [ ] `sqrt`, `log`, `log1p`, `log2`, `log10`
- [ ] `asin`, `acos`, `atan`
- [ ]  `tanh`, `asinh`, `acosh`, `atanh`
- [ ] `sinpi`, `cospi`, `sinc`, `cosc`

giordano:
I'm sure you already carefully thought about typing of `AbstractInterval`, but what about making it a subtype of `AbstractFloat`?  Are there unwanted side effects?

I did so for Measurement type in [`Measurements.jl`](https://github.com/giordano/Measurements.jl) package, and all complex functions work with it, without any effort on my side.

dpsanders:
Hmm, I haven't tried it; that's good idea, thanks! 

It's not actually very clear to me what that would imply though.

dpsanders:
See the branch `subtype_abstractfloat`. It currently leads to some ambiguity warnings in `convert` methods that will have to be fixed first.

giordano:
> It's not actually very clear to me what that would imply though.

For the complex functions, they would automatically use the `f(::Complex{<:AbstractFloat})` method, instead of going through `f(x::Complex{<:Real}) = f(float(x))`, that triggers the StackOverflowError.

See https://github.com/giordano/Measurements.jl/issues/1 for the discussion about typing of `Measurement`, I first started with `Number` and after discussion I eventually realized that `AbstractType` was the most useful and meaningful supertype of my type (see last message).

giordano:
However I see that `Interval{Rational}` is allowed, in that case may not be a good idea to make `AbstractInterval` a subtype of `AbstractFloat`.  Out of curiosity: why `Rational` and not `Integer`?

dpsanders:
I am not that convinced that we still need `Interval{Rational}`; cc @lbenet.

`Interval{Int}` is very restrictive: as soon as you do almost anything to it, you leave the integers.

giordano:
> `Interval{Int}` is very restrictive: as soon as you do almost anything to it, you leave the integers.

[...]

PR #611, opened 2024-01-08
Construction of interval when mixing real and complex arguments

Joel-Dahne:
When constructing an interval from one real and one complex argument I expected it to treat the imaginary part of the real argument as zero. What happens is instead that it takes the imaginary part of the interval to only depend on the complex argument, e.g.
```julia
julia> interval(0, complex(1, 1))
[0.0, 1.0]_com + ([1.0, 1.0]_com)im
```
Compare this to what we get if we make the first argument complex
```julia
julia> interval(complex(0), complex(1, 1))
[0.0, 1.0]_com + ([0.0, 1.0]_com)im
```

This behavior could of course be by design, but since there are no tests for it I would guess that it is just a mistake. Since we have `isequal(0, complex(0))` it seems natural that they would both give the same result.

This PR changes it so that the real argument is treated as a complex argument with imaginary part zero. So we would get identical results for the two above constructors. Feel free to discard this if the original behavior is indeed intended.

I also noticed that there are no constructors for `BareInterval` from `Complex`. This would maybe be by design though? Since in general `BareInterval` is intended to be very "bare".

### DIFF
--- a/src/intervals/construction.jl
+++ b/src/intervals/construction.jl
@@ -481,9 +481,9 @@ end
 _interval_infsup(::Type{T}, a::Complex, b::Complex, d::Decoration = com) where {T<:NumTypes} =
     complex(_interval_infsup(T, real(a), real(b), d), _interval_infsup(T, imag(a), imag(b), d))
 _interval_infsup(::Type{T}, a::Complex, b, d::Decoration = com) where {T<:NumTypes} =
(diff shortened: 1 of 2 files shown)


OlivierHnt:
Good catch, thx for opening this PR 🙂
We are still lacking a thorough complex support (see e.g. #41 and #295).

LGTM

Joel-Dahne:
Constructing them is a good first step at least! :smiley:

PR #614, opened 2024-01-15
intervals: pass precision to `BigFloat` instead of changing globally

nsajko:
Taking the lock is still necessary after this change. It'd be necessary to audit all `BigFloat` methods to prove otherwise. Also see https://github.com/JuliaLang/julia/pull/49749, which fixes some of them in Julia `Base`.


OlivierHnt:
Thx for giving a shot at fixing these issues. I went a different way in PR #615 by completely removing `_bigequiv` and the precision lock.

PR #617, opened 2024-01-18
Ensure correct rounding for irrationals

### LINKED ISSUE #616: Incorrect rounding for extreme irrationals
See [52941](https://github.com/JuliaLang/julia/issues/52941) and my comment in #615.

The rounding of irrationals in Base Julia is not fully rigorous and can give incorrect results in extreme cases. This has as a consequence that intervals constructed from irrationals are not guaranteed to enclose them. For example, defining the irrational $1 - \pi^{-300}$ gives us
```julia
julia> Base.@irrational almost_one (1 - big(π)^-300)

julia> almost_one
almost_one = 1.0...

julia> Float64(almost_one, RoundDown)
1.0

julia> interval(almost_one)
[1.0, 1.0]_com

```
where the interval clearly doesn't contain the irrational.

For Base Julia these types of extreme cases are maybe fine to not get fully correct. For interval arithmetic not so much. I would propose to limit the construction of intervals from irrationals to only those irrationals which are implemented in `mpfr`, which are `Irrational{:π}`, `Irrational{:γ}` and `Irrational{:catalan}`. One could also support `Irrational{:ℯ}`, but one has to call `mpfr` directly since the regular constructor doesn't expose the rounding parameter.

Note that the constructors in `src/constructors.jl` under the comment 
```julia
# by-pass the absence of `BigFloat(..., ROUNDING_MODE)` (cf. base/irrationals.jl)
# for some irrationals defined in MathConstants (cf. base/mathconstants.jl)
```
are not safe, as by the above the rounding to `Float64` is not safe.

OlivierHnt:
Fixes #616.

### DIFF
--- a/src/intervals/construction.jl
+++ b/src/intervals/construction.jl
@@ -83,14 +83,14 @@ struct BareInterval{T<:NumTypes}
 
     # need explicit signatures to avoid method ambiguities
 
-    global _unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Integer,T<:Rational{S}} =
+    global __unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Integer,T<:Rational{S}} =
         new{T}(_normalisezero(a), _normalisezero(b))
-    _unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Union{Int8,UInt8},T<:Rational{S}} =
+    __unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Union{Int8,UInt8},T<:Rational{S}} =
         new{T}(_normalisezero(a), _normalisezero(b))
-    _unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Union{Int16,UInt16},T<:Rational{S}} =
+    __unsafe_bareinterval(::Type{T}, a::T, b::T) where {S<:Union{Int16,UInt16},T<:Rational{S}} =
         new{T}(_normalisezero(a), _normalisezero(b))
 
-    _unsafe_bareinterval(::Type{T}, a::T, b::T) where {T<:AbstractFloat} =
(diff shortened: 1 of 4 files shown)


Joel-Dahne:
I took a quick look at the code and didn't spot anything odd. I don't understand the distinction between `_unsafe_bareinterval` and `__unsafe_bareinterval` though?

You could also support `Irrational{:ℯ}` by simply evaluating `exp(BareInterval{T}(1))` and taking those endpoints. 

It would of course be possible to support more irrationals by simply evaluating them with interval arithmetic. They would in general not be correctly (but faithfully) rounded in those cases though. I don't know what guarantees you give on correct rounding outside of the standard though (and the standard clearly doesn't covers these types of irrationals).

OlivierHnt:
> I took a quick look at the code and didn't spot anything odd. I don't understand the distinction between _unsafe_bareinterval and __unsafe_bareinterval though?

Oh just a way to fix method ambiguities when we introduce `Irrational` in the functions type signature.

> You could also support Irrational{:ℯ} by simply evaluating exp(BareInterval{T}(1)) and taking those endpoints.
>
> It would of course be possible to support more irrationals by simply evaluating them with interval arithmetic. They would in general not be correctly (but faithfully) rounded in those cases though. I don't know what guarantees you give on correct rounding outside of the standard though (and the standard clearly doesn't covers these types of irrationals).

Indeed, same for `:φ`. Well for such cases, I think it suffices for the returned interval to contain the true value without necessarily be tight.

Joel-Dahne:
For `:ℯ` it would even be tight!

I think non-tight enclosures would be perfectly fine for irrationals. Personally I don't care much for correct rounding (faithful rounding is of course necessary though!), you lose it after doing two operations anyway.

In that case I think it would make sense to also implement `:φ` in this PR! Then all irrationals in Base are covered at least.

OlivierHnt:
Sounds good, I'll find time to make the changes. Thx for looking over this PR btw!

PR #619, opened 2024-01-18
Avoid `convert` in `^` for literal powers

### LINKED ISSUE #618: Avoid `convert` to `Interval` in `^`
In current master:
```julia

julia> a = interval(1,2)
[1.0, 2.0]_com

julia> a^4
[1.0, 16.0]_com_NG
```
The *non-guaranteed* flag is disturbing. The problem arises from [these lines](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/946493a03a1c4535334956302ec7a2c9807e0d59/src/intervals/arithmetic/power.jl#L63C1-L66C74), which use `convert`. This, incidentally, does not respects the `PowerMode` in use.

lbenet:
Fix #618

### DIFF
--- a/src/intervals/arithmetic/power.jl
+++ b/src/intervals/arithmetic/power.jl
@@ -66,7 +66,8 @@ Base.:^(x::Rational, y::Interval) = ^(convert(Interval{typeof(x)}, x), y)
 Base.:^(x::Interval, y::Rational) = ^(x, convert(Interval{typeof(y)}, y))
 
 # overwrite behaviour for small integer powers from https://github.com/JuliaLang/julia/pull/24240
-Base.literal_pow(::typeof(^), x::Interval, ::Val{n}) where {n} = x^n
+# Base.literal_pow(::typeof(^), x::Interval, ::Val{n}) where {n} = x^n
+Base.literal_pow(::typeof(^), x::Interval, ::Val{n}) where {n} = _select_pown(power_mode(), x, n)
 
 # helper functions for power
 
--- a/test/interval_tests/power.jl
+++ b/test/interval_tests/power.jl
@@ -90,3 +90,14 @@ end
 #     @test real(b) == interval(-13.12878308146216, -13.128783081462153)
 #     @test imag(b) == interval(-15.200784463067956, -15.20078446306795)
 # end
+
+@testset "Literal powers" begin
+    x = interval(-1,1)
+    n = 2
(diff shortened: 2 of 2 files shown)


lbenet:
This is related to the test that fails:
```julia
julia> f(x)  = interval(2)^x

julia> f′(x) = log(interval(2)) * f(x)
f′ (generic function with 1 method)

julia> df(t) = ForwardDiff.derivative(f, t)
df (generic function with 1 method)

julia> f′(0) === df(0)
false

# But
julia> isequal_interval(f′(0), df(0))
true
```

The subtlety in `===` is that `df(0)` is `NG` (because of `ForwardDiff`).

lbenet:
I will need to add tests. Perhaps it is also good time to increase the tests related to all the power functions and power modes.

lbenet:
Followed @OlivierHnt suggestion [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/618#issuecomment-1900516668); it seems to work nicely, distinguishing literal powers from whatever. (All previous changes have been reverted.)

OlivierHnt:
@Kolaru any thoughts?

lbenet:
> Note (just as a memo): while `Val{n}` can have `n` to be a float or a bool, the call to `Base.literal_pow` only occurs when `n` is a literal integer which makes this implementation correct. Also, the issue is not entirely fixed for large literal integers:
> 
> ```julia
> julia> interval(1)^2305843009213693951
> [1.0, 1.0]_com
> 
> julia> interval(1)^2305843009213693952 # any integer larger than this one will also have the NG flag
> [1.0, 1.0]_com_NG
> ```
> 
> Whether or not this is a bug, it is on Julia's side.

Good catch! I will check if I can find what's going on, but I guess this is some sort of rounding issue (due perhaps to the conversion to Float64), as commented [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/618#issuecomment-1912617964).

Kolaru:
For me it can be merged once the tests are there :)

OlivierHnt:
At the end of test/interval_tests/power.jl, we can just add something like

```Julia
@testset "Literal powers" begin
    x = interval(1)
    n = 2
    @test isguaranteed(x ^ 2)
    @test !isguaranteed(x ^ n)
    @test_broken isguaranteed(x ^ 2.0)
    @test_broken isguaranteed(x ^ 2305843009213693952)
end
```

Issue #612, opened 2024-01-09
Non-thread safe use of `setrounding`


Joel-Dahne:
The slow versions of the `_round` methods in `src/interval/rounding.jl` use `BigFloat` together with `setrounding` in a way that is not thread-safe. For `setprecision` there is a `precision_lock` defined at the top of the file that is used to make it thread-safe, but not such look is used for the `setrounding`. Starting Julia with two threads I could reliably produce errors with the following code
```julia
x = interval(BigFloat, 1)
y = sin(x)
Threads.@threads for _ in 1:1000
    z = sin(x)
    @assert isequal_interval(y, z)
end
```
In many cases (though not all) I also found that after running this loop even the non-threaded behavior is wrong. Most of the time this fails:
```julia
setrounding(BigFloat, RoundNearest) # Reset rounding in case this would be an issue
z = sin(x)
@assert isequal_interval(y, z)
```
So there is some global state that gets messed up.

I made a quick attempt at fixing this by adding locks around `setround`, but I could still not get it to work properly. So there might be something else that I'm missing as well.

Joel-Dahne:
Some more details, maybe it helps someone with the debugging...

I can reproduce it also with the code
```julia
x = bareinterval(BigFloat, 1)
y = sin(x)
Threads.@threads for _ in 1:100000
    z = IntervalArithmetic._unsafe_bareinterval(BigFloat, IntervalArithmetic._sin_round(x.lo, RoundDown), IntervalArithmetic._sin_round(x.hi, RoundUp))
    if !isequal_interval(y, z)
        @show y z
    end
end
```
BUT, for this code my attempt at adding locks for the use of `setrounding` seems to work. I don't know why this is the case, as far as I understand the codes should do more or less the same.

dpsanders:
Ugh, `setrounding` has always been a mess, and `BigFloat`s have always had too much global state.

I suggest using the `accurate` mode (or whatever it is now called) that just uses `prevfloat` and `nextfloat` for rounding.

dpsanders:
But of course it would be great to get this fixed too.

Joel-Dahne:
Thinking more about it, it would not be enough to use locks anyway since any other part of the program could still change the rounding mode. The same holds true for the `precision_lock` that guards the `setprecision`.

I think the simplest solution is to just call MPFR directly and not use the `BigFloat` interface. So for example `_add_round` would become
```julia
[...]

PR #621, opened 2024-01-26
Ensure thread-safety for trigonometric functions

### LINKED ISSUE #620: `quadrant` is not thread-safe
See also https://github.com/JuliaLang/julia/issues/52862 and the discussion in https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/612.

OlivierHnt:
Closes #620

This PR makes sure we no longer rely on Julia default code to compare with `π` due to https://github.com/JuliaLang/julia/issues/52862.
This should ensure that trigonometric functions are thread-safe.

The PR also reworks the inner function `_quadrant`.

Now I consistently get the same precision before and after calling `sin` on multiple threads (see also #612):

```Julia
julia> x = interval(BigFloat, 1)
[1.0, 1.0]₂₅₆_com

julia> y = sin(x)
[0.84147, 0.841472]₂₅₆_com

julia> Threads.@threads for _ in 1:1000
           z = sin(x)
           @assert isequal_interval(y, z)
       end

julia> precision(BigFloat)
256
```

The implementation is based on @Joel-Dahne, @lbenet and @dpsanders suggestions.

Hopefully everything is in order! 🙂

### DIFF
--- a/src/intervals/arithmetic/trigonometric.jl
+++ b/src/intervals/arithmetic/trigonometric.jl
@@ -12,10 +12,10 @@ function _quadrant(x::AbstractFloat)
     r = rem2pi(x, RoundNearest)
     r2 = 2r # should be exact for floats
     r2 ≤ -PI_HI && return 2 # [-π, -π/2)
-    r2 < -PI_LO && return throw(ArgumentError("could not determine if $x is lesser or greater than -π/2"))
+    r2 < -PI_LO && return throw(ArgumentError("could not determine the quadrant, the remainder $r of the division of $x by 2π is lesser or greater than -π/2"))
     r2 <  0     && return 3 # [-π/2, 0)
     r2 ≤  PI_LO && return 0 # [0, π/2)
-    r2 <  PI_HI && return throw(ArgumentError("could not determine if $x is lesser or greater than π/2"))
+    r2 <  PI_HI && return throw(ArgumentError("could not determine the quadrant, the remainder $r of the division of $x by 2π is lesser or greater than π/2"))
     return 1 # [π/2, π]
 end
 
--- a/src/intervals/arithmetic/trigonometric.jl
+++ b/src/intervals/arithmetic/trigonometric.jl
@@ -5,15 +5,31 @@
 # helper functions
 
 function _quadrant(x::AbstractFloat)
-    r = rem2pi(x, RoundNearest)
-    -2r > π && return 2 # [-π, -π/2)
-    r   < 0 && return 3 # [-π/2, 0)
-    2r  < π && return 0 # [0, π/2)
-    return 1 # [π/2, π]
+    PI_LO, PI_HI = bounds(bareinterval(typeof(x), π))
+
+    rlo = rem2pi(x, RoundDown) # [0, 2π]
+    qlo =        2rlo ≥ 3PI_HI ? 3 :
+        3PI_LO ≥ 2rlo ≥ 2PI_HI ? 2 :
+        2PI_LO ≥ 2rlo ≥  PI_HI ? 1 :
+         PI_LO ≥ 2rlo          ? 0 :
+        return throw(ArgumentError("could not determine a specific quadrant, got $qlo"))
+    rlo == x && return qlo
+
+    rhi = rem2pi(x, RoundUp) # [-2π, 0]
+    qhi =        -2rhi ≥ 3PI_HI ? 0 :
+        3PI_LO ≥ -2rhi ≥ 2PI_HI ? 1 :
+        2PI_LO ≥ -2rhi ≥  PI_HI ? 2 :
+         PI_LO ≥ -2rhi          ? 3 :
+        return throw(ArgumentError("could not determine a specific quadrant, got $qhi"))
+    rhi == x && return qhi
+
+    qlo == qhi && return qlo
+
+    return throw(ArgumentError("could not determine a specific quadrant, got $qlo and $qhi"))
 end
 
 function _quadrantpi(x::AbstractFloat) # used in `sinpi` and `cospi`
-    r = rem(x, 2)
+    r = rem(x, 2) # [-2, 2] (should be exact for floats)
     2r < -3 && return 0 # [-2π, -3π/2)
     r  < -1 && return 1 # [-3π/2, -π)
     2r < -1 && return 2 # [-π, -π/2)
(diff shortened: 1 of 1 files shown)


src/intervals/arithmetic/trigonometric.jl
```diff
@@ -5,15 +5,31 @@
...
-    2r  < π && return 0 # [0, π/2)
-    return 1 # [π/2, π]
+    PI_LO, PI_HI = bounds(bareinterval(typeof(x), π))
+
+    rlo = rem2pi(x, RoundDown) # [0, 2π]
+    qlo =        2rlo ≥ 3PI_HI ? 3 :
+        3PI_LO ≥ 2rlo ≥ 2PI_HI ? 2 :
+        2PI_LO ≥ 2rlo ≥  PI_HI ? 1 :
+         PI_LO ≥ 2rlo          ? 0 :
+        return throw(ArgumentError("could not determine a specific quadrant, got $qlo"))
```
Joel-Dahne, line 15: I think it should be `$rlo` in the error.

src/intervals/arithmetic/trigonometric.jl
```diff
@@ -5,15 +5,31 @@
...
+    rlo == x && return qlo
+
+    rhi = rem2pi(x, RoundUp) # [-2π, 0]
+    qhi =        -2rhi ≥ 3PI_HI ? 0 :
+        3PI_LO ≥ -2rhi ≥ 2PI_HI ? 1 :
+        2PI_LO ≥ -2rhi ≥  PI_HI ? 2 :
+         PI_LO ≥ -2rhi          ? 3 :
+        return throw(ArgumentError("could not determine a specific quadrant, got $qhi"))
```
Joel-Dahne, line 23: I think it should be `$rhi` in the error.

src/intervals/arithmetic/trigonometric.jl
```diff
@@ -5,15 +5,31 @@
...
 
 function _quadrant(x::AbstractFloat)
-    r = rem2pi(x, RoundNearest)
-    -2r > π && return 2 # [-π, -π/2)
-    r   < 0 && return 3 # [-π/2, 0)
-    2r  < π && return 0 # [0, π/2)
-    return 1 # [π/2, π]
+    PI_LO, PI_HI = bounds(bareinterval(typeof(x), π))
+
+    rlo = rem2pi(x, RoundDown) # [0, 2π]
```
Joel-Dahne, line 10: I don't think `rem2pi` can be assumed to give correct, or even faithful rounding. 

For `BigFloat` it is defined by
```julia
rem2pi(x::BigFloat, r::RoundingMode) = rem(x, 2*BigFloat(pi), r)
```
which I don't think would guarantee the correct rounding since $2\pi$ is not computed to any  higher precision and also not rounded in any specific direction. There is even a TODO-comment
```julia
# TODO: use a higher-precision BigFloat(pi) here?
```
So the one who implemented most likely didn't prove that it would be correct. I'm not sure how easy it would be to find a counter example though.

For `Float64` it has a quite complicated implementation and it probably gives very good results in general. But it was also easy to find a counterexample to the correctness. If we define
```julia
g = (rng, N) -> begin
    for _ in 1:N
        x = reinterpret(Float64, rand(rng, UInt))
        rem2pi(x, RoundDown) == Float64(rem2pi(big(x), RoundDown)) || error("x = $x")
    end
end
```
then it was enough to run
```julia
g(MersenneTwister(42), 1000)
```
to get the counterexample `x = 8.197839233774979e66`. For this `x` we have that the correctly rounded result should be `1.9664659792957133`, but `rem2pi(x)` gives us `1.9664659792957135` (so not even faithfully rounded).

I don't think this should be regarded as a bug in Julia though, most numerical methods don't give correct rounding.

src/intervals/arithmetic/trigonometric.jl
```diff
@@ -5,15 +5,31 @@
...
+    qlo =        2rlo ≥ 3PI_HI ? 3 :
+        3PI_LO ≥ 2rlo ≥ 2PI_HI ? 2 :
+        2PI_LO ≥ 2rlo ≥  PI_HI ? 1 :
+         PI_LO ≥ 2rlo          ? 0 :
```
Joel-Dahne, line 14: Wouldn't you need to account for the rounding here?

src/intervals/arithmetic/trigonometric.jl
Joel-Dahne: This issue was of course already present in the previous implementation. It is maybe reasonable to merge this PR, which fixes some issues, and leave this issue for another day.

src/intervals/arithmetic/trigonometric.jl
```diff
@@ -5,15 +5,31 @@
...
+        return throw(ArgumentError("could not determine a specific quadrant, got $qlo"))
+    rlo == x && return qlo
+
+    rhi = rem2pi(x, RoundUp) # [-2π, 0]
+    qhi =        -2rhi ≥ 3PI_HI ? 0 :
+        3PI_LO ≥ -2rhi ≥ 2PI_HI ? 1 :
+        2PI_LO ≥ -2rhi ≥  PI_HI ? 2 :
+         PI_LO ≥ -2rhi          ? 3 :
```
Joel-Dahne, line 22: Wouldn't you need to account for the rounding here?

src/intervals/arithmetic/trigonometric.jl
Joel-Dahne: Of course it is probably extremely rare that any rounding errors in `rem2pi` actually affects the returned quadrant. It wouldn't surprise me if it actually gives correct results for all `Float64` values.

OlivierHnt: Thx for peeking at the code for `rem2pi`. Yes perhaps this can be addressed in a separate PR. Then perhaps I should go back to using `rem2pi(..., RoundNearest)` since the version in this PR is flawed too, and less performant.

Joel-Dahne: To get a correct implementation I believe the easiest, but certainly not the fastest, is to more or less compute the remainder fully in interval arithmetic. Then you get a proper enclosure of it at least, if that enclosure is not enough to determine the quadrant you can fall back to some more pessimistic version. This would guarantee faithful, but not necessarily exactly correct determination of the quadrant. 

For `Float64` there is an implementation in CRlibm [here](https://github.com/taschini/pycrlibm/blob/master/crlibm/rem_pio2_accurate.c). This is likely very fast and would also guarantee a correct result. I'm not sure if it is part of the public interface though...

OlivierHnt: Ah great thx for the reference.

I'll merge this PR (which now only focuses on thread-safety) and open an issue for `_quadrant` not being reliable.

dpsanders:
We could ask the CRlibm people if we can transcribe their code into Julia. (I think the license is problematic for compatibility with our MIT license.)

I believe this is a standard algorithm, so alternatively somebody could implement it based on the relevant paper.

PR #623, opened 2024-02-01
Extend boolean functions for intervals to numbers

OlivierHnt:
This PR improves compatibility between `Number` and `Interval` by permitting most boolean functions to take numbers as an argument.

The general idea is that a boolean function defined for intervals, say `foo(x::Interval) = ...`, is extended to numbers by simply `foo(x::Number) = foo(interval(x))`.

This leads to some corner cases, eg `isthin(Inf, Inf)` and `isthin(pi)` return false.
Not sure if that should be the behaviour. The other possibility is of course to fall back to default Base functionalities: `isthinzero(x::Number) = iszero(x)` instead of `isthinzero(x::Number) = isthinzero(interval(x))`, etc.

@lbenet do you think this can help with https://github.com/JuliaDiff/TaylorSeries.jl/pull/345? Perhaps then you won't have to define your own `_isthinzero` function.

### DIFF
--- a/src/intervals/interval_operations/boolean.jl
+++ b/src/intervals/interval_operations/boolean.jl
@@ -4,8 +4,13 @@
 # Some other (non required) related functions are also present, as well as some
 # of the "Recommended operations" (Section 10.6.3)
 
-# used internally, equivalent to `<` but with `(Inf < Inf) == true`
-_strictlessprime(x::Real, y::Real) = (x < y) | ((isinf(x) | isinf(y)) & (x == y))
+#
+
+_strictlessprime(x::Real, y::Real) = _strictlessprime(default_flavor(), x, y)
+_strictlessprime(::Flavor{:set_based}, x::Real, y::Real) =
+    (x < y) | ((isinf(x) | isinf(y)) & (x == y)) # equivalent to `<` but with `(Inf < Inf) == true`
+
+#
 
 """
     isequal_interval(x, y)
@@ -28,6 +33,8 @@ isequal_interval(x::Real, y::Complex) = isequal_interval(x, real(y)) & isthinzer
 
 isequal_interval(x) = Base.Fix2(isequal_interval, x)
 
+isequal_interval(x::Number, y::Number) = isequal_interval(interval(x), interval(y))
+
 """
     issubset_interval(x, y)
 
@@ -46,6 +53,8 @@ issubset_interval(x, y, z, w...) = issubset_interval(x, y) & issubset_interval(y
 issubset_interval(x::Complex, y::Complex) = issubset_interval(real(x), real(y)) & issubset_interval(imag(x), imag(y))
 issubset_interval(x::Complex, y::Real) = issubset_interval(real(x), y) & isthinzero(imag(x))
 issubset_interval(x::Real, y::Complex) = issubset_interval(x, real(y)) & in_interval(0, imag(y))
+
+issubset_interval(x::Number, y::Number) = issubset_interval(interval(x), interval(y))
 
 """
     isinterior(x, y)
@@ -78,6 +87,8 @@ isinterior(x::Complex, y::Complex) =
 isinterior(x::Complex, y::Real) = isinterior(real(x), y) & isthinzero(imag(x))
 isinterior(x::Real, y::Complex) = isinterior(x, real(y)) & in_interval(0, imag(y))
 
+isinterior(x::Number, y::Number) = isinterior(interval(x), interval(y))
+
 """
     isstrictsubset(x, y)
 
@@ -100,6 +111,8 @@ isstrictsubset(x::Complex, y::Complex) =
 isstrictsubset(x::Complex, y::Real) = isstrictsubset(real(x), y) & isthinzero(imag(x))
 isstrictsubset(x::Real, y::Complex) = isstrictsubset(x, real(y)) & in_interval(0, imag(y))
 
+isstrictsubset(x::Number, y::Number) = isstrictsubset(interval(x), interval(y))
+
 """
     isweakless(x, y)
 
@@ -115,6 +128,8 @@ function isweakless(x::Interval, y::Interval)
     return isweakless(bareinterval(x), bareinterval(y))
 end
 
+isweakless(x::Number, y::Number) = isweakless(interval(x), interval(y))
+
 """
     isstrictless(x, y)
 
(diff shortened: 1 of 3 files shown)


OlivierHnt:
The behaviour of `isthin`, etc., should be consistent with `inf`, `sup` and `bounds`. Currently `sup(Inf) == NaN` because `sup(x::Number) = sup(interval(x))`.

Also

```julia
julia> inf(pi)
3.141592653589793

julia> sup(pi)
3.1415926535897936
```

lbenet:
> The general idea is that a boolean function defined for intervals, say `foo(x::Interval) = ...`, is extended to numbers by simply `foo(x::Number) = foo(interval(x))`.
>
> This leads to some corner cases, eg `isthin(Inf, Inf)` and `isthin(pi)` return false. Not sure if that should be the behaviour. The other possibility is of course to fall back to default Base functionalities: `isthinzero(x::Number) = iszero(x)` instead of `isthinzero(x::Number) = isthinzero(interval(x))`, etc.
> 

Thanks for this addition and bringing it up for discusion.

A subtlety of the current proposal is that in some cases we do create the proper interval that contains the real value intended (or typed), e.g. `interval(pi)` which is not thin, and in other we do not, e.g. `interval(0.1)`. I understand why this is so internally (`::Irrational` vs `::Float64`), which is leave to the user the responsability of what is being typed. Yet, the question is if we should systematically use the proper interval that contains the real number that is typed, or we should treat it as a numeric computational type?

In https://github.com/JuliaDiff/TaylorSeries.jl/pull/345 I've been joggling with some internal functionality as a workaround (this was before #613). What I did was to define some internal function, say `_isthinzero`, which behaves as `iszero` always (`Number`), and for `Interval`s it corresponds to `isthinzero`. That way I am not incurring into type piracy --I think--, and allows me to have all Interval functionality as an external dependency. This approach is closer to the *other possibility* you mentioned above... 

Something I noticed in https://github.com/JuliaDiff/TaylorSeries.jl/pull/345 (and that I'm trying to solve) may help in the decision: In order to compute the Taylor series of certain functions (let us think of `log`), you first need to be able to compute the zero order term, which is the `log` of the constant term. For floating point numbers, `log(x)` throws an error for `x<0`, returns `-Inf` (which is useless) for `x==0`, and does what it should for positive `x`; so for floating points, it is enough to check `iszero(x)`. Now, if `x` is an interval, the `log` computation involves checking the correct domain (an intersection), and that process may be an empty interval, or an interval that contains zero. The empty interval propagates, so I should check it and throw an error in that case, and if the interval contains zero, there will be a lot of `-Inf`s, which are essentially useless. My point is, the `iszero` check returning `false` is not enough.

OlivierHnt:
Ah yes you have IntervalArithmetic as a weak dependancy. The way you do it with `_isthinzero` seems the right way.

In the example of `log` the effective check is `x >= 0` (if one does not want to rely on Base throwing an error whenever `x < 0`). Btw this now works in this PR if `x :: Interval` provided that `isthin(x)` holds.

> A subtlety of the current proposal is that in some cases we do create the proper interval that contains the real value intended (or typed), e.g. interval(pi) which is not thin, and in other we do not, e.g. interval(0.1). I understand why this is so internally (::Irrational vs ::Float64), which is leave to the user the responsability of what is being typed. Yet, the question is if we should systematically use the proper interval that contains the real number that is typed, or we should treat it as a numeric computational type?

Absolutely.
When I re-defined `inf` and `sup` to convert the input to an `Interval` first, I do not think that I took into account what would happen for an `Irrational`.
Mhmm from our own docstring for `isthin`, having `isthin(pi) == false` is kind of contradictory. So we could consider this a bug (and thus also change the behaviour of `inf` and `sup`), or alter the docstring.

lbenet:
> n the example of log the effective check is x >= 0 (if one does not want to rely on Base throwing an error whenever x < 0). Btw this now works in this PR if x :: Interval provided that isthin(x) holds.

Thanks for the remark. I'll give it a try and provide some feedback

lbenet:
> When I re-defined `inf` and `sup` to convert the input to an `Interval` first, I do not think that I took into account what would happen for an `Irrational`.
Mhmm from our own docstring for `isthin`, having `isthin(pi) == false` is kind of contradictory. So we could consider this a bug (and thus also change the behaviour of `inf` and `sup`), or alter the docstring.

Just to make it clear: I am not against that `interval(pi)` returns the correct interval that contains the (mathematically) real pi; I accept the fact that `interval(0.1)` returns a thin interval (which does not contain the actual 1//10); my point is that `interval` behaves slightly (but deterministically) inconsistent, depending on the (Julia) type of its entries. 
Once stated this, I think that `isthin(interval(pi))` correctly returns `false`, while `isthin(interval(0.1))` or `isthin(interval(1))` correctly return `true`; in my opinion there is no bug.

OlivierHnt:
> Thanks for the remark. I'll give it a try and provide some feedback

Oh I was just mentioning this by the way.
I do not think it will fit your needs since whenever `x` is not thin, you will get an error... which is very impractical but provide a slightly better support for some safe generic code.

>Just to make it clear: I am not against that interval(pi) returns the correct interval that contains the (mathematically) real pi; I accept the fact that interval(0.1) returns a thin interval (which does not contain the actual 1//10); my point is that interval behaves slightly (but deterministically) inconsistent, depending on the (Julia) type of its entries.
Once stated this, I think that isthin(interval(pi)) correctly returns false, while isthin(interval(0.1)) or isthin(interval(1)) correctly return true; in my opinion there is no bug.

Indeed. Now it's just a matter of deciding whether `isthin(pi)` should return false or true 🤷‍♂️ (and also for `isthin(Inf)`, but this one is more of a corner case)

lbenet:
Using this branch, I get the following:
```julia
julia> x = interval(0.5,1)
Interval{Float64}(0.5, 1.0, com)

julia> x >= 0
ERROR: ArgumentError: `==` is only supported for thin intervals. See instead `isequal_interval`
...
```

It seems to me that, using directly the comparison (`>=`, or something alike, as I do for Julia `Number`s), due to the error, I wouldn't be able to approximate `log(x)` as a Taylor series for the interval `x`. (This is part of the reason I check if `log(x)` is an empty interval, in which case because it includes the `intersect_interval` operation which is precisely the filter I need.

OlivierHnt:
Yea that's what I was referring to in my previous comment.

Somehow I can wrap my head around `isthin(pi)` being false, since an `Irrational` can never be an exact numerical value.
But this is puzzling
```Julia
julia> isunbounded(Inf)
┌ Warning: invalid interval, NaI is returned
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/construction.jl:417
false
```
One good thing is that we have a warning message...

lbenet:
> Somehow I can wrap my head around `isthin(pi)` being false, since an `Irrational` can never be an exact numerical value.

I agree!

> But this is puzzling
> 
> ```julia
> julia> isunbounded(Inf)
> ┌ Warning: invalid interval, NaI is returned
> └ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/construction.jl:417
> false
> ```
> One good thing is that we have a warning message... And it's also not wrong in the sense that `Inf` does not correspond to an unbounded interval (this is flavour dependant), but in practice one may expect `isunbounded` to be equivalent to `isinf` for numbers.

This is indeed puzzling, but ok in my opinion; I think we are consistent with the fact that `interval(Inf)` corresponds to NaI; I also think this is in also consistent with the standard, isn't it?

PR #627, opened 2024-02-12
New macro for guaranteed operations

Kolaru:
I discussed with Luis and I put up a quick prototype on how we could extend the current `@interval` macro.

The strategy drafted here is to convert every literals into `ExactNumber` with the `@exact` macro.

An `ExactNumber` promote to real when mixed with reals, and to a guaranteed interval when interacting with intervals.

```julia
julia> x = ExactNumber(2.2)
ExactNumber{Float64}(2.2)

julia> x + 1
3.2

julia> x + interval(1, 1)
[3.19999, 3.20001]_com
```

The `@exact` macro works on both expression and function definition (thanks to MacroTools.jl)
```julia
julia> y = interval(1, 2)
[1.0, 2.0]_com

julia> @exact 1.2 * y + 3
[4.19999, 5.40001]_com

julia> @exact f(x) = 1.2 * x + 3

julia> f(y)
[4.19999, 5.40001]_com

julia> f(mid(y))
4.8
```

I haven't yet checked how it would play out with IntervalRootFinding or ForwardDiff.


Kolaru:
ForwardDiff seems to not complain

```julia

julia> using ForwardDiff: derivative

julia> @exact f(x) = 3*sin(10x) - 2.2/x

julia> derivative(f, 1.2)
26.843396539752543

julia> derivative(f, interval(1.2, 1.2))
[26.8433, 26.8435]_com
```
Except when it uses comparisons of course
```julia
julia> @exact g(x) = x^2

julia> derivative(g, 1.2)
2.4

julia> derivative(g, interval(1.2, 1.2))
ERROR: ArgumentError: `==` is purposely not supported for intervals. See instead `isequal_interval`
```

Kolaru:
Had a bit of fun and found a way to make the floating point input problem apparent:

```julia
julia> x = interval(0, 1)
[0.0, 1.0]_com

julia> @exact 0.5x + 0.1
┌ Warning: Float64 displayed as 0.1 is equal to 0.1000000000000000055511151231257827021181583404541015625
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/exact_numbers.jl:33
[0.0999999, 0.600001]_com
```

The same trick should probably be used with `@interval` to warn the user.

lbenet:
This is very very nice! 🚀 

Let me begin with the following (side) remark: we can use the same trick to *find* the actual rounding (up or down) when converting a real number $x\in \mathbb{R}$ into a `Float64`. This could be used, e.g., to create a tight interval from the user provided $x$:
```julia
julia> x = 0.1
0.1

julia> ExactNumber(x)
ExactNumber{Float64}(0.1000000000000000055511151231257827021181583404541015625)

julia> string(x) < string(ExactNumber(x)) # so 0.1::Float64 is rounded up!
true

# A tight interval representation of the real 0.1 is:
julia> interval(prevfloat(x), x)
Interval{Float64}(0.09999999999999999, 0.1, com)

julia> x = 0.3
0.3

julia> ExactNumber(x)
ExactNumber{Float64}(0.299999999999999988897769753748434595763683319091796875)

julia> string(x) > string(ExactNumber(x)) # 0.3::Float64 is rounded down!
true

# A tight interval representation of the real 0.3 is:
julia> interval(x, nextfloat(x))
Interval{Float64}(0.3, 0.30000000000000004, com)
```
(I guess certain changes are required for negative numbers, but the idea is similar.)

The above raises the question: Do we want to "operate" with `interval(0.1)` or with the tight interval associated to $x = 0.1\in \mathbb{R}$?

lbenet:
Aside from the comment above, and knowing this is still a prototype, what about other numeric formats (`Rational`s, `BigFloat`s, ``Interval`s? While (correctly) `@exact` throws no warning for `Rational`s, there is a `MethodError` involving `//` (and also `/`) and `ExactNumber{Int64}`. Similar errors occur for `BigFloat's...

Kolaru:
> Let me begin with the following (side) remark: we can use the same trick to find the actual rounding (up or down)

Unfortunately, not quite, because we never know *what the user did input*. Anything that is parsed as `0.1` is equal for us.

```julia
julia> @exact 0.1
┌ Warning: Float64 displayed as 0.1 is equal to 0.1000000000000000055511151231257827021181583404541015625
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/exact_numbers.jl:33
ExactNumber{Float64}(0.1000000000000000055511151231257827021181583404541015625)

julia> @exact 0.09999999999999999999782
┌ Warning: Float64 displayed as 0.1 is equal to 0.1000000000000000055511151231257827021181583404541015625
└ @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/exact_numbers.jl:33
ExactNumber{Float64}(0.1000000000000000055511151231257827021181583404541015625)
```

> Aside from the comment above, and knowing this is still a prototype, what about other numeric formats

If we agree this is the way to go I'll add the required methods for Rational, BigFloat and the basic arithmetic operations.

OlivierHnt:
So the purpose of the macro `@exact` is to notify that every literal numbers are "guaranteed", but not only, it would also make something like `x= 1; @exact interval(1) + x` be also guaranteed, correct?
I think this is nice.

src/intervals/exact_numbers.jl
```diff
@@ -0,0 +1,68 @@
...
+
+# Promotion to Real
+Base.promote_rule(::Type{T}, ::Type{ExactNumber{S}}) where {T<:Real,S<:Real} =
+    promote_type(T, S)
+
+Base.promote_rule(::Type{ExactNumber{T}}, ::Type{S}) where {T<:Real,S<:Real} =
+    promote_type(T, S)
+
+Base.convert(::Type{T}, x::ExactNumber) where {T<:Real} =
+    convert(T, x.number)
```
OlivierHnt, line 25: This is more a comment, or just a note. This would allow two `ExactNumber` to be converted, but this in principle can lead do not correct rounding. That being said, I am not opposed to this, we should just make it explicitly in the docstring

src/intervals/exact_numbers.jl
```diff
@@ -0,0 +1,68 @@
+struct ExactNumber{T} <: Real
```
OlivierHnt, line 1: We would want `Complex{<:ExactNumber}` to be supported, no?
In which case it may be better to have `ExactReal`, or something of the sort.

src/intervals/exact_numbers.jl
```diff
@@ -0,0 +1,68 @@
...
+    @eval (Base.$binary)(x::ExactNumber, y::ExactNumber) = ($binary)(interval(x), interval(y))
+end
+
+macro exact(expr)
+    exact_expr = postwalk(expr) do x
+        if x isa Real
+            e = ExactNumber(x)
+            s = string(e)
+            if s != string(x)
+                @warn "$(typeof(x)) displayed as $x is equal to $s"
```
OlivierHnt, line 52: My first impulse would be to not have yet an other warning being prompted. If the user types in explicitly @exact, then it's on them that they know about floats.

OlivierHnt (changes requested):

src/intervals/exact_numbers.jl
Kolaru: Good point.

Because of the convert fallback normal reals can also silently be converted to `ExactReal`.

And that's not okay, so I will remove all conversions to `ExactReal`.

OlivierHnt:
Looking back at the current implementation, I find the goal of `@exact` a bit confusing.

I thought the purpose was to allow us to write general code (code used for both floating-points arithmetic and interval arithmetic), and avoid propagating the "NG" flag.
Of course, it is the responsibility of the user to know that their code is complies with the specifications of both arithmetics.

To be concrete, suppose I have a function `f(x) = 1 + 2x`. Then this functions is completely valid for both floats and interval inputs. But currently the result is marked by "NG" if I use an interval; typically I would duplicate the code, and write a new method `f(x::Interval) = @interval Float64 1+2x`.

Now if I define `g(x) = @exact 1+2x`, I do not want to always give me back an interval (as `@interval` would do). I would want `g` to also be used on floats (non-rigorous arithmetic). So `g(1) == 3`, while `g(interval(1))` is `interval(3)` (with no "NG" flag).

This means that an operation involving `ExactReal` is only rigorous if all the other operands involved are intervals.
Thus, `@exact sqrt(3) * x + 1` reads `sqrt(ExactReal(3)) * x + ExactReal(1)` with `sqrt(ExactReal(3)) == sqrt(3)`.

I think this would be a very powerful feature. One situation I have in mind is Newton's method for a function `f`, where we typically want to perform as many iterations in floats, and only use intervals once we get close enough to the zero.

src/intervals/exact_literals.jl
```diff
@@ -0,0 +1,66 @@
...
+# Promotion to Interval
+Base.promote_rule(::Type{Interval{T}}, ::Type{ExactReal{S}}) where {T<:NumTypes,S<:Real} =
+    Interval{promote_numtype(T, S)}
+
+Base.promote_rule(::Type{ExactReal{T}}, ::Type{Interval{S}}) where {T<:Real,S<:NumTypes} =
+    Interval{promote_numtype(T, S)}
+
+function Base.convert(::Type{Interval{T}}, x::ExactReal) where {T<:NumTypes}
+    y = interval(T, x)
+    return _unsafe_interval(bareinterval(y), decoration(y), true)
```
OlivierHnt, line 14: This may be unnecessary. Doesn't `y` already satisfy `isguaranteed(y) == true` already?

Kolaru:
I think this PR is roughly where we want it to be.

The formulation of the doc would probably benefit from the review.

Once think that this PR is not doing is to deal with conversion of `ExactReal{T}` to `Interval{S}` when `T` and `S` are different (and `T` is not an Integer). In principle with different precision floating points number some care is required. I think it is niche enough to leave it unimplemented for now.

OlivierHnt:
The CI failure seems to be due to broken tests introduced in https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/619, which now pass on nightly.

One tiny note: I am trying to think of a single word for the name of the macro. How about `@literal`? Or `@guarantee`?

OlivierHnt:
@Kolaru
I made very minor changes to the PR. I reorganised the docstrings a bit.
I also removed the constructor `ExactReal{T}` since this is a fallback used everywhere for `Real` in Julia.

Lastly I ran into issues when the expression contained `im`, so I made an exception for this symbol in the macro. Can you double check this?

Also, I do not understand how the display works. Is the string being prompted truly the exact mathematical number?

Kolaru:
> One tiny note: I am trying to think of a single word for the name of the macro. How about `@guarantee`, `@literals`? Or just `@exact`, since the purpose is to create an `ExactReal`?

The most descriptive name would probably be `@user_guaranteed_literals`... I think the most reasonable is `@exact`.

> Also, I do not understand how the display works. Is the string being prompted truly the exact mathematical number?

Yes. I ask the Ryu library (I have no idea what it is, but that's what Base uses) to write the number with 2000 decimals and truncate the extra 0s. According to my tests and estimation, 2000 decimals is more than enough to have the exact decimal representation of any Float64.

> Lastly I ran into issues when the expression contained `im`, so I made an exception for this symbol in the macro. Can you double check this?

I fixed the constructor for complex such that `@exact 1.2 + 3.4im` now works. It may still be a bit brittle, if you have any additional example feel free to add them to the tests and I can have a look.

OlivierHnt:
Looks good to merge, once the tests are successful. I glanced at it but I am not sure what went wrong.

Kolaru:
I forgot to rename the macro.

Kolaru:
I don't know what is going on with nightly or the doctests. Do the doctests use nightly julia ? It may be the same problem.

OlivierHnt:
It seems that the problem with the doctests is due to the printing of the intervals. Somehow it expects `setdisplay(:full)`.

PR #633, opened 2024-03-10
Use rem2pi and rem with BigFloats in _quadrant

lbenet:
This PR proposes to use `rem2pi` and `rem` with `BigFloat`s, instead of the same methods for the specific `AbsrtracFloat`s. The reason is that, in order to guarantee the precision one needs more "guard digits" than the actual precision; see J-M Muller, "Elementary Functions: Algorithms and implementations", Birkhäuser, 3rd ed (2005), chap. 11. While the solution to use `BigFloat`s should have a performance impact, it seems to be at least safe, without implementing more elaborate range reduction methods.

The following illustrates such a case:
```julia
julia> x = 6381956970095103 * 2.0^797 # worst case for C = pi/2 for Float64
5.319372648326541e255

julia> xb = big(x) 5.319372648326541416707296656673541083813475031793921822105998164685326343987747e+255

julia> rem2pi(x, RoundNearest)
1.5707963267948966

julia> rem2pi(xb, RoundNearest)
-2.545982325227109106133284489709240971297101919468998568916440481134254101023789
```

As far as I have checked, or the tests show, there haven't been any issues with this change.

This approach could be adapted for the `mod` (or should it be `rem`?) function; cf #129, #145 and #178).

### DIFF
--- a/src/intervals/arithmetic/trigonometric.jl
+++ b/src/intervals/arithmetic/trigonometric.jl
@@ -3,13 +3,23 @@
 # in Section 10.5.3
 
 # helper functions
+# NOTE: According to J-M Muller, "Elementary Functions: Algorithms and
+# implementations", Birkhäuser, 3rd ed (2005), chap. 11, pp 203,
+# "The first solution consists of using multiple-precision arithmetic,
+# but this may make the computation significantly slower. Moreover,
+# it is not that easy to predict the accuracy with which the computation
+# should be performed." Later (pp 210), it estimates that the precision needed
+# to compute correctly the worst case of the range reduction algorithm
+# for Float64 is about 61 "guard digits". Below we use `rem2pi` for
+# BigFloats, with at least 200 bits of precision.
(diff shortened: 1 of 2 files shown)


OlivierHnt:
Thx for looking into this! I will need to have a close look at the reference you provided.

Nota bene: at first glance Section 11.3.2 seems to give us a way to estimate the required number of guard digits, denoted by $-\log_\beta(\epsilon)$ in the text. Then we can infer the necessary precision needed for safely performing `rem2pi`, so that we can construct `bx = BigFloat(x; precision = precision(x) + N)`.

lbenet:
> Nota bene: at first glance Section 11.3.2 seems to give us a way to estimate the required number of guard digits, denoted by −logβ⁡(ϵ) in the text. Then we can infer the necessary precision needed for safely performing `rem2pi`, so that we can construct `bx = BigFloat(x; precision = precision(x) + N)`.

I fully agree; it is much smaller the actual guard bits needed. Yet, I thought it was a bit more messy to get the minimum precision, compute whatever for that precision, and go back, than simply check that the BigFloat precision is large enough.

PR #636, opened 2024-03-21
Improve support for complex intervals

OlivierHnt:
The goal of this PR is to address https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/41 and and https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/295.

TODO: `sqrt`, `^`, and inverse trigonometric/hyperbolic functions.

### DIFF
--- a/src/intervals/interval_operations/numeric.jl
+++ b/src/intervals/interval_operations/numeric.jl
@@ -88,7 +88,7 @@ See also: [`inf`](@ref), [`sup`](@ref), [`bounds`](@ref), [`diam`](@ref),
 [`radius`](@ref) and [`midradius`](@ref).
 """
 function mid(x::BareInterval{T}, α = 0.5) where {T<:AbstractFloat}
-    0 ≤ α ≤ 1 || throw(DomainError(α, "α must be between 0 and 1"))
+    0 ≤ α ≤ 1 || return throw(DomainError(α, "α must be between 0 and 1"))
     isempty_interval(x) && return convert(T, NaN)
     if isentire_interval(x)
         α == 0.5 && return zero(T)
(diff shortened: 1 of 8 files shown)


src/intervals/interval_operations/numeric.jl
```diff
@@ -88,7 +88,7 @@ See also: [`inf`](@ref), [`sup`](@ref), [`bounds`](@ref), [`diam`](@ref),
 [`radius`](@ref) and [`midradius`](@ref).
 """
 function mid(x::BareInterval{T}, α = 0.5) where {T<:AbstractFloat}
-    0 ≤ α ≤ 1 || throw(DomainError(α, "α must be between 0 and 1"))
+    0 ≤ α ≤ 1 || return throw(DomainError(α, "α must be between 0 and 1"))
```
lbenet, line 91: Do we need the return here? I just wonder if that would cause some sort of type unstability (sometimes, the returned type would be an `Interval`, or a `DomainError`)

(Similar comment applies below.)

OlivierHnt: I think it is safe. I checked `@code_warntype` for

```Julia
function foo(x)
    x < 1 && return throw(DomainError("x is smaller than 1"))
    return true
end

function goo(x)
    x < 1 && throw(DomainError("x is smaller than 1"))
    return true
end
```

and the two functions seem to be compiled equivalently.

OlivierHnt:
I think this PR should be merged ASAP, even though it does not provide support for `sqrt`, `^`, and inverse trigonometric/hyperbolic functions.
It can always be done in a next PR.

PR #638, opened 2024-04-22
Small fixes for ExactReal

Kolaru:
This contains two small fixes that are useful for the update of IntervalRootFinding. 

1. Allow to use `ExactReal{<:Integer}` to index an array.
2. Fix derivative of `^` when mixing `ExactReal` and `Interval` in the ForwardDiff extension.

The first one in particular may be handy for everyone, when working with vectors of interval for example. With this PR, indexing work as intuitively expected with `@exact`
```julia
julia> @exact f(x) = x[1] + x[2] + 1.2

julia> X = [interval(-1, 1), interval(-0.1, 3)]
2-element Vector{Interval{Float64}}:
 [-1.0, 1.0]_com
 [-0.100001, 3.0]_com

julia> f(X)
[0.0999999, 5.20001]_com
```


OlivierHnt:
Nice ! LGTM

PR #641, opened 2024-05-01
Add an option to not display the NG flag

### LINKED ISSUE #639: Don't display `NG` when `decorations = false`
```
julia> setdisplay(decorations = false);

julia> x = 3..Inf
[3.0, ∞)

julia> f(y) = y^2 - 2

julia> f(x)
[7.0, ∞)_NG
```

OlivierHnt:
This PR simply adds the boolean keyword `ng_flag` to the `setdisplay` function. When set to false, the flag "NG" is not displayed.

Before merging this PR, can anyone think of a better keyword name than `ng_flag`?

Closes #639

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -45,7 +45,7 @@ julia> setdisplay(:infsup; decorations = true, sigdigits = 6) # default display 
 Display options:
   - format: infsup
   - decorations: true
-  - ng_flag: true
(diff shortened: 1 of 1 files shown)


OlivierHnt:
Since I got a thumb up, I'll go ahead and merge it :).

Issue #640, opened 2024-04-30
ForwardDiff with min, max, and clamp


smkatz12:
Hi! We are working on a textbook chapter on reachability analysis and are interested in support for intervals when using ForwardDiff.jl to compute gradients and hessians of functions that use the `min`, `max`, and `clamp`. The current behavior with these functions does not cause an error but is incorrect. We have written the functions below to correct for this. These functions could be added to `IntervalArithmeticForwardDiffExt.jl`.

```
function Base.max(x::Dual{T,V,N}, y::AbstractFloat) where {T,V<:Interval,N}
    if value(x).hi < y
        return Dual{T,V,N}(y..y, (0..0) * partials(x))
    elseif value(x).lo > y
        return Dual{T,V,N}(value(x), (1..1) * partials(x))
    else
        return Dual{T,V,N}(y..value(x).hi, (0..1) * partials(x))
    end
end

function Base.max(x::Dual{T,Dual{T2,V2,N2},N}, y::AbstractFloat) where {T,T2,V2<:Interval,N2,N}
    if value(value(x)).hi < y
        return Dual{T,Dual{T2,V2,N2},N}(Dual{T2,V2,N2}(y..y), (0..0) * partials(x))
    elseif value(value(x)).lo > y
        return Dual{T,Dual{T2,V2,N2},N}(value(x), (1..1) * partials(x))
    else
        return Dual{T,Dual{T2,V2,N2},N}(Dual{T2,V2,N2}(y..value(value(x)).hi, partials(value(x))), (0..1) * partials(x))
    end
end

function Base.min(x::Dual{T,V,N}, y::AbstractFloat) where {T,V<:Interval,N}
    if value(x).lo > y
        return Dual{T,V,N}(y..y, (0..0) * partials(x))
    elseif value(x).hi < y
        return Dual{T,V,N}(value(x), (1..1) * partials(x))
    else
        return Dual{T,V,N}(value(x).lo..y, (0..1) * partials(x))
    end
end

function Base.min(x::Dual{T,Dual{T2,V2,N2},N}, y::AbstractFloat) where {T,T2,V2<:Interval,N2,N}
    if value(value(x)).lo > y
        return Dual{T,Dual{T2,V2,N2},N}(Dual{T2,V2,N2}(y..y), (0..0) * partials(x))
    elseif value(value(x)).hi < y
        return Dual{T,Dual{T2,V2,N2},N}(value(x), (1..1) * partials(x))
[...]

PR #642, opened 2024-05-08
support for clamp and min, max, clamp with ForwardDiff

smkatz12:
Based on the discussion in #640, we (@mossr, @ancorso, @mykelk) have implemented the following:

- Support for `min` and `max` when one input is an interval and the other is a float
- Support for `clamp`
- Support for `min`,` max`, and `clamp` when computing first and second derivatives with ForwardDiff
- Additional tests for the new functionality

### DIFF
--- a/src/intervals/arithmetic/absmax.jl
+++ b/src/intervals/arithmetic/absmax.jl
@@ -50,7 +50,12 @@ for f ∈ (:min, :max)
             isempty_interval(y) && return y
             return _unsafe_bareinterval(T, $f(inf(x), inf(y)), $f(sup(x), sup(y)))
         end
+        function Base.$f(x::BareInterval{T}, y::Rational) where {T<:NumTypes}
+            isempty_interval(x) && return x
+            return _unsafe_bareinterval(T, $f(inf(x), y), $f(sup(x), y))
+        end
         Base.$f(x::BareInterval, y::BareInterval) = $f(promote(x, y)...)
+        Base.$f(x::BareInterval, y::Rational) = $f(promote(x), y)
 
         function Base.$f(x::Interval, y::Interval)
             r = $f(bareinterval(x), bareinterval(y))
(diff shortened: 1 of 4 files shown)


src/intervals/arithmetic/absmax.jl
```diff
@@ -50,13 +50,27 @@ for f ∈ (:min, :max)
             isempty_interval(y) && return y
             return _unsafe_bareinterval(T, $f(inf(x), inf(y)), $f(sup(x), sup(y)))
         end
+        function Base.$f(x::BareInterval{T}, y::Rational) where {T<:NumTypes}
+            isempty_interval(x) && return x
+            return _unsafe_bareinterval(T, $f(inf(x), y), $f(sup(x), y))
+        end
```
OlivierHnt, line 56: We purposely do not allow functions mixing `BareInterval` and `Number` in their computations, because `BareInterval` cannot keep track of guarantees ("NG" flag)or decorations.
So my first impulse here would be not to define such function.

src/intervals/arithmetic/absmax.jl
```diff
@@ -50,13 +50,27 @@ for f ∈ (:min, :max)
...
         Base.$f(x::BareInterval, y::BareInterval) = $f(promote(x, y)...)
+        Base.$f(x::BareInterval, y::Rational) = $f(promote(x), y)
```
OlivierHnt, line 58: Same comment as above.

src/intervals/arithmetic/absmax.jl
```diff
@@ -50,13 +50,27 @@ for f ∈ (:min, :max)
...
             t = isguaranteed(x) & isguaranteed(y)
             return _unsafe_interval(r, d, t)
         end
+
+        function Base.$f(x::Interval, y::Rational)
+            r = $f(bareinterval(x), y)
+            d = decoration(x)
+            return _unsafe_interval(r, d, false)
+        end
+        Base.$f(x::Rational, y::Interval) = $f(y, x)
```
OlivierHnt, line 72: On the latest version of IntervalArithmetic.jl, `min` and `max` already do this ; so defining these functions is maybe superfluous here ? Or am I missing something ?

Kolaru:
In term of decorations, for consistency we must have:
- `trv` when differentiated over the non-differentiable region (`x == y` for `min` and `max`, similar for `clamp`), for consistency with `abs`.
- We need to decide what to do with `clamp(x, lo, hi)` when `lo > hi`. `Base.clamp` doesn't complain, but we probably want to return an empty interval.
- `NG` when numbers and interval are mixed, for consistency with all other mixing operations.

PR #644, opened 2024-05-16
Ensure `string` is consistent with `show`

### LINKED ISSUE #643: `string(x::Interval)` is not working correctly
```jl

julia> x = interval(1, 2)
[1.0, 2.0]_com

julia> string(x)
"Interval{Float64}(1.0, 2.0, com)"
```

`string(x)` should return the string version of what is displayed.

This points to a possible issue in the implementation of `Base.show(io::IO, x::Interval)`

OlivierHnt:
Closes #643.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -100,6 +100,9 @@ end
 
 # printing mechanism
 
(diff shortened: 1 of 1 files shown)


dpsanders:
I suggest to add a test for this :)

PR #649, opened 2024-05-23
Fast intersection for BareInterval

dpsanders:
Redo of https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/379, which seems to have got lost.

3–4x faster on my machine.


OlivierHnt:
Nice!

PR #652, opened 2024-05-24
Reliable quadrant

### LINKED ISSUE #622: Finding quadrants is flawed
As pointed out here https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/621#discussion_r1469320240, the function `_quadrant` used internally for trigonometric functions may yield wrong results.

See https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/621#discussion_r1469423662 for a proposal to fix the issue (at least partially).

Also, this issue can be related to #129 on the implementation of `mod` for intervals (see also PR #145 and #178).

OlivierHnt:
This PR implements a reliable code to determine quadrants for trigonometric functions.

@dpsanders do you mind having a look?

Closes #622

### DIFF
--- a/src/intervals/arithmetic/trigonometric.jl
+++ b/src/intervals/arithmetic/trigonometric.jl
@@ -4,19 +4,22 @@
 
 # helper functions
 
-function _quadrant(x::AbstractFloat)
-    # NOTE: this algorithm may be flawed as it relies on `rem2pi(x, RoundNearest)`
-    # to yield a very tight result. This is not guaranteed by Julia, see e.g.
-    # https://github.com/JuliaLang/julia/blob/9669eecc99bc4553e28d94d7dd3dc9fd40b3bf3f/base/mpfr.jl#L845-L846
-    PI_LO, PI_HI = bounds(bareinterval(typeof(x), π))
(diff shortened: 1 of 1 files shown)


src/intervals/arithmetic/trigonometric.jl
```diff
@@ -65,8 +68,8 @@ function Base.sin(x::BareInterval{T}) where {T<:AbstractFloat}
 
     lo, hi = bounds(x)
 
-    lo_quadrant = _quadrant(lo)
-    hi_quadrant = _quadrant(hi)
+    lo_quadrant = _quadrant(min, lo)
```
dpsanders, line 71: Why do you need the `min` and `max` here?

dpsanders (approved):
LGTM as far as I can tell, thanks!

src/intervals/arithmetic/trigonometric.jl
OlivierHnt: `_quadrant` may return two distinct integers. What `min`/`max` aims to do is to "overshoot" the quadrants: for the lowest (resp. largest) value of the interval, we want the smallest (resp. largest) quadrant.

PR #656, opened 2024-05-26
Fix ExactReal with power in ForwardDiff

Kolaru:
The fixes the support for derivatives of powers containing ExactReal.

This *should* be the last missing piece before IntervalRootFinding can be upgraded to IA v0.22.

### DIFF
--- a/ext/IntervalArithmeticForwardDiffExt.jl
+++ b/ext/IntervalArithmeticForwardDiffExt.jl
@@ -56,4 +56,22 @@ function Base.:(^)(x::ExactReal, y::Dual{<:Any, I}) where I<:Interval
     return convert(I, x)^y
 end
 
+function Base.:(^)(x::Dual{<:Tx}, y::ExactReal) where Tx
+    v = value(x)
+    expv = v^y
+    if iszero(y) || all(iszero.(values(partials(x))))
+        new_partials = zero(partials(x))
+    else
+        new_partials = partials(x) * y * v^(y - 1)
+    end
+    return Dual{Tx}(expv, new_partials)
+end
+
+function Base.:(^)(x::ExactReal, y::Dual{<:Ty}) where Ty
+    v = value(y)
+    expv = x^v
+    deriv = (iszero(x) && inf(v) > 0) ? zero(expv) : expv*log(x)
+    return Dual{Ty}(expv, deriv * partials(y))
+end
+
 end
--- a/ext/IntervalArithmeticForwardDiffExt.jl
+++ b/ext/IntervalArithmeticForwardDiffExt.jl
@@ -59,7 +59,7 @@ end
 function Base.:(^)(x::Dual{<:Tx}, y::ExactReal) where Tx
     v = value(x)
     expv = v^y
-    if iszero(y) || all(iszero.(values(partials(x))))
+    if iszero(y.value) || all(iszero.(values(partials(x))))
         new_partials = zero(partials(x))
     else
         new_partials = partials(x) * y * v^(y - 1)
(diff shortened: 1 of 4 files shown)


src/intervals/exact_literals.jl
```diff
@@ -53,6 +53,8 @@ struct ExactReal{T<:Real} <: Real
     ExactReal(value::T) where {T<:Real} = new{T}(value)
 end
 
+Base.iszero(x::ExactReal) = iszero(x.value)
```
OlivierHnt, line 56: Maybe we should not hook too much into Base interface for `ExactReal` for now; so I would not define this function.
It may be safe to do so, but I think it is worth being very conservative as much as possible at the moment since `ExactReal` is relatively new.

ext/IntervalArithmeticForwardDiffExt.jl
```diff
@@ -56,4 +56,22 @@ function Base.:(^)(x::ExactReal, y::Dual{<:Any, I}) where I<:Interval
     return convert(I, x)^y
 end
 
+function Base.:(^)(x::Dual{<:Tx}, y::ExactReal) where Tx
+    v = value(x)
+    expv = v^y
+    if iszero(y) || all(iszero.(values(partials(x))))
+        new_partials = zero(partials(x))
+    else
+        new_partials = partials(x) * y * v^(y - 1)
```
OlivierHnt, line 65: I have not tried this PR yet, but just looking at the code I would have expected that we need `v^(y - interval(1))`.
In which case, we should have a test that reflects this (maybe if `y = ExactReal(pi)` this is relevant?)

ext/IntervalArithmeticForwardDiffExt.jl
```diff
@@ -56,4 +56,22 @@ function Base.:(^)(x::ExactReal, y::Dual{<:Any, I}) where I<:Interval
...
+    end
+    return Dual{Tx}(expv, new_partials)
+end
+
+function Base.:(^)(x::ExactReal, y::Dual{<:Ty}) where Ty
+    v = value(y)
+    expv = x^v
+    deriv = (iszero(x) && inf(v) > 0) ? zero(expv) : expv*log(x)
```
OlivierHnt, line 73: Similar comment here: I would have expected `expo * log(interval(x))`, but maybe I am missing something

ext/IntervalArithmeticForwardDiffExt.jl
Kolaru: The interval functions were previously implemented and work fine. The two new functions are for mixing `ExactReal` with non-interval `Real` numbers, so no conversion to interval should be added.

The default ForwardDiff implementation uses `x == zero(x)` to check the exponent, and is thus failing with `ExactReal`s, just like it was failing with `Interval`s.

src/intervals/exact_literals.jl
Kolaru: Removed.

ext/IntervalArithmeticForwardDiffExt.jl
OlivierHnt: You are right, but `ExactReal` should propagate though, no?
Otherwise `v^(y - 1)` would give a NG flag despite `v` being an interval and `y` an ExactReal; maybe `new_partials = partials(x) * y * v^(y - convert(typeof(v), 1))` (or something in this vein)?

Kolaru: If you mix intervals and `ExactReal`s, you hit the definition at lines 51-57, and the `ExactReal` is immediately converted to an interval.

The new functions have nothing to do with intervals, it is a fix around the fact that ForwardDiff decided to use branching in the derivatives of powers, which also breaks `ExactReal`s.

OlivierHnt:
Thx for addressing my questions @Kolaru :) LGTM

PR #660, opened 2024-05-28
Remove inner constructor for ExactReal

dpsanders:
This was preventing

```
ExactReal{Int64}(3)
```
which should be allowed.

### DIFF
--- a/src/intervals/exact_literals.jl
+++ b/src/intervals/exact_literals.jl
@@ -49,8 +49,6 @@ See also: [`@exact`](@ref).
 """
 struct ExactReal{T<:Real} <: Real
     value :: T
-
-    ExactReal(value::T) where {T<:Real} = new{T}(value)
 end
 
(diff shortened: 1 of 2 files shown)


OlivierHnt:
I think this was done on purpose to avoid silent errors, e.g.
```Julia
julia> ExactReal{Float64}(1//3)
ExactReal{Float64}(0.3333333333333333)
```

Kolaru:
I think David's example should indeed be allowed, and Olivier's forbidden.

To make it so, we can add a second inner constructor:
```julia
struct ExactReal{T<:Real} <: Real
    value :: T

    ExactReal{T}(value::T) where {T<:Real} = new{T}(value)
    ExactReal(value::T) where {T<:Real} = new{T}(value)
end
```

dpsanders:
Thanks @Kolaru, I added that inner constructor instead.

Kolaru:
Quickly added simple tests, and LGTM, thanks :)

PR #663, opened 2024-05-30
Speed up hull

dpsanders:


OlivierHnt:
In https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/664 I added a feature of this PR: the resulting decoration does not depend on `decoration(r)`.

In the current status of this PR, I think this is the only small improvement we can get.
Unless you are still trying things out, I suggest to close this PR.

dpsanders:
What does removing  `isempty_interval(x) & isempty_interval(y) && return x` break?

dpsanders:
Oh, I see what you're saying :/ That... is unfortunate. Closing for now, thanks.

PR #666, opened 2024-06-02
Restore partial support for `:fast` rounding mode

### LINKED ISSUE #665: Reinstate fast rounding with `prevfloat` and `nextfloat`
The IEEE 758 standard for floating-point arithmetic guarantees that `+`, `-`, `*`, `/` and `sqrt` are faithfully rounded.

Thus directed rounding with `prevfloat` and `nextfloat` does give a guaranteed enclosure for interval arithmetic computations using those operations.

In particular this is enough for fast guaranteed enclosure of the range of polynomials.

OlivierHnt:
This PR essentially uncomments the old code we had for `+`, `-`, `*`, `/`, `inv`, `sqrt` with `:fast` rounding.
Since Julia should be compliant with the IEEE specifications, these functions can only return a result within one ulp of the correct one.
Maybe we should find a reference in the Julia repo where this is mentioned?

Closes #665

cc-@dpsanders

### DIFF
--- a/src/intervals/rounding.jl
+++ b/src/intervals/rounding.jl
@@ -4,9 +4,11 @@
 Interval rounding type.
 
 Available rounding types:
-- `:fast` (unsupported): rounding via `prevfloat` and `nextfloat`.
+- `:fast`: rounding via `prevfloat` and `nextfloat`. Currently only the functions
+`+`, `-`, `*`, `/`, `inv` and `sqrt` are supported.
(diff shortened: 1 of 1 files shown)


dpsanders:
LGTM, thanks!

Are there tests for this?

dpsanders:
I don't think this closes #592 , since CRlibm is used for `sin`, `exp` etc.

OlivierHnt:
So CORE-MATH would be faster than CRlibm?

Concerning the tests: I am not sure if there ever were any. I could add some simple ones.

PR #667, opened 2024-06-02
Base utilities for exact reals

dpsanders:
Defines `Base.hash` and `Base.isfinite` for `ExactReal`s.

This is necessary so that they work with Symbolics.jl.


Kolaru:
Do you think it would be good to have tests checking the compatibility with Symbolics ? (I think you are probably the main user of Symbolics + IA, so I would say it's up to you)

dpsanders:
Yes I think that would be great. I'm just not sure how to do that (a test-only dependency).

OlivierHnt:
There is a Project.toml file in the test folder. You could `pkg> add Symbolics` to it.

OlivierHnt:
Btw you should rebase this branch

PR #677, opened 2024-08-23
add IntervalSets conversions

aplavin:


ext/IntervalArithmeticsIntervalSetsExt.jl
```diff
@@ -0,0 +1,9 @@
+module IntervalArithmeticsIntervalSetsExt
+
+import IntervalSets as IS
+import IntervalArithmetic as IA
+
+Base.convert(T::Type{<:IS.Interval}, i::IA.Interval) = T(IA.inf(i), IA.sup(i))
+Base.convert(T::Type{<:IA.Interval}, i::IS.Interval) = T(IS.infinum(i), IS.supremum(i))
```
OlivierHnt, line 7: This won't work.
It should probably be in the same vein as how we deal with `Real` number conversion, i.e.

```Julia
function Base.convert(::Type{Interval{T}}, i::IS.Interval) where {T<:NumTypes}
    x = IA.interval(T, IS.infinum(i), IS.supremum(i))
    return IA._unsafe_interval(IA.bareinterval(x), IA.decoration(x), false)
end
```

OlivierHnt:
Thanks for opening this PR. Looks reasonable to have this extension 🙂.

OlivierHnt:
Mhmm I am not sure we want to have a conversion between `IntervalSet.Interval` and `IntervalArithmetic.BareInterval` (eg we do not allow conversion between `IntervalArithmetic.BareInterval` and `Real`).
I know nothing about IntervalSet.jl, can you perform any sort of arithmetic operations on `IntervalSet.Interval`? E.g if `x = IntervalSet.Interval(1, 2)`, then is `x + 1` or `x * 3.1` or `x + x` allowed?

PS: as mentioned before I am not opposed to this PR, it seems very reasonable for interval libraries to communicate. Though I am curious 🙂, is there any specific reason that motivated opening the PR?

ext/IntervalArithmeticsIntervalSetsExt.jl
```diff
@@ -0,0 +1,27 @@
...
+# BareInterval <- IS.Interval
+Base.convert(::Type{IA.BareInterval}, i::IS.Interval) = IA.bareinterval(IS.endpoints(i)...)
+IA.bareinterval(i::IS.Interval) = IA.bareinterval(IS.endpoints(i)...)
+
+# BareInterval -> IS.Interval
+Base.convert(::Type{IS.Interval}, i::IA.BareInterval) = IS.Interval(IA.inf(i), IA.sup(i))
+IS.Interval(i::IA.BareInterval) = IS.Interval(IA.inf(i), IA.sup(i))
+
+# Interval <- IS.Interval
+Base.convert(::Type{IA.Interval}, i::IS.Interval{<:Any,<:Any,T}) where {T} = convert(IA.Interval{float(T)}, i)
```
OlivierHnt, line 16: Looking at this line, it seems that `IS.Interval` can have bounds that are of different real types (eg `Rational`, or even maybe `Int`)?
In our case, `IA.Interval` can have bounds that are either `Rational` or `AbstractFloat`.
Hence, if you would like to preserve as much as possible the same bound type as the initial interval, I suggest the following:
```Julia
Base.convert(::Type{IA.Interval}, I::IS.Interval{<:Any,<:Any,T}) where {T} = convert(IA.Interval{IA.promote_numtype(T, T)}, i)
```

OlivierHnt:
Yes it affects the `BareInterval` advice. Long story short, we disallowed conversion between `Real` and `BareInterval` because one could perform non-validated numerics and obtain some `Real` (typically I mean a float here) and then convert it to an interval type (whose bounds are therefore not guaranteed to contain the correct answer).
The `Interval` type in IntervalArithmetic.jl has a special "guaranteed" flag to track this stuff, so conversion is allowed but the resulting interval has a special tag.

So if an `IntervalSets.Interval` can never be touched by any non-validated numerics (*all defined operations are rounding errors free*) then I think it is not a problem to have a conversion to `BareInterval`.

Then you could implement something like
```Julia
# BareInterval <- IS.Interval
IA.bareinterval(i::IS.Interval) = IA.bareinterval(IS.endpoints(i)...)
IA.bareinterval(::Type{T}, i::IS.Interval) where {T<:IA.NumTypes} = IA.bareinterval(T, IS.endpoints(i)...)
Base.convert(::Type{IA.BareInterval}, i::IS.Interval) = IA.bareinterval(i)
Base.convert(::Type{IA.BareInterval{T}}, i::IS.Interval) where {T<:IA.NumTypes} = IA.bareinterval(T, i)

# BareInterval -> IS.Interval
IS.Interval(i::IA.BareInterval) = IS.Interval(IA.inf(i), IA.sup(i))
# maybe also IS.Interval{T,S,R}(i::IA.BareInterval) where {T,S,R} = IS.Interval{T,S,R}(IA.inf(i), IA.sup(i))
Base.convert(::Type{IS.Interval}, i::IA.BareInterval) = IS.Interval(i)
# maybe you want to add a conversion to a more specific signature, eg Base.convert(::Type{T}, i::IA.BareInterval) where {T<:IS.BareInterval} = T(i)

# Interval <- IS.Interval
IA.interval(i::IS.Interval) = IA.interval(IS.endpoints(i)...)
IA.interval(::Type{T}, i::IS.Interval) where {T<:IA.NumTypes} = IA.interval(T, IS.endpoints(i)...)
Base.convert(::Type{IA.Interval}, i::IS.Interval) = IA.interval(i)
Base.convert(::Type{IA.Interval{T}}, i::IS.Interval) where {T<:IA.NumTypes} = IA.interval(T, i)

# Interval -> IS.Interval
IS.Interval(i::IA.Interval) = IS.Interval(IA.inf(i), IA.sup(i))
# maybe also IS.Interval{T,S,R}(i::IA.Interval) where {T,S,R} = IS.Interval{T,S,R}(IA.inf(i), IA.sup(i))
Base.convert(::Type{IS.Interval}, i::IA.Interval) = IS.Interval(i)
# maybe you want to add a conversion to a more specific signature, eg Base.convert(::Type{T}, i::IA.Interval) where {T<:IS.Interval} = T(i)
```

OlivierHnt:
A more perverse potential issue: does `convert(::Type{<:IS.Interval}, x::Number)` exists?

OlivierHnt:
Ok I looked at IntervalSets.jl. I found two issues (from the point of view of IntervalArithmetic.jl):

```Julia
julia> using IntervalSets

julia> Interval(1e-120, 1e-119) # non trivial interval set
1.0e-120 .. 1.0e-119

julia> ClosedInterval{Float32}(1e-120 .. 1e-119) # issue #1: conversion did not widen the interval set
0.0 .. 0.0

julia> 1 ± 1e-120 # issue #2: the interval only contains 1
1.0 .. 1.0
```

If these are intentional, then we should not convert to `BareInterval` and implement

```Julia
# Interval <- IS.Interval
function IA.interval(i::IS.Interval)
    x = IA.interval(endpoints(i)...)
    return IA._unsafe_interval(IA.bareinterval(x), IA.decoration(x), false)
end
function IA.interval(::Type{T}, i::IS.Interval) where {T<:IA.NumTypes}
    x = IA.interval(T, endpoints(i)...)
    return IA._unsafe_interval(IA.bareinterval(x), IA.decoration(x), false)
end
Base.convert(::Type{IA.Interval}, i::IS.Interval) = IA.interval(i)
Base.convert(::Type{IA.Interval{T}}, i::IS.Interval) where {T<:IA.NumTypes} = IA.interval(T, i)

# Interval -> IS.Interval
IS.Interval(i::IA.Interval) = IS.Interval(IA.inf(i), IA.sup(i))
# maybe also IS.Interval{T,S,R}(i::IA.Interval) where {T,S,R} = IS.Interval{T,S,R}(IA.inf(i), IA.sup(i))
Base.convert(::Type{IS.Interval}, i::IA.Interval) = IS.Interval(i)
# maybe you want to add a conversion to a more specific signature, eg Base.convert(::Type{T}, i::IA.Interval) where {T<:IS.Interval} = T(i)
```

OlivierHnt:
LGTM from my side 👍

@lbenet, @Kolaru how do you feel about this PR?

Kolaru:
It looks very reasonable to me.

There is just one edge case: IA intervals are always closed, except when one of the bounds is infinity, then the interval is always open on the side of the infinity.

So accordingly, `IA.Interval(1, Inf)` should convert to `IS.Interval{:closed, :open}(1, Inf)`.

OlivierHnt:
Indeed. Also this raises the question for NaI and empty intervals; @aplavin do you have a special signature for "empty intervals"?

OlivierHnt:
Yes sadly.

So I am a bit puzzled, with IntervalSets you can construct `Interval(2, 1)`. Is this supposed to be equivalent to `Interval(1, 2)`?

OlivierHnt:
Mhmm yeah. What differs when using `IS.Interval{:closed,:open}(1, Inf)` vs `IS.Interval{:closed,:closed}(1, Inf)`?
If nothing changes, then maybe we do not need to return an NaI (Not an Interval).

OlivierHnt:
Sorry I was traveling. Merging now 🙂

PR #686, opened 2024-10-10
Extend conversion

### LINKED ISSUE #685: Conversion issue when combining `StaticArrays`, `ForwardDif`, and `IntervalArithmetic`
In trying to combine `ForwardDiff` with `StaticArrays` and `IntervalArithmetic` I ran into the following issue:

```julia
using StaticArrays, ForwardDiff, IntervalArithmetic

x  = SVector(1.0, 2.0)
xI = SVector(interval(0.0, 1.0), interval(0.0, 1.0))
g  = (x) -> sum(insert(x, 1, 1.0))
∇g = x -> ForwardDiff.gradient(g, x)
∇g(x) # OK
∇g(xI) # error
```

I did not look too deep, but it seems that somewhere in the conversion pipeline `StaticArrays` calls a conversion method, which in turn calls the constructor `Interval{Float64}(x)`. I saw that this constructor was intentionally commented out

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/429ad1634c9417c4919946ced8c9e251a201ae00/src/intervals/construction.jl#L371

Uncommenting the line above does "solve" the issue, but the `IntervalArithmetic` tests fail because it is explicitly tested that `Interval{T}(x)` should throw

https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/429ad1634c9417c4919946ced8c9e251a201ae00/test/interval_tests/construction.jl#L47

So I see that this is intentional. I was wondering if there is a workaround? 

Thanks.

OlivierHnt:
This PR re-introduce `Interval(::Real)` and `Interval{T}(::Real)`, marking the output with the "NG flag".

It also adds a conversion in the ForwardDiff extension between `ExactReal` and `Dual`.

Closes #685

### DIFF
--- a/ext/IntervalArithmeticForwardDiffExt.jl
+++ b/ext/IntervalArithmeticForwardDiffExt.jl
@@ -1,7 +1,7 @@
 module IntervalArithmeticForwardDiffExt
 
 using IntervalArithmetic, ForwardDiff
-using ForwardDiff: Dual, ≺, value, partials
+using ForwardDiff: Dual, Partials, ≺, value, partials
 
 #
 
@@ -69,5 +69,9 @@ function Base.:(^)(x::ExactReal, y::Dual{<:Ty}) where {Ty}
         return Dual{Ty}(expv, expv * log(x) * partials(y))
     end
 end
+
+# resolve ambiguity
+
+Base.convert(::Type{Dual{T,V,N}}, x::ExactReal) where {T,V,N} = Dual{T}(V(x), zero(Partials{N,V}))
 
 end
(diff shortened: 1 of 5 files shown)


OlivierHnt:
Previously broken tests now pass.

lbenet:
> This PR re-introduce Interval(::Real) and Interval{T}(::Real), marking the output with the "NG flag".

Do we really want/need this? We have discussed that, indeed without it becomes *unconfortable*, but also imposes being carefull wrt silent errors. 

Adding the conversion in the ForwardDiff extension is really appreciated!

OlivierHnt:
Yes we should be cautious.
I could be missing something, but at the time we removed `Interval(x::Real)` and `Interval{T}(x::Real)`, we did not have the "NG flag" mechanism right? Since we currently allow `convert(Interval{Float64}, 1)` maybe it makes sense to have `Interval{Float64}(1.0)`?

OlivierHnt:
So what should we do with this PR? Is reintroducing `Interval(x)` and `Interval{Float64}(x)` in the 0.22 release reasonable if we use the "NG flag" (so same behaviour as `convert(Interval{Float64}, x)`)?

OlivierHnt:
I will go ahead and merge this PR. If one has strong feelings against it, we can revert it before the next release.

Issue #655, opened 2024-05-26
Piecewise functions


dpsanders:
Here's a possible implementation for piecewise functions:

```jl
using IntervalArithmetic
using IntervalArithmetic.Symbols # to use the constructor `..`

# represents a piecewise 1D function
struct Piecewise{T <: Tuple}
    pieces::T 
end

Piecewise(pairs...) = Piecewise(pairs)

function (piecewise::Piecewise)(X::Interval)
    return reduce(hull, f(intersect_interval(X, region)) for (region, f) ∈ piecewise.pieces)
end

f = Piecewise(
        0..3 => x -> x + interval(1),
        3..6 => identity,
        6..Inf => x -> x + interval(2)
)

f(2..5)
```

Then Heaviside should be definable with
```jl
H = Piecewise(-Inf..0 => zero, 0..Inf => one)
```
but that is not working for me right now 🤔  E.g. it gives
```jl
julia> H(1..1)
[0.0, 1.0]_trv
```
whereas it should give `[1, 1]`.

dpsanders:
Ah that's because 
```jl
julia> zero(emptyinterval())
[0.0, 0.0]_com
```

I have an idea for how to make a default 0 value.

dpsanders:
This version works:
```jl
function (piecewise::Piecewise)(X::Interval)
    return reduce(hull, begin
                            val = intersect_interval(X, region)
                            if isempty_interval(val)
                                val
                            else
                                f(val)
                            end
                        end
                    for (region, f) ∈ piecewise.pieces)
end

H = Piecewise(-Inf..0 => zero, 0..Inf => one)

```

dpsanders:
It would probably be worth having a `PiecewiseConstant` too.

dpsanders:
We also need to add the decoration of course.

OlivierHnt:
This issue seems to supersedes #653 and #654 no?

Also, with such a feature, implementing the Heaviside function would be even simpler than it already is so I am not sure it would warrant being defined inside the library.

Issue #661, opened 2024-05-29
Expose domain of functions


dpsanders:
Currently the user has no way of knowing which domain a function is defined over. 

E.g.
```
domain(sqrt) = 0..Inf
```
or `domain_interval(sqrt)`.

OlivierHnt:
I do not think we should do this. Sometimes the domain is not closed, eg for `log`.

However I think this should appear in the docstring of `sqrt`.

dpsanders:
Good point about the domain for `log`. 
Nonetheless I think there ideally should be *some* way for the user to obtain this information programatically.

OlivierHnt:
In the end, this is truly a math question and not interval related.
How the domain relates to intervals (which is what matters for the library) depends on the function (what we do for `sqrt` is different than what we do for other functions).

That being said, we can leave this issue open for a little while and see if someone comes up with something.
But I suspect that this is a matter of documentation.

dpsanders:
The reason I was thinking about this is that I wanted to do something like `Piecewise(sqrt, extend=0)` meaning "extend sqrt with 0 outside its domain of definition", but that requires programmatic access to the domain.

OlivierHnt:
Oh I see, interesting!

Then `Piecewise(log, extend = 0)` would also work I think, or even `Piecewise(asec; extend = 0)`.
The logic for this feature is more like a `isindomain` function

Something in the vein of
```
isindomain(::typeof(sqrt), x::Float64) = 0 <= x

isindomain(::typeof(sqrt), x::Interval) = issubset_interval(x, interval(0, Inf))
```

OlivierHnt:
Btw since PR #689 was merged last year, we have a `Domain` type which can handle closed / open intervals.
So I believe we have enough to implement your suggestion:

```Julia
Piecewise(sqrt; extension = x -> 0) # extend `sqrt` by 0 beyond its domain of definition.
```

Issue #424, opened 2020-11-17
`setdiff` returns a disjoint representation, whereas `union` does not


goretkin:
This strikes me as inconsistent:

```julia
julia> union(interval(0, 1), interval(2, 3))
[0, 3]

julia> setdiff(interval(0, 3), interval(1,2))
2-element Array{Interval{Float64},1}:
 [0, 1]
 [2, 3]
```

dpsanders:
It's true that in some sense it's inconsistent, but it's what is useful.
`union` here really means "interval hull". (There is also the synonym `hull`.)

If you want `union` to return a vector of intervals, that's a whole different world (which I have coded somewhere).

goretkin:
I suppose there are two separate issues, one is consistency in which some operations have an implicit "hull" and others do not.

The other issue is about type closure. It would be good if e.g. (even assuming `b` and `c` are overlapping to remove that question)

`setdiff(a, union(b,c)) == setdiff(setdiff(a, b), c))`

but since `setdiff(a, b)` returns something that itself cannot carry the semantics of a union of intervals (Unless everywhere `Array{Interval}` means that, which is almost certainly not possible).

So probably `setdiff` should return an object that can only mean "union of intervals".

dpsanders:
Yes, that's a good idea.

PR #689, opened 2024-11-28
Implement piecewise function

Kolaru:
Implement piecewise function, as described in #655 (and handling decorations).

For example, the `abs` function would be defined as

```julia
myabs = Piecewise(
    interval(-Inf, 0) => x -> -x,
    interval(0, Inf) => identity ;
    continuous = true
)
```

It also introduce the `Constant` constructor, as a workaround for `Returns` from Base: the later can not convert its output to interval when the input is an interval, which causes problems.

For example:

```julia
julia> f = Returns(1.0)
Returns{Float64}(1.0)

julia> f(22.2)
1.0

julia> f(interval(22.2))
1.0

julia> g = Constant(1.0)
Constant{Float64}(1.0)

julia> g(22.2)
1.0

julia> g(interval(22.2))
[1.0, 1.0]_com
```

It can be used for example to define a step function as

```julia
step = Piecewise(
    interval(-Inf, 0) => Constant(0),
    interval(0, Inf) => Constant(1)
)
```

### DIFF
--- a/src/IntervalArithmetic.jl
+++ b/src/IntervalArithmetic.jl
@@ -23,6 +23,11 @@ const RealIntervalType{T} = Union{BareInterval{T},Interval{T}}
 
 #
 
+include("piecewise.jl")
+    export Constant, Piecewise
+
+#
+
 include("display.jl")
     export setdisplay
 
--- a/src/piecewise.jl
+++ b/src/piecewise.jl
@@ -20,6 +20,7 @@ intersecting(X::Interval, Y::Interval) = !isempty_interval(intersect_interval(X,
 function (piecewise::Piecewise)(X::Interval)
     intersections = intersecting.(X, subdomains(piecewise))
     dom = domain(piecewise)
+    !intersecting(X, dom) && return emptyinterval()
     if !issubset_interval(X, dom)
         dec = trv
     elseif count(intersections) > 1
@@ -35,11 +36,8 @@ function (piecewise::Piecewise)(X::Interval)
     used_pieces = filter(piece -> intersecting(X, piece[1]), piecewise.pieces)
     outputs = map(used_pieces) do (region, f)
         S = IntervalArithmetic.setdecoration(intersect_interval(X, region), decoration(X))
-        @show S
         return f(S)
     end
-
-    @show outputs
 
     dec = min(dec, minimum(decoration.(outputs)))
     return IntervalArithmetic.setdecoration(reduce(hull, outputs), dec)
--- a/src/piecewise.jl
+++ b/src/piecewise.jl
@@ -1,0 +1,54 @@
+using IntervalArithmetic
+
+# represents a piecewise 1D function
+struct Piecewise{T <: Tuple}
+    pieces::T 
+    continuous::Bool
+end
+
+Piecewise(pairs... ; continuous = false) = Piecewise(pairs, continuous)
+
+domain(piecewise::Piecewise) = reduce(hull, subdomains(piecewise))
+subdomains(piecewise::Piecewise) = first.(piecewise.pieces)
+
+intersecting(X::Interval, Y::Interval) = !isempty_interval(intersect_interval(X, Y))
+
+function (piecewise::Piecewise)(X::Interval)
+    intersections = intersecting.(X, subdomains(piecewise))
+    dom = domain(piecewise)
+    if !issubset_interval(X, dom)
+        dec = trv
+    elseif count(intersections) > 1
+        if piecewise.continuous
+            dec = dac
+        else
+            dec = def
+        end
+    else
+        dec = com
+    end
+
+    used_pieces = filter(piece -> intersecting(X, piece[1]), piecewise.pieces)
+    outputs = map(used_pieces) do (region, f)
+        S = IntervalArithmetic.setdecoration(intersect_interval(X, region), decoration(X))
+        return f(S)
+    end
+
+    dec = min(dec, minimum(decoration.(outputs)))
+    return IntervalArithmetic.setdecoration(reduce(hull, outputs), dec)
+end
+
+function (piecewise::Piecewise)(x::Real)
+    for (region, f) in piecewise.pieces
+        in_interval(x, region) && return f(x)
+    end
+    throw(DomainError("piecewise function was called with $x which is outside of its domain $(domain(piecewise))"))
+end
+
+f = Piecewise(
+        0..3 => x -> x + 1,
+        3..6 => identity,
+        6..Inf => x -> x + 2
+)
+
+f(2..5)
--- a/src/piecewise.jl
+++ b/src/piecewise.jl
@@ -1,6 +1,10 @@
-using IntervalArithmetic
+struct Constant{T}
+    value::T
+end
 
-# represents a piecewise 1D function
+(constant::Constant)(::Any) = constant.value
+(constant::Constant)(::Interval) = interval(constant.value)
+
 struct Piecewise{T <: Tuple}
     pieces::T 
     continuous::Bool
@@ -31,8 +35,11 @@ function (piecewise::Piecewise)(X::Interval)
     used_pieces = filter(piece -> intersecting(X, piece[1]), piecewise.pieces)
     outputs = map(used_pieces) do (region, f)
         S = IntervalArithmetic.setdecoration(intersect_interval(X, region), decoration(X))
+        @show S
         return f(S)
     end
+
+    @show outputs
 
     dec = min(dec, minimum(decoration.(outputs)))
     return IntervalArithmetic.setdecoration(reduce(hull, outputs), dec)
@@ -44,11 +51,3 @@ function (piecewise::Piecewise)(x::Real)
     end
     throw(DomainError("piecewise function was called with $x which is outside of its domain $(domain(piecewise))"))
 end
-
-f = Piecewise(
-        0..3 => x -> x + 1,
-        3..6 => identity,
-        6..Inf => x -> x + 2
-)
-
-f(2..5)
--- a/test/interval_tests/piecewise.jl
+++ b/test/interval_tests/piecewise.jl
@@ -1,0 +1,32 @@
+@testset "Step function" begin
+    step = Piecewise(
+        interval(-Inf, 0) => Constant(0),
+        interval(0, Inf) => Constant(1)
+    )
+
+    @test step(-1) == 0
+    @test step(100) == 1
+    @test isequal_interval(step(interval(-3.2, -2.1)), interval(0))
+    @test decoration(step(interval(-3.33))) == com
+    @test isequal_interval(step(interval(2.3, 3.4)), interval(1))
+    @test decoration(step(interval(4.44))) == com
+    @test isequal_interval(step(interval(-22.2, 33.3)), interval(0, 1))
+    @test decoration(step(interval(-11, 11))) == def
+end
+
+@testset "abs" begin
+    myabs = Piecewise(
+        interval(-Inf, 0) => x -> -x,
+        interval(0, Inf) => identity ;
+        continuous = true
+    )
+
+    @test myabs(-1) == 1
+    @test myabs(100) == 100
+    @test isequal_interval(myabs(interval(-3.2, -2.1)), interval(2.1, 3.2))
+    @test decoration(myabs(interval(-3.33))) == com
+    @test isequal_interval(myabs(interval(2.3, 3.4)), interval(2.3, 3.4))
+    @test decoration(myabs(interval(4.444))) == com
+    @test isequal_interval(myabs(interval(-22.2, 33.3)), interval(0, 33.3))
+    @test decoration(myabs(interval(-11, 11))) == dac
+end
--- a/test/interval_tests/piecewise.jl
+++ b/test/interval_tests/piecewise.jl
@@ -30,3 +30,16 @@ end
     @test isequal_interval(myabs(interval(-22.2, 33.3)), interval(0, 33.3))
     @test decoration(myabs(interval(-11, 11))) == dac
 end
+
+@testset "Out of domain" begin
+    window = Piecewise(
+        interval(-π, π) => x -> 1/2 * (cos(x) + 1)
+    )
+
+    @test_throws DomainError window(123)
+    @test isequal_interval(window(interval(0, π)), interval(0, 1))
+    @test decoration(window(interval(-π, 0))) == com
+    @test isequal_interval(window(interval(-10, 10)), interval(0, 1))
+    @test decoration(window(interval(-10, 10))) == trv
+    @test isempty_interval(window(interval(100, 1000)))
+end


OlivierHnt:
Oh nice, looks great!

One thing is that there should maybe be a check to prevent overlapping domains. Eg the two following functions do not behave the same, and are essentially ill-defined

```Julia
julia> foo = Piecewise(
                  interval(-Inf, 0) => x -> -x,
                  interval(-1, 1) => x -> 2x,
                  interval(0, Inf) => x -> x ;
                  continuous = true
              )

julia> goo = Piecewise(
                  interval(-Inf, 0) => x -> -x,
                  interval(0, Inf) => x -> x,
                  interval(-1, 1) => x -> 2x ;
                  continuous = true
              )
```

Moreover, `domain` should probably not be a convex  hull, maybe a vector of interval (or a union type, related to #424 in some sense)? I also wonder if the resulting interval need to always have the `trv` decoration. Eg

```Julia
Julia> julia> f = Piecewise(
                         interval(-1, 0) => identity, 
                         interval(1, 2) => identity);

julia> domain(f)
[-1.0, 2.0]_trv

Julia> julia> g = Piecewise(
                         interval(-1, 0) => identity, 
                         interval(0, 1) => identity);

julia> domain(g) # should it have the decoration `trv`?
[-1.0, 2.0]_trv
```

One question on the implementation. Currently you can only indicate if a `Piecewise` function is continuous or not. This may be mitigated by creating nested `Piecewise` functions... but what would you think about 

```Julia
struct Piecewise2{N,T<:NTuple{N}}
    pieces :: T 
    continuous :: NTuple{N,Bool}
end

Piecewise2(pieces::NTuple{N}, continuous::Bool) where {N} = Piecewise2(pieces, ntuple(_ -> continuous, Val(N)))
```
This way you do not hit a `def` decoration when evaluating the piecewise function at continuous points.

src/piecewise.jl
```diff
@@ -0,0 +1,51 @@
...
+        return f(S)
+    end
+
+    dec = min(dec, minimum(decoration.(outputs)))
+    return IntervalArithmetic.setdecoration(reduce(hull, outputs), dec)
+end
+
+function (piecewise::Piecewise)(x::Real)
+    for (region, f) in piecewise.pieces
+        in_interval(x, region) && return f(x)
```
dpsanders, line 48: This doesn't seem completely right: you need to take the union of `f(x)` over all the pieces that `x` overlaps with, I think.

dpsanders: e.g. the line
```
    return reduce(hull, f(intersect_interval(X, region)) for (region, f) ∈ piecewise.pieces)

```
in my original implementation in the issue you limked to.

Kolaru: This part is the implementation for non-interval real only, so I think it is fine, as it would be a bit awkward to return an interval for non-interval input.

As @OlivierHnt mentioned, this implementation allows ill-defined functions, but I don't think the answer is the interval-ification of everything.

Kolaru:
> One thing is that there should maybe be a check to prevent overlapping domains. Eg the two following functions do not behave the same, and are essentially ill-defined

Yes, that reasonable.

To do that properly, we would need open and closed intervals (otherwise the boundary points are always included, and thus points on the boundary of pieces are always duplicated).

I am wondering if it would be worth to pull intervals from `Intervals.jl` to do this.

> Moreover, `domain` should probably not be a convex hull, maybe a vector of interval (or a union type, related to #424 in some sense)? I also wonder if the resulting interval need to always have the trv decoration. Eg

Actually yes, here you would have a bug if there is a hole in the middle of the pieces.

For the decoration, I don't know. Those intervals, really have not much in common with the arithmetic intervals...

> This way you do not hit a def decoration when evaluating the piecewise function at continuous points.

Well, I thought about this. I am not strongly against implementing it. It is just a bit awkward that the order of the pieces matter in the input.

OlivierHnt:
Ah I did not notice the behaviour when evaluating at the boundary point. So it is also order dependent

```Julia
julia> foo = Piecewise(
           interval(-Inf, 0) => x -> -x,
           interval(0, Inf) => x -> 3 ;
           continuous = false
       );

julia> foo(0)
0

julia> goo = Piecewise(
           interval(0, Inf) => x -> 3,
           interval(-Inf, 0) => x -> -x;
           continuous = false
       );

julia> goo(0)
3
```

My first impulse would be not to add a dependency on Intervals.jl. One idea could be to introduce yet an other option to specify the behaviour. Eg `Piecewise(...; continuous = (false, boundary_behaviour))`, where `boundary_behaviour::Symbol` could be `:left`, `:right`, `:average` (maybe the latter ought to be the default).

> Actually yes, here you would have a bug if there is a hole in the middle of the pieces.

Right, so for `f = Piecewise(interval(0, 1) => x -> 1, interval(10, 11) => x -> 2)`, I would expect `domain(f)` to return something in the vein of `(interval(0, 1, trv), interval(10, 11, trv))`.

> For the decoration, I don't know. Those intervals, really have not much in common with the arithmetic intervals...

Yes you are right, maybe `trv` is appropriate after all.

> I thought about this a bit, I am not strongly against implementing it. It is just a bit awkward that the order of the pieces matter in the input.

Indeed, but maybe it is fair to expect it:
- if you pass a single boolean then the order is irrelevant (you prescribe the same behaviour for all boundary points);
- if you pass a list of boolean then these instructions must match the ordering of the boundary points

OlivierHnt:
Mhm I gave it a bit more thoughts. Here is an attempt (order independent)

```Julia
struct Piecewise{T<:Tuple}
    pieces :: T
    # The symbol indicates whether or not the boudnary is included in the interval:
    # Symbol(11) means that the interval contains both extremities
    # Symbol(10) means that the interval contains only at the left extremity
    # Symbol(01) means that the interval contains only at the right extremity
    # Symbol(00) means that the interval does not contain either extremity
    # TODO: need to normalize `pieces`
    # TODO: missing `continuous` (to determine the decoration `def`, or `com`)
end

Piecewise(pieces...) = Piecewise(pieces)

_subdomains(piecewise::Piecewise) = first.(piecewise.pieces)

function _intersect_subdomain(x::Interval, region::Tuple{Interval,Symbol})
    y = IntervalArithmetic.setdecoration(intersect_interval(x, region[1]), decoration(x))
    if isempty_interval(y)
        return y
    elseif region[2] === Symbol(11)
        return y
    elseif region[2] === Symbol(01)
        isthin(y, inf(region[1])) && return emptyinterval(y)
        return y
    elseif region[2] === Symbol(10)
        isthin(y, sup(region[1])) && return emptyinterval(y)
        return y
    else # region[2] === Symbol(00)
        isthin(y, inf(region[1])) | isthin(y, sup(region[1])) && return emptyinterval(y)
        return y
    end
end

function (piecewise::Piecewise)(x::Interval)
    subdoms = _subdomains(piecewise)
    vals = _intersect_subdomain.(x, subdoms)
    all(v -> isempty_interval(v[1]), vals) && return throw(DomainError("piecewise function was called with $x which is outside of its domain $subdoms"))
    return _evaluate(piecewise.pieces, vals, nothing)
end

function _evaluate(p, x, r) # maybe not type stable
    isempty_interval(x[1][1]) && return _evaluate(Base.tail(p), Base.tail(x), r)
    fx = last(p[1])(x[1][1])
    return _evaluate(Base.tail(p), Base.tail(x), IntervalArithmetic.setdecoration(hull(fx, r), min(decoration(fx), decoration(r))))
end

function _evaluate(p, x, ::Nothing) # maybe not type stable
    isempty_interval(x[1][1]) && return _evaluate(Base.tail(p), Base.tail(x), nothing)
    fx = last(p[1])(x[1][1])
    return _evaluate(Base.tail(p), Base.tail(x), fx)
end

function _evaluate(p::Tuple{Pair}, x, r) # maybe not type stable
    isempty_interval(x[1][1]) && return r
    fx = last(p[1])(x[1][1])
    return IntervalArithmetic.setdecoration(hull(fx, r), min(decoration(fx), decoration(r)))
end

function _evaluate(p::Tuple{Pair}, x, ::Nothing) # maybe not type stable
    isempty_interval(x[1][1]) && return x[1][1]
    return last(p[1])(x[1][1])
end
```

then

```Julia

foo = Piecewise((interval(-Inf, 0), Symbol(11)) => x -> -x,
                (interval( 0, Inf), Symbol(01)) => x -> interval(3));

foo(interval(-1, 0))

foo(interval(-1, 1)) # decoration is wrong, currently not implemented

foo(interval(0, 1))

foo(interval(0))

goo = Piecewise((interval(0, 1), Symbol(00)) => x -> interval(1),
                (interval(2, 3), Symbol(01)) => x -> interval(3));

goo(interval(0)) # now it throws

goo(interval(1, 2)) # now it throws

goo(interval(2, 3))
```

Kolaru:
I will try to implement the ForwardDiff extension for this, it may give some directions as to how to to handle the boundary points.

Kolaru:
In the current version (which still has a bug in the derivative computation), I am using `Interval` and `IntervalSet` from `Intervals`. It is very convenient, as they allow to specify inclusion of the boundary points and implement all required set operations.

To avoid confusion, I am aliasing `Intervals.Interval` to `IntervalArithmetic.Domain` (exported), so that a piecewise function can be defined as

```julia
julia>     slide = Piecewise(
                   Domain(-Inf, -1, false, true) => x -> -2x - 1,  # `false, true` means open on the left, closed on the right
                   Domain(-1, 0, false, true) => x -> x^2,
                   Domain(0, Inf, false, false) => Constant(0) ;
                   continuity = [1, 1]
           )
Piecewise function with 3 pieces:
  (-Inf .. -1.0] -> var"#135#137"()
  (-1 .. 0] -> var"#136#138"()
  (0.0 .. Inf) -> Constant{Int64}(0)
```

The domains need to be explicitely disjoint and ordered, which is checked at construction.

OlivierHnt:
One thing I would like to discuss is how necessary it is to add the dependancy on `IntervalSet`. Ideally I think we should always try to keep everything "in house", a fortiori for small features.

Kolaru:
It is not striclty necessary, the only things I import are the following:

```julia
import Intervals
import Intervals: IntervalSet, less_than_disjoint, Open, Closed

const Domain = Intervals.Interval
```

I have no strong feeling about it, it must just be noted that in this case `Intervals.jl` is doing exactly what we need (dealing with open/closed intervals, and computing overlaps).

One advantage of pulling the code out is to avoid weird error showing something went wrong with `Interval.Interval` that are very confusing, considering that the user is not facing anything named like that.

OlivierHnt:
> One advantage of pulling the code out is to avoid weird error showing something went wrong with Interval.Interval that are very confusing, considering that the user is not facing anything named like that.

Indeed 😞 

Maybe if we use what we have for `interval` we can get just what we need without too much effort, 
```Julia
struct Domain{T<:Symbol,S<:Symbol,R<:NumTypes}
    left :: T
    right :: S
    interval :: Interval{R}
    Domain{T,S,R}() = ...
end
```

Kolaru:
I will try that :)

I think that it will probably be better to not wrap an Interval here, just because we don't want any of the complication that come with it, just two numbers, bound types and some set operations is enough. If more is needed we can always provide conversion function to our `Intervals` or to `Intervals.Interval` (as an extension).

src/intervals/interval_operations/set_operations.jl
```diff
@@ -201,3 +201,17 @@ function _interiordiff(x::Interval, y::Interval)
     t = isguaranteed(x) & isguaranteed(y)
     return (_unsafe_interval(h₁, d, t), _unsafe_interval(h₂, d, t), _unsafe_interval(inter, d, t))
 end
+
+function interval_diff(x::Interval{T}, y::Interval) where {T<:NumTypes}
+    isdisjoint_interval(x, y) && return [x]
+    isequal_interval(x, y) && return Interval{T}[]
```
OlivierHnt, line 207: Maybe `issubset_interval`?

src/piecewise.jl
```diff
@@ -0,0 +1,208 @@
+struct Domain{L, R, T, S}
+    lo::T
+    hi::S
+end
```
OlivierHnt, line 4: Maybe there should be an inner constructor checking that `L` and `R` are either `:closed` or `:open` ?

src/piecewise.jl
```diff
@@ -0,0 +1,208 @@
...
+
+function leftof(d1::Domain, d2::Domain)
+    val1, bound1 = upperbound(d1)
+    val2, bound2 = lowerbound(d2)
+
+    val1 == val2 && return !(bound1 == bound2 == :closed)
+    return val1 < val2
+end
+
+Base.in(x::Real, domain::Domain) = rightof(x, lowerbound(domain)) && leftof(x, upperbound(domain))
```
OlivierHnt, line 62: Mhm maybe we should only export `Domain` to the user and not overload Base functions for it.

OlivierHnt:
With this new `Domain` type, I think this PR can also address #661.

src/piecewise.jl
Kolaru: Why not ? They seem quite natural operation to perform.

Kolaru:
> With this new `Domain` type, I think this PR can also address #661.

For clarity and brevity, I would keep it for a follow up PR.

A good question though is: would we then define function with limited domains (e.g. `log`) as `Piecewise` functions? It would add a bit of overhead, but then the logic of checking the domain would be common across the package.

src/piecewise.jl
OlivierHnt: I am not entirely sure 😄. My reasoning is based on maintaining consistency : it feels somewhat inconsistent that we have `in_interval` for `BareInterval`, while for `Domain`, we overload `Base.in`. This is maybe a "flaw" or something to be improved, I do not know....
However, for now, it seems more prudent to first implement the necessary internal functionalities and then extend them incrementally on 0.22.x (this way we also avoid having to deal with breaking changes).

src/piecewise.jl
```diff
@@ -0,0 +1,208 @@
+struct Domain{L, R, T, S}
+    lo::T
+    hi::S
+end
+
+Domain{L, R}(lo::T, hi::S) where {T, S, L, R} = Domain{L, R, T, S}(lo, hi)
+Domain(lo, hi) = Domain{:open, :closed}(lo, hi)
```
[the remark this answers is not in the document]
OlivierHnt: Isn't that a good thing to get `trv` when there are singularities ? What am I missing ?

OlivierHnt:
Interesting. The main question is probably "how bad" this overhead is.
This may warrant opening a dedicated issue to discuss it.

src/IntervalArithmetic.jl
```diff
@@ -23,6 +23,11 @@ const RealIntervalType{T} = Union{BareInterval{T},Interval{T}}
 
 #
 
+include("piecewise.jl")
+    export Domain, Constant, Piecewise, Domain, domain, subdomains, discontinuities, pieces
```
OlivierHnt, line 27: `Domain` is exported twice I think

src/piecewise.jl
```diff
@@ -0,0 +1,215 @@
...
+           throw(ArgumentError("domains are either not ordered or not disjoint"))
+        end
+    end
+
+    singularities = sup.(subdomains[1:end-1])
+
+    return Piecewise(subdomains, fs, continuity, singularities)
+end
+
+function Piecewise(pairs::Vararg{Pair} ; continuity = fill(-1, length(pairs) - 1))
```
OlivierHnt, line 131: By default `piecewise.continuity` is a vector containing `-1`, which means all points are discontinuous points (cf. `discontinuities` functions)?

src/intervals/interval_operations/set_operations.jl
```diff
@@ -201,3 +201,17 @@ function _interiordiff(x::Interval, y::Interval)
     t = isguaranteed(x) & isguaranteed(y)
     return (_unsafe_interval(h₁, d, t), _unsafe_interval(h₂, d, t), _unsafe_interval(inter, d, t))
 end
+
+function interval_diff(x::Interval{T}, y::Interval) where {T<:NumTypes}
```
OlivierHnt, line 205: I think this function should not be exported. It is a bit misleading as it is **not** the set difference of intervals, but some other sort of difference operation.

OlivierHnt: The other option would be to document clearly the function since its name is ambiguous.

src/piecewise.jl
```diff
@@ -0,0 +1,215 @@
...
+(constant::Constant)(::Any) = constant.value
+(constant::Constant)(::Interval) = interval(constant.value)
+
+# TODO Proper type parameters
+struct Piecewise
+    domain::Vector
+    fs
+    continuity::Vector{Int}
+    singularities::Vector
+end
```
OlivierHnt, line 101: You probably already thought about it, but I would have imagined using `Tuple` here instead of `Vector`. I think these are good below length 16, or 32?

But maybe an implementation like
```Julia
struct Piecewise{N,T<:Tuple,S<:Tuple,R<:Tuple}
     domains :: T
     functions :: S
     continuity_orders :: NTuple{N,Int}
     singularities :: R
 end
```
is messy, or can hurt the performance?

src/piecewise.jl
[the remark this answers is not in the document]
Kolaru: It's fine, it's just a little bit stricter than a user may expect by default and it may be a bit confusing since it will register all endpoints as discontinuities, regardless of the `continuity` given.

At the same time, having a symmetric default sounds good... and a third option is to not have a default at all.

I think I will just remove the default for now, and it can be readded later without even being breaking :)

src/piecewise.jl
Kolaru: A `Domain` is a set, and never a number, so we don't have the same ambiguity we have for an `Interval`. I can't see how it could go wrong (famous words to regrets, I know ^^).

src/piecewise.jl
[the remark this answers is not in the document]
OlivierHnt: If it is a singularity in the math sense (like $0$ for $\log$) then I would expect `trv`. Though, if "singularity" means "endpoints of adjacent Intervals", then yes it may be too strict since the decoration is case dependent.

>I think I will just remove the default for now, and it can be readded later without even being breaking :)

Yea, if it's too hard to come up with a default case, then that makes sense.
In any case, which ever way you choose to go (asymmetrical default, or symmetrical, or no default) sounds good to me 🙂

src/piecewise.jl
Kolaru: We can use tuples. It's just a bit messy because domains and functions can different types, so the type has whole tuple type as type parameters. But that's mostly a technicality.

OlivierHnt: Indeed, I let you decide what you think is best 🙂

Kolaru:
I've addressed all you comments and added docstrings and doc , therefore (if the doc builds properly), this is ready for a final review and merging.

src/piecewise.jl
```diff
@@ -0,0 +1,312 @@
...
+        end
+    end
+
+    singularities = sup.(domains[1:end-1])
+
+    return Piecewise(Tuple(domains), Tuple(fs), Tuple(continuity), Tuple(singularities))
+end
+
+function Piecewise(pairs::Vararg{Pair} ; continuity = fill(-1, length(pairs) - 1))
+    pairs = collect(pairs)
```
OlivierHnt, line 225: No need to `collect` since the code now uses `Tuple`

src/piecewise.jl
```diff
@@ -0,0 +1,312 @@
...
+        end
+
+        return new{N, D, F, S}(domains, fs, continuity, singularities)
+    end
+end
+
+function Piecewise(
+        domains::Vector{<:Domain},
+        fs,
+        continuity::Vector{Int} = fill(-1, length(domains) - 1))
```
OlivierHnt, line 199: Maybe these should be:
```Julia
function Piecewise(
         domains::NTuple{N,Domain},
         fs,
         continuity::Tuple{Vararg{Int}} = ntuple(i -> -1, Val(N-1))) where {N}
# .....
```

src/piecewise.jl
```diff
@@ -0,0 +1,312 @@
...
+           throw(ArgumentError("domains are either not ordered or not disjoint"))
+        end
+    end
+
+    singularities = sup.(domains[1:end-1])
+
+    return Piecewise(Tuple(domains), Tuple(fs), Tuple(continuity), Tuple(singularities))
+end
+
+function Piecewise(pairs::Vararg{Pair} ; continuity = fill(-1, length(pairs) - 1))
```
OlivierHnt, line 224: Similarly, maybe
```Julia
function Piecewise(pairs::Vararg{Pair,N} ; continuity = ntuple(i -> -1, Val(N-1))) where {N}
    #....
```

OlivierHnt:
Thank you so much @Kolaru for taking the time to address this issue. This is a great new functionality for IntervalArithmetic! 🥳

Besides my minor comments, the only remaining item for me is that I am not convinced we should overload `Base.in`, `Base.isempty`, etc. for `Domain` at the moment, since we do not do it for `BareInterval`.
I am not necessarily saying what we have going on is the best, but I believe we should be consistent. E.g., it's peculiar to have the following three behaviours
```Julia
julia> 1 in bareinterval(1, 2) # no method defined
ERROR: MethodError: [...]

julia> 1 in interval(1, 2) # it defaults to `Base.in` for `Number`
false

julia> 1 in Domain(:closed, :closed, 1, 2) # this should throw the same error as `1 in bareinterval(1, 2)`
true
```
If we have some (internal?) functions -- say `domain_in`, `domain_isempty`, `domain_intersect` etc -- we can still make the corresponding Base functions aliases later on without a major version bump.

Kolaru:
For now I have remove the overload of base function for `Domain`. I think it can wait for now anyway, because using piecewise functions should not require the user to manipulate the domains. We'll have to discuss it when addressing #661 

I also just did a little final bit of refactoring.

PR #698, opened 2025-02-21
Release 1.0

OlivierHnt:
This PR aims to prepare the 1.0 release.


dpsanders:
Very nice, thanks!

Whats the reason for removing @interval T a b?

OlivierHnt:
SparseArrays seems to have a hook `_iszero` and `_isnotzero` to handle `missing`; we may be able to use this as well for intervals.

OlivierHnt:
> Whats the reason for removing `@interval T a b`?

Well that's just for consistency, since the method `@interval a b` is missing (I do not know how to implement it without conflicting with `@interval T a`)

OlivierHnt:
Aqua tests are failing on Julia 1.10 due to method ambiguities with `mul!`.

OlivierHnt:
I have removed the old functions which controlled the default behaviour for the flavor, matrix multiplication, the rounding, the power.
Now this is controlled by a `configure(; rounding=:correct, flavor = :set_based, power = :fast, matmul = :fast)` function.

OlivierHnt:
Note: maybe `interval(1, 2) == interval(3, 4)` is also valid since there is no overlapping numbers. Whatever is decided should also be applied to `isapprox`.

lbenet:
The problem I see with the discussion around `==`, `<`, `>`, `iszero`, etc, is that we are ending up defining methods by value, not by type; obviously, this touches type stability. The return type of these operators, if allowed, will be a `Union{Bool, SomeError}`, and while errors may be good and appreciated in certain cases, they also break the flow which may or may not  be desired. 

I think we could have them but within a submodule (e.g., `Symbols`). Then, we encourage the usage of `isequal_interval`, or if the submodule is loaded for these extensions, the user is concient of the subtleties, e.g. the returned union types. This proposal is to mimic the current usage of `IntervalArithmetic.Symbols`.

Kolaru:
I think a reasonable philosophy for v1.0 is
- `Interval` act as much as a number as possible while keeping a guarantee of correctness. Which means that for them, we need to allow things like `<` and `==`, but error when the operation is ill-defined as the pointwise extension of the function on number (e.g. `(1..10) < (2..7)` must be both true and false, and thus should error).
- `BareInterval` do not extend any properties of numbers. They are pure intervals, do not interface with numbers and, notably, do not support the pointwise extension of boolean functions.

It could be nice to have some in-between for specific applications. But I think it is reasonable for the time being to redirect users to use `BareInterval` if the extended functionalities are a concern.

OlivierHnt:
Other things to do:
- [ ] check this PR against IntervalRootFinding.jl @Kolaru
- [ ] move `det`, `eigvals`, `norm` into a package extension

OlivierHnt:
I have added the following definition in the ForwardDiff extension:
```Julia
ForwardDiff.DiffRules._abs_deriv(x::Dual{T,<:Interval}) where {T} =
    Dual{T}(ForwardDiff.DiffRules._abs_deriv(value(x)), zero(partials(x)))
```

We now get the behaviour:
```Julia
julia> using IntervalArithmetic, ForwardDiff

julia> g(x) = abs(x)^2

julia> a = ForwardDiff.hessian(v -> g(v[1]), [interval(0)])
1×1 Matrix{Interval{Float64}}:
 [0.0, 0.0]_trv_NG

julia> b = ForwardDiff.hessian(v -> g(v[1]), [interval(-1, 1)])
1×1 Matrix{Interval{Float64}}:
 [-2.0, 2.0]_trv_NG

julia> c = ForwardDiff.hessian(v -> g(v[1]), [interval(1)])
1×1 Matrix{Interval{Float64}}:
 [2.0, 2.0]_com_NG

julia> d = ForwardDiff.hessian(v -> g(v[1]), [interval(-1)])
1×1 Matrix{Interval{Float64}}:
 [2.0, 2.0]_com_NG
```
(it used to error before)

Obviously, the correct answer is $2$, and `c` and `d` are consistent.
- `a` is not correct due to us setting the derivative of `abs` at `0` to `0` (see the discussion https://github.com/JuliaDiff/DiffRules.jl/pull/98 and our decision https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/564).
- `b` is a bit peculiar and is probably due to a product operation similar to how `interval(-1, 1) * interval(-1, 1)` is different than `interval(-1, 1)^2`.

So `a` may be pretty bad; `b` is not ideal, but at least it contains the correct value $2$.

@Kolaru any thoughts? Could this break [IntervalRootFinding.jl](https://github.com/JuliaIntervals/IntervalRootFinding.jl)?
cc @agerlach

OlivierHnt:
On this version of the 1.0 release, `==`, `<` are defined and other functions such as `in` and `issubset` work.
Whenever we have non-thin intervals that are overlapping when calling `==` and `<`, an error is thrown.

```Julia
julia> x = interval(1); y = interval(2); z = interval(2, 3)
[2.0, 3.0]_com

julia> x == y
false

julia> y == z
ERROR: ArgumentError: `==` is purposely not supported when the number is contained in the interval. See instead `isthin`

julia> z == z
ERROR: ArgumentError: `==` is purposely not supported when the intervals are overlapping. See instead `isequal_interval`

julia> x < y
true

julia> y < y
false

julia> z < z
ERROR: ArgumentError: `<` is purposely not supported when the intervals are overlapping. See instead `strictprecedes`

julia> y in z
ERROR: ArgumentError: `==` is purposely not supported when the number is contained in the interval. See instead `isthin`

julia> z in z
ERROR: ArgumentError: `==` is purposely not supported when the intervals are overlapping. See instead `isequal_interval`

julia> issubset(y, z)
ERROR: ArgumentError: `==` is purposely not supported when the number is contained in the interval. See instead `isthin`

julia> issubset(z, z)
ERROR: ArgumentError: `==` is purposely not supported when the intervals are overlapping. See instead `isequal_interval`
```

src/intervals/real_interface.jl
```diff
@@ -78,96 +78,77 @@ Base.hash(x::Interval, h::UInt) = hash(sup(x), hash(inf(x), hash(Interval, h)))
...
+#     throw(ArgumentError("`isdisjoint` is purposely not supported for intervals. See instead `isdisjoint_interval`"))
+
+# Base.issubset(::Interval, ::Interval) =
+#     throw(ArgumentError("`issubset` is purposely not supported for intervals. See instead `issubset_interval`"))
+
+# Base.issetequal(::Interval, ::Interval) =
+#     throw(ArgumentError("`issetequal` is purposely not supported for intervals. See instead `isequal_interval`"))
+
+# Base.in(::Interval, ::Interval) =
+#     throw(ArgumentError("`in` is purposely not supported for intervals. See instead `in_interval`"))
```
Kolaru, line 106: These overrides should be brought back to be consistent with #719

src/intervals/real_interface.jl
```diff
@@ -78,96 +78,77 @@ Base.hash(x::Interval, h::UInt) = hash(sup(x), hash(inf(x), hash(Interval, h)))
...
 Base.:(==)(x::Interval, y::Complex) = isreal(y) & (real(y) == x)
 Base.:(==)(x::Complex, y::Interval) = y == x
 
-# follows docstring of `Base.iszero`
-Base.iszero(x::Union{BareInterval,Interval}) = isthinzero(x)
 
-# follows docstring of `Base.isone`
-Base.isone(x::Union{BareInterval,Interval}) = isthinone(x)
+Base.:<(x::Interval, y::Real) = (!isthin(x) & in_interval(y, x)) | isempty_interval(x) ? throw(ArgumentError("`<` is purposely not supported when the number is contained in the interval, or if the interval is empty")) : sup(x) < y
+Base.:<(x::Real, y::Interval) = (!isthin(y) & in_interval(x, y)) | isempty_interval(y) ? throw(ArgumentError("`<` is purposely not supported when the number is contained in the interval, or if the interval is empty")) : x < inf(y)
```
Kolaru, line 151:
```suggestion
Base.:<(x::Interval, y::Real) = ((!isthin(x) & in_interval(y, x)) | isempty_interval(x)) ? throw(ArgumentError("`<` is purposely not supported when the number is contained in the interval, or if the interval is empty")) : sup(x) < y
Base.:<(x::Real, y::Interval) = ((!isthin(y) & in_interval(x, y)) | isempty_interval(y)) ? throw(ArgumentError("`<` is purposely not supported when the number is contained in the interval, or if the interval is empty")) : x < inf(y)
```

I got tripped by the precedence of `?` compared to `|`.

Why are `|` and `&` used instead of `||` and `&&` here ?

test/interval_tests/set_operations.jl
```diff
@@ -1,17 +1,17 @@
...
-    @test_throws ArgumentError intersect(interval(1), interval(2.), interval(3.))
-    @test_throws ArgumentError union(interval(1))
-    @test_throws ArgumentError union(interval(1), 2, [1], 4., 5)
-    @test_throws ArgumentError union(interval(1), interval(2.), interval(3.))
-    @test_throws ArgumentError setdiff(interval(1))
-    @test_throws ArgumentError setdiff(interval(1), 2, [1], 4., 5)
-    @test_throws ArgumentError setdiff(interval(1), interval(2.), interval(3.))
-    @test_throws ArgumentError symdiff(interval(1), interval(2.), interval(3.))
-end
+# @testset "removed interval" begin
```
Kolaru, line 3: Why is this commented out ?

OlivierHnt:
@Kolaru if you can double check the changes
We may still be missing some documentation, but we can add it in a separate PR.

Kolaru:
@OlivierHnt I was a bit slow, but still double checked the last changes, and everything looks good to me.

It is really great we finally made it :D

PR #710, opened 2025-05-01
Add Random package extension

### LINKED ISSUE #709: Implement rand(::Interval)
Is it worth to have a function returning a "random" point inside agiven interval? I use something like this for testing in TaylorModels, but implementing it there sounds like type pyracy. (I think we actually had something alike in v0.20)

OlivierHnt:
Closes #709


OlivierHnt:
The error is unrelated and only occurs on Julia nightly. Merging this PR now.

lbenet:
Thanks a lot!!

OlivierHnt:
Should we make a new release now of IntervalArithmetic ?

lbenet:
As you wish; it makes life easier, but I can adapt...

OlivierHnt:
Then let's do it right away :)

PR #711, opened 2025-05-01
Add `rand(::Interval)`

### LINKED ISSUE #709: Implement rand(::Interval)

OlivierHnt:
This properly addresses #709. Closes #709.

This also contains a fix for a bug with `mid(x, t)` and $t \ne 1/2$, where floating point errors could result in a midpoint value outside the prescribed interval.

cc @lbenet @dpsanders


OlivierHnt:
Again the error is unrelated to this PR.

PR #714, opened 2025-05-02
Avoid circular dependency issue on 1.10 with Random

### LINKED ISSUE #713: Circular Dependency issues
On the new releases it looks like #710 is creating a large number of circular dependency issues on 1.10.

Here is a MWE in a fresh environment

```julia
Info Packages marked with ⌃ and ⌅ have new versions available. Those with ⌃ may be upgradable, but those with ⌅ are restricted by compatibility constraints from upgrading. To see why use `status --outdated -m`
┌ Warning: Circular dependency detected.
│ Precompilation will be skipped for dependencies in this cycle:
│  ┌ IntervalArithmetic → IntervalArithmeticRandomExt
│  └─ IntervalArithmetic → IntervalArithmeticLinearAlgebraExt
│ Precompilation will also be skipped for the following, which depend on the above cycle:
│   IntervalArithmetic → IntervalArithmeticSparseArraysExt
│   ExactPredicates
│   DelaunayTriangulation
│   IntervalArithmetic → IntervalArithmeticIntervalSetsExt
│   Makie
└ @ Pkg.API.Precompilation ~/.julia/juliaup/julia-1.10.9+0.x64.linux.gnu/share/julia/stdlib/v1.10/Pkg/src/precompilation.jl:583
```

I believe this is related to https://github.com/JuliaObjects/Accessors.jl/issues/192 and is discussed in some detail https://github.com/JuliaLang/julia/issues/52511. A temporary solution is to revert #710, which is what Accessors did as well.

OlivierHnt:
We cannot have two standard libraries as package extensions due to an error on 1.10 (see https://github.com/JuliaLang/julia/issues/52511).
This makes Random a direct dependency.

Closes #713.


OlivierHnt:
@ptiede can you check that everything works fine with this PR?

ptiede:
Yep I can try later today!

ptiede:
Yep this fixes my issues!

Issue #647, opened 2024-05-18
Set membership doesn't work


dpsanders:
```jl
julia> 3.5 ∈ interval(3, 4)
false
```
This should either work or give me an error message telling me that this is not supported for intervals and telling me which special interval function to use instead (which is not discoverable).

dpsanders:
Apparently it's `in_interval`

OlivierHnt:
I took the freedom of re-naming this issue since this behaviour is intentional and has been extensively discussed. In a nutshell, in Julia the function `in` pertains to containers (which `Interval` is not); we cannot remove it completely (throw an error) for otherwise some generic functions fail, we cannot overload it with our own definition since this leads to silent errors in generic code.

We should highlight this in the docs, and maybe just add a docstring for `in`?

Kolaru:
I feel that we may have gone a long way since those discussions, and I am wondering if we shouldn't use the same trick for `in` that we use for `==`, namely

```julia
3.2 in interval(3.2)  # True, because the interval is a singleton and `3.2 in 3.2` is true

3.2 in interval(7.9, 12.1)  # False, because for every x in the interval `3.2 in x` is  false

3.2 in interval(0.3, 7.1)  # Error, because no conclusion can be drawn
```

In fact, for numbers, `==` and `in` are equivalent, and we don't have that anymore in the package.

OlivierHnt:
That does seem fair.

Kolaru:
In today's call we decide to be strict and consider by default that intervals are numbers but never sets with builtin julia function. In particular, this means `in` should be disabled as ambiguous.

Also, the documentation should contain a section on our general philosophy about it, so that we can refer to it in future discussion.

Issue #675, opened 2024-07-28
Document `isapprox`


dpsanders:
```jl
julia> isapprox(1..2, 3..4)
ERROR: ArgumentError: `==` is purposely not supported for intervals. See instead `isequal_interval`
Stacktrace:
 [1] ==(x::Interval{Float64}, y::Interval{Float64})
   @ IntervalArithmetic ~/Dropbox/packages/IntervalArithmetic/src/intervals/real_interface.jl:86
 [2] #isapprox#809
   @ ./floatfuncs.jl:308 [inlined]
 [3] isapprox(x::Interval{Float64}, y::Interval{Float64})
   @ Base ./floatfuncs.jl:304
 [4] top-level scope
   @ REPL[3]:1
```

This should either work or fail with a sensible error message.

OlivierHnt:
There is a section in the documentation [Comparisons and set operations](https://juliaintervals.github.io/IntervalArithmetic.jl/stable/manual/usage/#Comparisons-and-set-operations) whose first sentence reads "All comparisons and set operations for Real have been purposely disallowed to prevent silent errors".

We should explicitly mention `isapprox` there.

dpsanders:
Sorry I meant that isapprox has not been explicitly disallowed - we haven't defined it (to error) for intervals. It's falling back to a generic definition in base as far as I can see.

OlivierHnt:
Actually it was disallowed on purpose. We currently just rely on a fallback error.
This is the case for several other functions, eg `symdiff` or `intersect`, etc. (see `intersect(interval(1, 2), interval(1, 2))`).

That being said, I am not opposed to customise the error message eventually.

PR #719, opened 2025-05-18
Document package design choices

Kolaru:
As discussed last Friday, I have drafted an addition to the documentation to discuss some design choices that we have  made.

I have summarized all discussion that I could think of for now, but please let me know if any other controversies should be mentionned to.

### DIFF
--- a/docs/src/philosophy.md
+++ b/docs/src/philosophy.md
@@ -1,0 +1,127 @@
+# Philosophy
+
+The goal of the `Interval` type is to be directly used to replace floating point
+number in arbitrary julia code, such that in any calculation,
+the resulting intervals are guaranteed to bound the true image of the starting
+intervals.
+
+Due to the way that the julia ecosystem has evolved,
+this means that `Interval` **must** be a subtype of `Real`,
+as it is the default supertype used to describe
+"numbers that are not complex".
+Then, for any function `f(x::Real)`,
+we want the following to hold for all real `x` in the interval `X`
+(note that it holds for **all** real numbers in `X`,
+even those that can not be represented as floating point numbers):
+```julia
+f(x) ∈ f(X)
+```
+
+At first glance, this is reasonable:
+all arithmetic operations are well-defined for both real numbers and intervals,
+therefore we can use multiple dispatch to define the interval behavior of
+operations such has `+`, `/`, `sin` or `log`.
+Then a code written for `Real`s can be used as is with `Interval`s.
+
+However, being a `Real` means way more than just being compatible with
+arithmetic operations.
+`Real`s are also expected to
+
+1. Be compatible with any other `Number` through promotion.
+2. Support comparison operations, such as `==` or `<`.
+3. Act as a container of a single element,
+   e.g. `collect(x)` returns a 0-dimensional array containing `x`.
+
+Each of those points lead to specific design choice for `IntervalArithmetic.jl`,
+choices that we detail below.
+
+
+## Compatibility with other `Number`s
+
+In julia it is expected that `1 + 2.2` silently promoted the integer `1`
+to a `Float64` to be able to perform the addition.
+Following this logic, it means that `0.1 + interval(2.2, 2.3)` should 
+silently promote `0.1` to an interval.
+
+However, in this case we can not guarantee that `0.1` is known exactly,
+because we do not know how it was produced in the first place.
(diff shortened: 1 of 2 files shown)


docs/src/philosophy.md
```diff
@@ -0,0 +1,127 @@
...
+
+The goal of the `Interval` type is to be directly used to replace floating point
+number in arbitrary julia code, such that in any calculation,
+the resulting intervals are guaranteed to bound the true image of the starting
+intervals.
+
+Due to the way that the julia ecosystem has evolved,
+this means that `Interval` **must** be a subtype of `Real`,
+as it is the default supertype used to describe
+"numbers that are not complex".
```
OlivierHnt, line 11: Maybe "must" is too strong?

"
Due to the evolution of the Julia ecosystem -- specifically that `Real` serves as the default supertype for numerical types that are not complex --, it is particularly advantageous for `Interval` to be defined as a subtype of `Real`.
"

docs/src/philosophy.md
```diff
@@ -0,0 +1,127 @@
...
+Then, for any function `f(x::Real)`,
+we want the following to hold for all real `x` in the interval `X`
+(note that it holds for **all** real numbers in `X`,
+even those that can not be represented as floating point numbers):
+```julia
+f(x) ∈ f(X)
```
OlivierHnt, line 17: Should we use a math-block instead? Since `Base.in` is ambiguous; e.g.,

```math
f(x) \in f(X), \qquad \forall x \in X
```

docs/src/philosophy.md
```diff
@@ -0,0 +1,127 @@
...
+
+
+## Comparison operators
+
+We can extend our above definition of the desired behavior for two real numbers
+`x` and `y`, and their respective intervals `X` and `Y`.
+With this, we want to have, for any function`f`,
+for all `x` in `X` and all `y` in `Y`,
+```julia
+f(x, y) ∈ f(X, Y)
```
OlivierHnt, line 63: Same comment, maybe
```math
f(x, y) \in f(X, Y), \qquad \forall x \in X, \, y \in Y
```

docs/src/philosophy.md
```diff
@@ -0,0 +1,127 @@
...
+However, this is not convenient, as any `if` statement would error when used
+with an interval.
+Instead, we have opted to return respectively `false` and `true`
+for cases 1 and 2, and to immediately error otherwise.
+
+In this way, we can return a more informative error,
+but we only do it when the result is ambiguous.
+
+This has a clear cost, however, in that some expected behaviors do not hold.
+For example, an `Interval` is not equal to itself.
```
OlivierHnt, line 85: Maybe this can be removed. What you wrote above, seems already very clear.

docs/src/philosophy.md
```diff
@@ -0,0 +1,127 @@
...
+ [1] ==(x::Interval{Float64}, y::Interval{Float64})
+   @ IntervalArithmetic C:\Users\Kolaru\.julia\packages\IntervalArithmetic\XjBhk\src\intervals\real_interface.jl:86
+ [2] top-level scope
+   @ REPL[6]:1.
+```
+
+
+## Intervals as sets
+
+We have taken the perspective to always let `Interval`s act as if they were numbers.
```
OlivierHnt, line 102: If we want to not avoid the word "number":

"
We have taken the perspective to always let `Interval`s behave as numerical values.
"

OlivierHnt:
Thanks @Kolaru for the PR. It looks really good.

Closes #647.

Maybe due to #675, we can add a sentence (near where `Base.:(==)` is discussed) about `Base.isapprox`?

docs/src/philosophy.md
Kolaru: I have tuned it down a little bit, but I like "must" so that it is clear it is not open for discussion (at least for the current document).

docs/src/philosophy.md
Kolaru: I think it doesn't hurt to have an example.

Kolaru:
I've added a summary of the behaviors, I think it address your concerns.

I have also added `setdiff_interval` and `union_interval` for consistency (it simplified the summary :D).

I can add the doc for `@exact`.

I am a bit worried, as the documentation here does not match the behavior on master. Should I port the change from #698 for consistency ?

Issue #721, opened 2025-05-21
Arblib.jl ❤️ IntervalArithmetic.jl


Joel-Dahne:
As discussed in #717, there is potential in making [Arblib.jl](https://github.com/kalmarek/Arblib.jl/) and IntervalArithmetic.jl work together. I see two potential benefits:
1. Users can easily switch between doing different parts of their computations in different packages depending on which package suits that particular part of the computation best.
2. The IntervalArithmetic.jl library could use some of the Arblib.jl functionality to speed up certain computations (in particular high precision computations).

For the first point it would suffice to create a package extension, which in principle could live in either of the packages (though as I extend upon below it would likely be more natural for it to live in the IntervalArithmetic.jl package). The second point would mean IntervalArithmetic.jl taking on Arblib.jl as a full dependency.

## For users
To make the packages work well together for users the main thing to implement would be conversions between intervals and the types in Arblib.jl (primarily `Arb` and `Acb`, representing real and complex balls respectively). The user could then easily convert between the two packages to suit their needs.

Some things that Arblib.jl excel at:
1. Fast high precision computations
2. A large library of special functions
3. Support for mutable arithmetic
4. Taylor series expansions (in particular at high degrees)

Some things that IntervalArithmetic.jl excel at:
1. Fast computations at `Float32` and `Float64` precision
2. Computations with wide intervals (Arblib.jl is not optimized for this, though the situation is improving)
3. Built in safety features such as decorations (it is much easier to shoot yourself in the foot with Arblib.jl)
4. Plotting (this is actually my primary use case for using it)

In principle the conversion functions could be very simple. I have been using
```julia
Arblib.Arb(x::Union{Interval,BareInterval}) =
    isempty_interval(x) ? indeterminate(Arb) : Arb((inf(x), sup(x)))

Arblib.Acb(z::Complex{<:Interval}) = Acb(Arb(real(z)), Arb(imag(z)))

IntervalArithmetic.interval(::Type{T}, x::Arb) where {T} =
    isnan(x) ? nai(T) : interval(T, getinterval(BigFloat, x)...)

IntervalArithmetic.interval(::Type{T}, (x, y)::NTuple{2,Arb}) where {T} =
    (isnan(Arblib.midref(x)) || isnan(Arblib.midref(y))) ? nai(T) :
    interval(T, BigFloat(lbound(x)), BigFloat(ubound(y)))
```
There are a number of design decisions to make though. Some that come to mind are:
[...]

Issue #717, opened 2025-05-17
Setting `matmul` to fast makes all computations run in low precision


Joel-Dahne:
Toggling `IntervalArithmetic.configure(matmul = :fast)` seems to make it so that matrix multiplication for `BigFloat` only runs in `Float64` precision.
```julia
julia> using IntervalArithmetic

julia> A = sin.(interval.(rand(BigFloat, 4, 4))) # Apply sin to get non-zero radius

4×4 Matrix{Interval{BigFloat}}:
 [0.089218, 0.0892181]₂₅₆_com  [0.578398, 0.578399]₂₅₆_com  …  [0.732301, 0.732302]₂₅₆_com
 [0.50122, 0.501222]₂₅₆_com    [0.374433, 0.374434]₂₅₆_com     [0.677579, 0.67758]₂₅₆_com
 [0.499401, 0.499402]₂₅₆_com   [0.519317, 0.519319]₂₅₆_com     [0.52119, 0.521191]₂₅₆_com
 [0.470956, 0.470957]₂₅₆_com   [0.722737, 0.722738]₂₅₆_com     [0.245764, 0.245765]₂₅₆_com

julia> IntervalArithmetic.configure(matmul = :slow)
(Float64, :set_based, :correct, :fast, :slow)

julia> B = A * A
4×4 Matrix{Interval{BigFloat}}:
 [1.04141, 1.04142]₂₅₆_com    [1.212, 1.21201]₂₅₆_com     …  [1.05327, 1.05328]₂₅₆_com
 [0.835392, 0.835393]₂₅₆_com  [1.21502, 1.21504]₂₅₆_com      [1.08355, 1.08356]₂₅₆_com
 [0.964768, 0.96477]₂₅₆_com   [1.29097, 1.29098]₂₅₆_com      [1.27822, 1.27823]₂₅₆_com
 [0.601367, 0.601368]₂₅₆_com  [0.805239, 0.80524]₂₅₆_com     [0.979897, 0.979898]₂₅₆_com

julia> radius.(B)
4×4 Matrix{BigFloat}:
 3.45447e-77  5.1817e-77   5.1817e-77   3.45447e-77
 2.59085e-77  3.45447e-77  5.1817e-77   3.45447e-77
 2.59085e-77  3.45447e-77  5.1817e-77   5.1817e-77
 1.72723e-77  2.59085e-77  2.59085e-77  3.45447e-77

[...]

PR #723, opened 2025-06-03
Add extension for Arblib.jl

Joel-Dahne:
As discussed in #721, it would be nice to make the Arblib.jl and IntervalArithmetic.jl packages talk to each other. This is an attempt at making basic conversion between `Interval` and `Arb` work.

```julia
julia> using Arblib, IntervalArithmetic

julia> x = interval(BigFloat, π)
[3.14159, 3.1416]₂₅₆_com

julia> y = Arb(x)
[3.1415926535897932384626433832795028841971693993751058209749445923078164062862 +/- 5.38e-77]

julia> cos(x)
[-1.0, -0.999999]₂₅₆_com

julia> cos(y)
[-1.000000000000000000000000000000000000000000000000000000000000000000000000000 +/- 1.78e-77]
```

It is also possible to construct matrices and do basic linear algebra

```julia
julia> A = interval.(rand(BigFloat, 2, 2))
2×2 Matrix{Interval{BigFloat}}:
 [0.15383, 0.153831]₂₅₆_com  [0.655025, 0.655026]₂₅₆_com
 [0.165259, 0.16526]₂₅₆_com  [0.828643, 0.828644]₂₅₆_com

julia> B = ArbMatrix(A)
2×2 ArbMatrix:
 [0.15{…73 digits…}85 ± 3.64e-78]  [0.65{…73 digits…}67 ± 4.69e-78]
 [0.16{…73 digits…}30 ± 4.82e-78]  [0.82{…73 digits…}91 ± 1.48e-78]

julia> Arblib.overlaps(B * B, ArbMatrix(A * A))
true
```

# Implementation
Construction of `Arb` from `Interval` is straightforward. It only requires overloading `Arblib.set!` to handle inputs of type `Interval`. To make some of the internal precision handling correct, I also updated `Arblib._precision` to be aware of the `Interval` type.

I realized my first attempt at the construction of `Interval` from `Arb` was incorrect. The second attempt was super complicated and used a lot of internals of IntervalArithmetic. The final attempt ended up being fairly simple. For construction, we basically want to treat `Arb` as equivalent to `Interval{BigFloat}`. For this, we define

```julia
IA._inf(x::Arblib.ArbOrRef) = BigFloat(Arblib.lbound(x))
IA._sup(x::Arblib.ArbOrRef) = BigFloat(Arblib.ubound(x))
```

and make sure that `IntervalArithmetic.promote_numtype` handles promotion with `Arb` in the way we want. Note that this generates quite a lot of allocation, to compute `_inf(x)` it first computes the lower bound as and `Arf`, then converts it to a `BigFloat` and then further converts it to whatever type the interval should be. I could however note figure out a better way to do it at this point.

I also made the following updates:
- Made `Base.promote_rule` explicitly throw an error for promotion between `Arb` and `Interval`.
- Handle some ambiguities related to `ExactReal` as well as `Arf`.

I added **a lot** of tests. This was partially because my first few attempts of implementing the constructor were fairly complicated and I wanted to make sure that I checked all edge cases. The final implementation is a lot simpler and maybe doesn't warrant quite as many tests. But testing more never hurts!

Note that a few of the implemented tests are marked as broken. While doing the implementation, I noticed some issues with the `Arblib.getinterval` function for infinite intervals with infinite endpoints. This is due to an issue in Flint that I [reported](https://github.com/flintlib/flint/issues/2329).

### DIFF
--- a/ext/IntervalArithmeticArblibExt.jl
+++ b/ext/IntervalArithmeticArblibExt.jl
@@ -1,0 +1,153 @@
+module IntervalArithmeticArblibExt
+
+import IntervalArithmetic as IA
+import Arblib
+
+# Promotion rules
+
+# These are needed for ambiguity reasons
+Base.promote_rule(
+    ::Type{IA.Interval{T}},
+    ::Type{S},
+) where {T<:IA.NumTypes,S<:Arblib.ArfOrRef} = IA.Interval{promote_type(T, S)}
+Base.promote_rule(
+    ::Type{T},
+    ::Type{IA.Interval{S}},
+) where {T<:Arblib.ArfOrRef,S<:IA.NumTypes} = IA.Interval{promote_type(T, S)}
+
+Base.promote_rule(
+    ::Type{IA.Interval{T}},
+    ::Type{S},
+) where {T<:IA.NumTypes,S<:Arblib.ArbOrRef} =
+    throw(ArgumentError("promotion between Interval and Arb is purposely not supported"))
+Base.promote_rule(
+    ::Type{T},
+    ::Type{IA.Interval{S}},
+) where {T<:Arblib.ArbOrRef,S<:IA.NumTypes} =
+    throw(ArgumentError("promotion between Interval and Arb is purposely not supported"))
+
+# Construction of Arb from Interval
+# Note that this automatically gives support for construction of Acb.
+
+Arblib._precision(x::Union{IA.Interval,IA.BareInterval}) =
+    Arblib._precision(IA.inf(x), IA.sup(x))
+
+function Arblib.set!(
+    res::Arblib.ArbLike,
+    x::Union{IA.Interval,IA.BareInterval};
+    prec::Integer = precision(res),
+)
+    if IA.isempty_interval(x)
+        Arblib.indeterminate!(res)
+    else
+        Arblib.set!(res, (IA.inf(x), IA.sup(x)); prec)
+    end
+    return res
+end
+
+# Construction of Interval from Arb
+
+# We essentially treat Arb as Interval{BigFloat} for the purposes of
+# constructing Intervals. This means that we only have to change the
+# behavior of _inf and _sup since these are called early on in the
+# construction process to handle endpoints which are Intervals.
+
+IA._inf(x::Arblib.ArbOrRef) = BigFloat(Arblib.lbound(x))
+IA._sup(x::Arblib.ArbOrRef) = BigFloat(Arblib.ubound(x))
+
+# Make promote_numtype for Arb behave like promotion with BigFloat
+IA.promote_numtype(::Type{<:Arblib.ArbOrRef}, ::Type{<:Arblib.ArbOrRef}) = BigFloat
+IA.promote_numtype(::Type{<:Arblib.ArbOrRef}, ::Type{S}) where {S} =
+    IA.promote_numtype(BigFloat, S)
+IA.promote_numtype(::Type{<:Arblib.ArbOrRef}, ::Type{S}) where {S<:IA.NumTypes} =
+    IA.promote_numtype(BigFloat, S)
+IA.promote_numtype(::Type{T}, ::Type{<:Arblib.ArbOrRef}) where {T} =
+    IA.promote_numtype(T, BigFloat)
+IA.promote_numtype(::Type{T}, ::Type{<:Arblib.ArbOrRef}) where {T<:IA.NumTypes} =
+    IA.promote_numtype(T, BigFloat)
+
+# Construction of Complex{Interval} from Acb
+IA.numtype(::Type{<:Arblib.AcbOrRef}) = Arblib.Arb
+
+# Handle Acb in the same way as Complex.
+IA._interval_infsup(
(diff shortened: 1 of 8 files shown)


OlivierHnt:
This looks great!
So presumably `BigFloat(Arblib.lbound(x))` is equivalent to `BigFloat(Arblib.lbound(x); precision = precision(Arblib.lbound(x)))`?

In #721, we also mentioned special functions; do you think doing something like

```Julia
for f ∈ list_of_special_functions
    @eval Arblib.$f(x::Interval) = interval($f(Arblib.Arb(x)))
end
```

is worth it?

In the same spirit, maybe we can address #717 as well, which is to add the Arblib routines for multiprecision matrix product. I am not sure what is the best way of doing so, maybe 

```Julia
function IA._mul!(::IntervalArithmetic.MatMulMode{:fast}, C::AbstractVecOrMat{<:RealOrComplexI{<:BigFloat}}, A, B, α, β)
    # perform the fast Arblib algorithm in extended precision
end
```

is ok? One thing that could be puzzling is that loading Arblib will change the result (though still a rigorous enclosure) of the matrix product when `matmul` is set to `:fast`... or we disable the `:fast` algorithm on `BigFloat` unless Arblib is loaded...

Joel-Dahne:
> So presumably BigFloat(Arblib.lbound(x)) is equivalent to BigFloat(Arblib.lbound(x); precision = precision(Arblib.lbound(x)))?

Yes, so the conversion to `BigFloat` is exact in this case (as long as there is no overflow in the exponent, in which case you would get `Inf` or `-Inf`).

I think directly wrapping the special functions come with some issues. Ball arithmetic (like Arb uses) and infsup interval arithmetic (which IntervalArithmetic uses) tend to have fairly different assumptions about how in particular wide inputs are handled. As an example consider evaluating the gamma function on the interval $[2, 20]$ (where it is increasing). With direction evaluation with Arb we get

```julia
julia> using Arblib, IntervalArithmetic

julia> interval(Arblib.SpecialFunctions.gamma(Arb((2, 20))))
[-1.67773e+09, 1.21646e+17]₂₅₆_com
```

whereas most user would probably expect something more like

```julia
julia> interval(Arblib.SpecialFunctions.gamma(Arb(2)), Arblib.SpecialFunctions.gamma(Arb(20)))
[1.0, 1.21646e+17]₂₅₆_com
```

If the user explicitly converts to Arb, calls the special functions and then converts back it would hopefully be more clear to them where this discrepancy is coming from.

For #717 I'm not sure what the best approach is, it is certainly a bit awkward to have to different package extensions interact in this way. That `LinearAlgebra` is always implicitly loaded by Julia at least makes it easier. But I could definitely see people being confused about the result of matrix multiplication changing if they load another package. In particular since they don't even have to load it explicitly, but it is enough that they load another package that depend on it.

I was thinking that I could add some page to the documentation, something like "Interfacting with Arblib.jl", and write something about the discussions here. Then users would at least be aware of the existence of these tools. Then we could leave any direct implementations to the future. Does that sound reasonable to you? In that case I should hopefully be able to write some sort of documentation by the beginning of next week.

OlivierHnt:
I think adding a documentation page with short descriptions of all our package extensions is a great idea. Since we haven’t done this before, @Joel-Dahne, feel free to get it started in whatever format you think works best.
After that, let us merge this PR.

Joel-Dahne:
I have now added some documentation for the interface. The goal was to give users some ideas of what type of computations could be reasonable to use Arblib.jl for. In the future we might add a similar page of the documentation to Arblib.jl to discuss what type of computations it could be reasonable to use IntervalArithmetic.jl for.

I have also added conversion between complex intervals and `Acb` and that I have fixed `conversion` to so that it doesn't add the `NG` flag for implicit conversions from `Arb` and `Acb`.

Let me know if you have any comments, in particular regarding the documentation! Otherwise I think this should in principle be ready to merge. 

I had some issues building the documentation locally, it didn't load the package extension as I expected. It seems to work in the CI though.

ext/IntervalArithmeticArblibExt.jl
```diff
@@ -0,0 +1,153 @@
...
+
+# Construction of Complex{Interval} from Acb
+IA.numtype(::Type{<:Arblib.AcbOrRef}) = Arblib.Arb
+
+# Handle Acb in the same way as Complex.
+IA._interval_infsup(
+    ::Type{T},
+    a::Arblib.AcbOrRef,
+    b::Arblib.AcbOrRef,
+    d::IA.Decoration = com,
```
OlivierHnt, line 77: Is a default value for the decoration really needed? And if so, shouldn't that be `IA.com`?

Joel-Dahne: Good point, it should be `IA.com`. Indeed, it is probably not needed. I copied it from the [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/d948dabc770de047b8a74b628b5c4238af6d5026/src/intervals/construction.jl#L462C1-L474C1) and I see now that the these have default values for `d` but the versions above don't. I'll remove the default value.

PR #728, opened 2025-09-04
Release 1.0

OlivierHnt:
🥳

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,6 +13,7 @@ jobs:
         version:
           - '1.10'
           - '1'
+          - 'pre'
           - 'nightly'
         os:
           - ubuntu-latest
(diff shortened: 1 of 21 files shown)


lbenet:
Can you give me a couple of days to check if Taylor-related packages are ok with these changes? Would there be a v0.23.1, or shall we jump directly to v1.0?

OlivierHnt:
I think we should add `exact` as a constructor of `ExactReal` (to do things like `exact(::AbstractArray)` etc.)

Actually, one thought occurred to me: since `ExactReal <: Real`, should we only define `exact` as a constructor and disallow `ExactReal`?
It seems possible that the conversion / promotion system can end up wrapping a `Real` as an `ExactReal` "implicitly".

If this is indeed a reasonable concern, then I suggest we replace the `ExactReal` constructors with `exact` as a conservative measure for the 1.0 release. Non breaking changes (even restoring `ExactReal` as a constructor) can always be made later on.

lbenet:
I just restarted the tests of the failed jobs (nightly)....

OlivierHnt:
Testa are passing with the next release candidate 1.12.
So I think we are fine.

mforets:
congratulations for the release, thanks for the hard work 🚀

PR #745, opened 2026-03-09
Improve power of complex intervals

OlivierHnt:
Since we use the formula `exp(y * log(x))` to evaluate `x^y` when `x::Complex{<:Interval}`, `y::Complex{<:Interval}`, we get the following undesirable overestimation:

```julia
x = interval(-0.5) + interval(im) * interval(-1e-14, 1e-14)
x^1 # [-0.5, 0.5]_com + im*[-0.5, 0.5]_com
x^2 # [-0.25, 0.25]_com + im*[-0.25, 0.25]_com
```

yet

```Julia
x*x # [0.25, 0.25]_com + im*[-1.0e-14, 1.0e-14]_com
```

This PR implements the following definition for complex powers:
- If $y = [a, b]$ ($a,b$ finite) with $b-a \le 2$, then there exist $n \in \mathbb{Z}$, $\delta \subset [-1, 1]$ such that $y = n + \delta$.
We compute
```math
x^y =
\underbrace{x \cdot \ldots \cdot x}_{n-\text{times}} \cdot \exp(\delta \, \ln(x)).
```
- If $b-a > 2$, then we can write $y = \cup_{j=1}^d y_j$ where $y_j = n_j + \delta_j$ with $n_j \in \mathbb{Z}$, $\delta_1 \subset [-1, 0.5]$, $\delta_d \subset [-0.5, 1]$, and $\delta_j = [-0.5, 0.5]$ for $j=2, \dots, d-1$.
We compute
```math
x^y = \cup_{j=1}^d x^{y_j} = \cup_{j=1}^d \underbrace{x \cdot \ldots \cdot x}_{n_j-\text{times}} \cdot \exp(\delta_j \, \ln(x)).
```
- If $y$ is not in the above two forms (i.e., $y$ is real and unbounded, or $y$ has a non-zero imaginary part), then $x^y = \exp(y \\, \ln(x))$.

With this PR

```Julia
x = interval(-0.5) + interval(im) * interval(-1e-14, 1e-14)
x^1 # [-0.5, -0.5]_com + im*[-1.0e-14, 1.0e-14]_com
x^2 # [0.25, 0.25]_com + im*[-1.0e-14, 1.0e-14]_com
```

### DIFF
--- a/test/interval_tests/power.jl
+++ b/test/interval_tests/power.jl
@@ -119,6 +119,24 @@ end
     @test issubset_interval(interval(0, 1)*interval(im), sqrt(interval(-1, 0) + interval(0)*interval(im)))
     @test issubset_interval(interval(0, 1) + interval(0, 1)*interval(im), sqrt(interval(-1, 1) + interval(0)*interval(im)))
     @test issubset_interval(interval(0, Inf) + interval(-3//8, Inf)*interval(im), sqrt(interval(-9//32, Inf)*interval(im)))
+
+    x = interval(-0.5) + interval(im) * interval(-1e-14, 1e-14)
+
+    y = interval(1)
+    @test issubset_interval(x^y, exp(y*log(x)))
+
+    y = interval(1.5, 2.5)
+    res = x^y
+    ref = exp(y*log(x))
+    # tighter on the real part (right half-plane)
+    @test inf(real(ref)) < -0.3 < -1e-13 < inf(real(res))
+
+    y = interval(-2, Inf)
+    @test isequal_interval(x^y, exp(y*log(x)))
+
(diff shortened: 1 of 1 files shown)


OlivierHnt:
I could not simply have a test that `x^y` is a subset of `exp(y*log(x))` since the new algorithm uses power by squaring which can accumulates a bit more rounding error than simply performing the exp/log version.

Instead, I check that we do have an improved real part:
```Julia
x = interval(-0.5) + interval(im) * interval(-1e-14, 1e-14)
y = interval(1.5, 2.5)
res = x^y
ref = exp(y*log(x))
# tighter on the real part (right half-plane)
@test inf(real(ref)) < -0.3 < -1e-13 < inf(real(res))
```

Let me know if you have any ideas for tests before we merge the PR.

PR #746, opened 2026-03-12
Improved matrix exponential for real matrices

BaronNashville:
The exponential of some real interval matrices outputted complex intervals. Added a check to rule that case out.

### DIFF
--- a/ext/IntervalArithmeticLinearAlgebraExt.jl
+++ b/ext/IntervalArithmeticLinearAlgebraExt.jl
@@ -207,12 +207,10 @@ function LinearAlgebra.exp!(A::AbstractMatrix{T}) where T<:RealOrComplexI
     # note: this function does not overwrite `A`
     Λ, V = LinearAlgebra.eigen(A)
     V⁻¹ = inv(V)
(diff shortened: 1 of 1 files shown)


ext/IntervalArithmeticLinearAlgebraExt.jl
```diff
@@ -203,11 +203,15 @@ end
...
+function LinearAlgebra.exp!(A::AbstractMatrix{T}) where T<:RealOrComplexI
     # note: this function does not overwrite `A`
     Λ, V = LinearAlgebra.eigen(A)
     V⁻¹ = inv(V)
-    return V * LinearAlgebra.Diagonal(exp.(Λ)) * V⁻¹
+    if T <: Real
+        return real(V * LinearAlgebra.Diagonal(exp.(Λ)) * V⁻¹)
+    else
+        return V * LinearAlgebra.Diagonal(exp.(Λ)) * V⁻¹
+    end
```
OlivierHnt: while this works, this is not idiomatic code in Julia. I suggest dispatching via the function signature.

PR #752, opened 2026-04-15
Fix display of non-thin intervals showing identical bounds

### LINKED ISSUE #751: Non-thin interval is displayed as thin interval
```julia
julia> x = interval(0.1)
[0.1, 0.1]

julia> sin(x)
[0.0998334, 0.0998334]

julia> diam(sin(x))
1.3877787807814457e-17
```

The interval `sin(x)` is not thin (the diameter is > 0), but the display shows the same lower and upper bounds.

dpsanders:
## Summary

Fixes #751.

- Restores directed rounding for interval display: lower bounds are rounded down (`RoundDown`) and upper bounds are rounded up (`RoundUp`), so displayed intervals are always valid enclosures of the true interval.
- This was accidentally removed in the 1.0 release (commit a8ccabf). The old code converted to `BigFloat` and used MPFR string formatting with manual digit-level rounding — that approach is restored here.
- Updates test expectations to match the correct directed-rounding output.

**Before:** `sin(interval(0.1))` displayed as `[0.0998334, 0.0998334]` (identical bounds despite non-zero diameter)

**After:** `sin(interval(0.1))` displays as `[0.0998334, 0.0998335]` (distinct bounds, valid enclosure)

## Test plan

- [x] All existing display tests pass (updated expectations for directed rounding)
- [x] Full test suite passes (`Pkg.test()`)
- [x] Manual verification of the issue example: `sin(interval(0.1))` now shows distinct bounds
- [x] Thin intervals (e.g. `interval(1.0, 1.0)`) still display as `[1.0, 1.0]`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -21,6 +21,7 @@ makedocs(;
         "Overview" => "intro.md",
         "Philosophy" => "philosophy.md",
         "Manual" => [
+            "Cheat Sheet" => "manual/cheatsheet.md",
             "Constructing intervals" => "manual/construction.md",
             "Guarantee and `ExactReal`" => "manual/guarantee.md",
             "Usage" => "manual/usage.md",
--- a/docs/src/manual/cheatsheet.md
+++ b/docs/src/manual/cheatsheet.md
@@ -1,0 +1,481 @@
+# Cheat Sheet
+
+A quick reference for IntervalArithmetic.jl v1.
+
+## Type Hierarchy
+
+| Type | Description | Subtype of `Real`? |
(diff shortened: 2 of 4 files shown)


OlivierHnt:
The display of intervals is a bit tricky because it is always lying. But indeed, there is an argument to be made that it is better for the display to indicate, at least, that an interval is not thin even if that means showing a wider interval.

I'll have a look at the PR.

OlivierHnt:
To be superseded.

PR #757, opened 2026-05-11
Complete Decoration docstring

Kolaru:
As discussed in https://github.com/JuliaIntervals/IntervalRootFinding.jl/pull/243 this PR export `Decoration` and complete the associated docstrings, in order to make clear that `com`, `dac` and `def` are fine in general.

### DIFF
--- a/src/intervals/construction.jl
+++ b/src/intervals/construction.jl
@@ -308,8 +308,6 @@ bareinterval(::Type{T}, x::Interval) where {T} = bareinterval(T, bareinterval(x)
 
 Return the decoration of interval x.
 
-Return the default decoration of a `BareInterval`.
-
 See [`Decoration`](@ref) for more infomation.
 """
 decoration(x::Interval) = x.decoration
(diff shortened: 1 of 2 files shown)


src/intervals/construction.jl
```diff
@@ -275,6 +302,16 @@ function bareinterval(x::Interval)
     return x.bareinterval
 end
 bareinterval(::Type{T}, x::Interval) where {T} = bareinterval(T, bareinterval(x))
+
+"""
+    decoration(x::Interval)
+
+Return the decoration of interval x.
+
+Return the default decoration of a `BareInterval`.
```
OlivierHnt: This sentence feels out of place no?

OlivierHnt (changes requested):

src/intervals/construction.jl
Kolaru: Absolutely, I copy pasted too hard.

PR #759, opened 2026-05-25
Avoid inlining from `@warn`

schillic:
See https://github.com/JuliaLang/julia/issues/61900#issuecomment-4533696985. Kudos to @KristofferC.


OlivierHnt:
Looks good to me, thanks.

PR #753, opened 2026-04-15
Eigenbox

blegat:
As discussed during the meeting, the current implementation in the LinearAlgebra extension is incorrect for large intervals while the implementation in IntervalLinearAlgebra works for large interval but is its computation of `eigvals` for the matrix of centers and radius is just using floating point computation. By combining both, we have a fully rigorous eigen function for matrices of intervals.

This was done by Claude Code, still need to read it carefully

- [x] Not sure why Claude named it eigenbox, should be renamed to eigen
- [x] Add tests
- [x] Allow the user to choose it with `alg`

PR #764, opened 2026-06-21
Eigenbox

blegat:
As discussed during the meeting, the current implementation in the LinearAlgebra extension is incorrect for large intervals while the implementation in IntervalLinearAlgebra works for large interval but is its computation of `eigvals` for the matrix of centers and radius is just using floating point computation. By combining both, we have a fully rigorous eigen function for matrices of intervals.

This was done by Claude Code but I have done a pass too, it looks good.

- [x] Add tests
- [x] Allow the user to choose it with `alg`

Closes https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/753

### DIFF
--- a/docs/src/manual/usage.md
+++ b/docs/src/manual/usage.md
@@ -152,6 +152,28 @@ and deal properly with the decorations of intervals.
 
 
 
+### Linear algebra
+
+Interval matrices support standard linear algebra operations from the `LinearAlgebra`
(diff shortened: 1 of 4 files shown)


blegat:
@OlivierHnt kind reminder :)

OlivierHnt:
I have not forgotten but I do not know the math behind these algorithms so I need to find time to look into that.

Btw I put your fix to disambiguate `mul!` for Julia v1.10 in #775.

PR #769, opened 2026-07-23
ForwardDiff with Interval partials

timholy:
This change supports differentiation of functions with interval-valued partials, like

    x, w = 2.0, interval(-0.5, 0.5)
    ForwardDiff.derivative(t -> (x + t*w)^4, 0.0)

Previously, those failed due to a short-circuit in ForwardDiff, triggering

    ERROR: InconclusiveBooleanOperation: The operation
    `[0.0, 0.0]_com == [-0.5, 0.5]_com_NG` cannot be determined
    unambiguously.

This short-circuits the short-circuit, by specializing `Base.iszero` on Interval-containing `ForwardDiff.Dual` and `ForwardDiff.Partials`. No behavior change in IntervalArithmetic itself is needed.

### DIFF
--- a/ext/IntervalArithmeticForwardDiffExt.jl
+++ b/ext/IntervalArithmeticForwardDiffExt.jl
@@ -28,6 +28,32 @@ Base.:(==)(x::Dual, y::Interval) = value(x) == y
 Base.:<(x::Interval, y::Dual) = x < value(y)
 Base.:<(x::Dual, y::Interval) = value(x) < y
 
+# ForwardDiff tests partials for structural zero with `iszero`, whose generic
+# fallback is `x == zero(x)` and therefore throws `InconclusiveBooleanOperation`
+# for a non-thin interval. A partial carries no perturbation exactly when it is
+# (recursively) the thin interval [0, 0], so `isthinzero` is a safe and correct
+# answer.
+#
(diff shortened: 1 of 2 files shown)


OlivierHnt:
I am not sure in what context `ForwardDiff.Dual` and `ForwardDiff.Partials` show up. Ideally we probably want the dual and the number to have the same behavior.
Anyways, it's maybe unlikely that someone uses these types in a very different context so this PR can be a quick temporary "hot fix" and paired up with https://github.com/JuliaDiff/ForwardDiff.jl/pull/667 so that `_iszero` (the helper function that checks exact thin equality with zero) lives inside `ForwardDiff`.

Kolaru:
For the rest of the extension, the strategy has been to reimplement the short-circuiting behaviors, removing the short circuit altogether (that's why there is a version of `^` there).

This was a conscious choice: we unfortunately do not have a function testing if "this value can be discarded for the purpose of addition" (which is how ForwardDiff uses `iszero` and what the proposed `_iszero` means), so there is no way to have a guarantee that `iszero` is used only in this context.

Therefore, if it is a reasonable amount of work, I would keep de-short-circuiting ForwardDiff for now. Finding a more general solution would be great, but it probably requires a deep discussion with the ForwardDiff team.

timholy:
Sorry for the delay. The new version follows @Kolaru's recommended design.

PR #770, opened 2026-07-23
Add ThickNumbers extension

timholy:
ThickNumbers.jl (https://github.com/HolyLab/ThickNumbers.jl) provides a general interface for "numbers with width," of which intervals are the premier example. Other examples include Gaussian random variables and other implementations of "fuzzy numbers." While this does not make this package's intervals a subtype of `ThickNumber`, it does add a compatibility shim that is sufficient for at least one tested non-trivial consumer.

### DIFF
--- a/ext/IntervalArithmeticThickNumbersExt.jl
+++ b/ext/IntervalArithmeticThickNumbersExt.jl
@@ -1,0 +1,100 @@
+module IntervalArithmeticThickNumbersExt
+
+import ThickNumbers as TN
+import IntervalArithmetic as IA
+
+# `IA.Interval <: Real`, so it cannot subtype `TN.ThickNumber`; it opts into the
+# ThickNumbers interface via the `isthick` trait instead.
+TN.isthick(::Type{<:IA.Interval}) = true
+
+# Required interface
+
+TN.loval(x::IA.Interval) = IA.inf(x)
+TN.hival(x::IA.Interval) = IA.sup(x)
+
+TN.lohi(::Type{IA.Interval{T}}, lo, hi) where {T} = IA.interval(T, lo, hi)
+TN.lohi(::Type{IA.Interval}, lo, hi) = IA.interval(lo, hi)
+
+TN.basetype(::Type{<:IA.Interval}) = IA.Interval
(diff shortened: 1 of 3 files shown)


ext/IntervalArithmeticThickNumbersExt.jl
```diff
@@ -0,0 +1,100 @@
...
+TN.issupset_tn(a::IA.Interval, b::IA.Interval) = IA.issubset_interval(b, a)
+TN.is_strict_supset_tn(a::IA.Interval, b::IA.Interval) = IA.isstrictsubset(b, a)
+
+# `ThickNumbers` derives `Base.in(x::Real, a::ThickNumber)` from `loval`/`hival` for every
+# `ThickNumber` subtype, but a type that opts in via `isthick` (like `IA.Interval`) does not
+# inherit that dispatch and must define it directly. Unlike `IA.Interval == IA.Interval` or
+# `in_interval(::Interval, ::Interval)`, a real point's membership in an interval is always
+# decidable, and `IA.in_interval` already computes exactly that (`Base.in(::Interval,
+# ::Interval)` is the one IA purposely leaves unsupported, for `issubset_interval` instead).
+Base.in(x::Real, a::IA.Interval) = IA.in_interval(x, a)
```
OlivierHnt, line 69: This looks like type piracy, no?

OlivierHnt: In the sense that this method has nothing to do with ThickNumber, and the package extension does not own `Interval`. In any case, it's probably not a feature we'd want to add to IA.jl.

OlivierHnt:
Aside from my comment on `Base.in`; I think this package extension is ready to roll.

timholy:
Thanks. To explain: I'm deliberately dragging my feet on this. I want to make it the last of this sequence of changes, because I don't want to saddle you with something that needs maintenance but which no one actually uses.

#778 is the only real thing I'm waiting on here. I'm also thinking of making breaking changes to ThickNumbers to bring it into better compliance with the choices made here.

#779 is not blocking me, I just noticed it and think this is a rare case where ThickNumbers may have gotten to a better answer than IA.

Issue #585, opened 2023-09-28
A common "thick numbers" package?


timholy:
CC @giordano, @devmotion (and others?)

I've not been `Watching` this package but it looks like interesting and important changes are afoot, and I wonder if it's an opportunity for some shared development among packages like Measurements and possibly even dual-number packages like ForwardDiff (and probably more, feel free to CC others). All of these packages share a somewhat-awkward desire to do math on objects that act like numbers but also have some "thickness" to them. (In ForwardDiff the thickness is infinitesimal, so maybe it's not quite as relevant to this discussion.) All of these packages have been tempted to inherit from `Real`, but many have recognized that this has some issues. I wonder if it's time to create a core package---let me throw out `ThickNumbers.jl` as a possible name---that defines abstract types and exports common operators, traits, fallbacks, errors, etc., but leaves all concrete implementations to downstream packages.

To make this a bit concrete, here's an example from the development (`master`) branch of IntervalArithmetic:

```julia
julia> using IntervalArithmetic, IntervalArithmetic.Symbols

julia> 1..3 == 1..3
ERROR: ArgumentError: == is purposely not supported, use ≛ instead
Stacktrace:
 [1] ==(::Interval{Float64}, ::Interval{Float64})
   @ IntervalArithmetic ~/.julia/dev/IntervalArithmetic/src/intervals/interval_operations/boolean.jl:31
 [2] top-level scope
   @ REPL[2]:1

julia> 1..3 ≛ 1..3
true

julia> 1..3 ≛ 1..3.1
false
```

Note the use of a new operator `≛`. My understanding is that IntervalArithmetic is trying to adhere more strictly to the fundamental principle that `f(X, Y)` (here `f` is `==`) includes all possible values of `f(x, y)` for any `x ∈ X, y ∈ Y`; since almost any *independent* choice of `x, y` are not equal, testing interval equality is nearly useless. One could make a case for `Interval{Bool}(false, true)`, but having the operation be undefined is at least safe, and gets you to the same result (an error) that putting `a == b` inside an `if` statement would generate even if it did return `Interval{Bool}(false, true)` (because `if` requires a `Bool`). This is just one example among many.

[...]

PR #771, opened 2026-07-30
fix: Handle empty and unbounded intervals in `mince`

timholy:
The nodes of `mince` are interpolated between the bounds of the interval, which is not possible for a degenerate one: `LinRange` between infinite endpoints yields `NaN`. An empty interval and an NaI now produce `n` copies of themselves, while an unbounded interval is rejected, `bisect` being the operation that splits one.

Splitting an unbounded interval used to return pieces that do not cover it, e.g. `mince(interval(0, Inf), 2)` gave two copies of `[Inf, Inf]`.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>

This PR is ground-work for the empty interval representation proposed in https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/585#issuecomment-5102354395; I ran a "soundness check" for operations that could produce `NaN`, and this is the one thing that popped up. **Note**: it is a behavior change since it now throws for unbounded intervals.

### DIFF
--- a/src/intervals/interval_operations/bisect.jl
+++ b/src/intervals/interval_operations/bisect.jl
@@ -32,7 +32,8 @@ end
 """
     mince(x, n)
 
-Split an interval `x` in `n` intervals of the same diameter.
+Split an interval `x` in `n` intervals of the same diameter. An unbounded `x`
+has no such splitting and is rejected; use [`bisect`](@ref) instead.
 
 Split the `i`-th component of a vector `x` in `n[i]` intervals of the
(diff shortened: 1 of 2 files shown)


OlivierHnt:
Oh. I would not consider this breaking, it's a bug fix to me.
@Kolaru @lbenet any objection?

Kolaru (approved):
I agree, to me it looks like a bugfix.

I had a look, and everything looks fine to me.

Issue #637, opened 2024-03-22
Warning seriously (using Logging.jl and with user settings)


Kolaru:
With the emergence of the NG flag and similar considerations, it seems like implementing a proper logging system is becoming increasingly pressing.

Amongst the thing that we would want to *optionally* warn the user are:
- Creating an interval with NG
- Creating an interval with decoration that is not `:com`
- Creating an interval from floating point numbers with `@interval` (related to #627)
- Using a function that is not tight (like power which has a much slower tight version)
- Creation an invalid interval
- Creating an empty interval

Depending on their needs, users will probably be interested in enabling or disabling different warnings.

The logging will need to include the full stracktrace in most case, for example to quickly pinpoint where a computation lose its guarantee.

Logging.jl probably allow to do all that.

Looking at the list (please comment if you see more case that could do with a warning) it seems like most of the warnings would fall into the interval constructors, so it may not be that hard to do.

timholy:
Salvage-worthy ideas from #772:
- Creating an invalid interval: this seems like mostly a NaN issue to me
- One could configure the warning via Preferences.jl. It could be made overhead-free if everything is a compile-time preference. Note that Preferences.jl distinguishes itself from alternatives like `ENV` in that compile-time preferences are stored in the package cache header, and thus interact seamlessly with Julia's stale-precompile-caching mechanisms.

Use of Preferences.jl might be a potential enhancement of `configure`

timholy:
For what little it's worth (nothing), I think IEEE 1788-2015 got it wrong when it comes to constructing `interval(4, 3)`: it should be a perfectly-normal way to construct an empty interval. So I am strongly in favor of making it possible to omit the warning in that case. A compile-time preference would mean that no performance is lost.

PR #772, opened 2026-07-30
Reduce the cost of `hull` and `intersect_interval`

timholy:
This follows the discussion in #585: can empty intervals be handled more cheaply, and can variadic set operations inspect their arguments only once? Both are possible after changing the internal representation of an empty interval, as @OlivierHnt suggested.

The prerequisite `mince` fix is already merged in #771.

## Empty-interval representation

An empty `BareInterval{<:AbstractFloat}` was stored as `(NaN, NaN)`. Every call to `inf`, `sup`, or `bounds` then had to translate those fields into the bounds required by IEEE 1788. Empty intervals are now stored as `(typemax(T), typemin(T))`, as they already were for `Rational`:

```julia
sup(x::BareInterval) = x.hi
bounds(x::BareInterval) = (x.lo, x.hi)
inf(x::BareInterval{<:AbstractFloat}) =
    ifelse(iszero(x.lo), copysign(x.lo, -1), x.lo)
```

The only remaining normalization restores the standard's `-0` lower bound.

The implementation now relies on stored bounds never being `NaN`. `is_valid_interval` rejects `NaN` in checked constructors, and `--check-bounds=yes` checks the invariant in the internal constructor. The check compiles away otherwise because `check_bounds` is part of Julia's precompilation cache key.

The package's exported callables were swept over empty, entire, thin-zero, half-bounded, and ordinary intervals at the supported bound types. The only operation that produced a stored `NaN` was `mince` on empty or unbounded intervals, which #771 fixed.

## Set operations

The new representation removes the special empty case from both operations. The empty bounds are neutral under `min` and `max`, while an empty operand of `intersect_interval` makes the computed lower bound exceed the upper bound:

```julia
hull(x::BareInterval{T}, y::BareInterval{T}) where {T<:NumTypes} =
    _unsafe_bareinterval(T, min(x.lo, y.lo), max(x.hi, y.hi))
```

For decorated intervals, the minimum decoration is `ill` exactly when an operand is NaI. The same minimum supplies the result for `dec = :auto`. After excluding NaI, `dec = :default` is always `trv`, so it needs no further decoration adjustment.

The variadic methods now reduce the bare intervals and compute the decoration minimum and `isguaranteed` conjunction once. Dedicated variadic methods for `BareInterval` also fix a `MethodError`: the previous generic fallback passed a `dec` keyword to a method that does not accept one.

## Benchmarks

Measured on an i7-14700KF with Julia 1.12.6:

| Operation | master | branch | IntervalFastMath |
| --- | ---: | ---: | ---: |
| `hull(bare, bare)` | 2.47 ns | 1.81 ns | 1.82 ns |
| `hull(Interval, Interval)` | 6.18 ns | 2.18 ns | 1.82 ns |
| `hull(w, x, y, z)` | 16.97 ns | 3.78 ns | 2.82 ns |
| `intersect_interval(Interval, Interval)` | 6.34 ns | 2.01 ns | — |

In loops over 1000-element vectors, consuming both bounds of every result:

| Operation | master | branch | speedup |
| --- | ---: | ---: | ---: |
| `reduce(hull, v)` | 9.76 ns | 2.66 ns | 3.7× |
| `out[i] = hull(x[i], y[i])` | 6.26 ns | 1.75 ns | 3.6× |
| `reduce(intersect_interval, v)` | 6.09 ns | 1.08 ns | 5.7× |
| `out[i] = intersect_interval(x[i], y[i])` | 6.32 ns | 2.01 ns | 3.1× |

Cheaper bound access also improves other operations: addition and subtraction by about 1.4×, multiplication by 1.24×, and `mid` by 1.3× on `Interval` and 2.3× on `BareInterval`. There is no allocation change for `BigFloat` or `Rational`.

## Testing

New tests cover the representation invariants, empty and NaI operands, decoration semantics, variadic and pairwise agreement, `isguaranteed` propagation, and bound-type promotion.

### DIFF
--- a/src/intervals/interval_operations/set_operations.jl
+++ b/src/intervals/interval_operations/set_operations.jl
@@ -13,39 +13,57 @@ end
 """
     intersect_interval(x, y; dec = :default)
 
-Returns the intersection of the intervals `x` and `y`, considered as (extended)
-sets of real numbers. That is, the set that contains the points common in `x`
-and `y`.
+Return the intersection of `x` and `y`, considered as extended sets of real
+numbers.
 
-The keywork `dec` argument controls the decoration of the result. It can be
-either a specific decoration, or one of two following options:
+The keyword argument `dec` controls the decoration of the result. It can be a
+specific decoration or one of the following two options:
 - `:default`: if at least one of the input intervals is `ill`,
   then the result is `ill`, otherwise it is `trv` (Section 11.7.1).
-- `:auto`: the ouptut has the minimal decoration of the inputs.
+- `:auto`: the output has the minimal decoration of the inputs.
 
 Implement the `intersection` function of the IEEE Standard 1788-2015 (Section 9.3).
 """
 function intersect_interval(x::BareInterval{T}, y::BareInterval{T}) where {T<:NumTypes}
-    lo = max(inf(x), inf(y))
-    hi = min(sup(x), sup(y))
-    if lo > hi
-        return emptyinterval(BareInterval{T})
-    else
-        return _unsafe_bareinterval(T, lo, hi)
-    end
+    # An empty interval has bounds `(typemax(T), typemin(T))`, so an empty
+    # operand makes `lo > hi` without requiring a separate check.
+    lo = max(x.lo, y.lo)
+    hi = min(x.hi, y.hi)
+    lo > hi && return emptyinterval(BareInterval{T})
+    return _unsafe_bareinterval(T, lo, hi)
 end
 intersect_interval(x::BareInterval, y::BareInterval) = intersect_interval(promote(x, y)...)
 
 function intersect_interval(x::Interval{T}, y::Interval{S}; dec = :default) where {T<:NumTypes,S<:NumTypes}
-    isnai(x) | isnai(y) && return nai(promote_type(T, S))
-    r = intersect_interval(bareinterval(x), bareinterval(y))
     d = min(decoration(x), decoration(y))
+    d == ill && return nai(promote_type(T, S)) # one of the inputs is an NaI
+    r = intersect_interval(_bareinterval(x), _bareinterval(y))
     t = isguaranteed(x) & isguaranteed(y)
+    dec === :default && return _unsafe_interval(r, trv, t)
     return _set_decoration(_unsafe_interval(r, d, t), dec)
 end
 
 intersect_interval(x, y, z, w...; dec = :default) =
     reduce((a, b) -> intersect_interval(a, b; dec = dec), (x, y, z, w...))
+
+# Bare intervals carry no decoration, so they take no `dec` keyword.
+intersect_interval(x::BareInterval, y::BareInterval, z::BareInterval, w::Vararg{BareInterval,N}) where {N} =
+    reduce(intersect_interval, (x, y, z, w...))
+
+# Share the decoration checks across all arguments. `Vararg{Interval,N}` is
+# required over `Interval...`: without the `N` the method is not specialized on
+# the number of arguments past six, and the reductions below then allocate and
+# dispatch dynamically.
+function intersect_interval(x::Interval, y::Interval, z::Interval, w::Vararg{Interval,N}; dec = :default) where {N}
+    xs = (x, y, z, w...)
+    r = mapreduce(_bareinterval, intersect_interval, xs)
+    d = mapreduce(decoration, min, xs)
+    d == ill && return nai(numtype(r)) # one of the inputs is an NaI
+    t = mapreduce(isguaranteed, &, xs)
+    dec === :default && return _unsafe_interval(r, trv, t)
+    return _set_decoration(_unsafe_interval(r, d, t), dec)
+end
+
 intersect_interval(x::Complex, y::Complex; dec = :default) =
     complex(intersect_interval(real(x), real(y); dec = dec), intersect_interval(imag(x), imag(y); dec = dec))
 intersect_interval(x::Real, y::Complex; dec = :default) =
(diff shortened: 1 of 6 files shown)


src/intervals/interval_operations/set_operations.jl
```diff
@@ -13,39 +13,54 @@ end
...
+# Share the decoration checks across all arguments.
+function intersect_interval(x::Interval, y::Interval, z::Interval, w::Interval...; dec = :default)
+    xs = (x, y, z, w...)
+    r = reduce(intersect_interval, map(_bareinterval, xs))
+    d = reduce(min, map(decoration, xs))
+    d == ill && return nai(numtype(r)) # one of the inputs is an NaI
+    t = reduce(&, map(isguaranteed, xs))
+    dec === :default && return _unsafe_interval(r, trv, t)
+    return _set_decoration(_unsafe_interval(r, d, t), dec)
+end
```
OlivierHnt: Isn't `mapreduce` relevant here?

src/intervals/interval_operations/set_operations.jl
```diff
@@ -60,30 +75,48 @@ Return the interval hull of the intervals `x` and `y`, considered as (extended)
...
+# Share the decoration checks across all arguments.
+function hull(x::Interval, y::Interval, z::Interval, w::Interval...; dec = :default)
+    xs = (x, y, z, w...)
+    r = reduce(hull, map(_bareinterval, xs))
+    d = reduce(min, map(decoration, xs))
+    d == ill && return nai(numtype(r)) # one of the inputs is an NaI
+    t = reduce(&, map(isguaranteed, xs))
+    dec === :default && return _unsafe_interval(r, trv, t)
+    return _set_decoration(_unsafe_interval(r, d, t), dec)
+end
```
OlivierHnt: Same comment about `mapreduce`

OlivierHnt:
1. The use of `--check-bounds` is clever, but I am not sure it's good to conflate its meaning with the `NaN` check of interval bounds.
One hope I had was that the `NaN` check could occur before the call to `_unsafe_bareinterval`, say somewhere in the (bare)interval constructor pipeline. Then maybe it's always safe (or anyways for `hull` and some others) to skip the `NaN` check.
Perhaps this is also supported by this sentence:
>The only operation that produced a stored NaN was mince on empty or unbounded intervals, which https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/771 fixed.

2. ~There is still a little performance gap between IFM and this PR. Is this due to the normalization of `0.0`? Then maybe, as 1. above, the `normaliszero` call could be handled outside of `_unsafe_bareinterval`.~

EDIT: Oh nevermind, the performance gap is for `Interval` input, so that's just coming from the handling of `Decoration`. So checking `NaN` has a noticeable impact on the perf, whereas the normalization of `0.0` is completely innocuous.
Interesting 🤔

timholy:
Yes, the main idea here is to never produce NaNs so you never have to check for them (this is why #771 had to come first). That check is expensive because it introduces branching, and branches break things like vectorization (https://en.wikipedia.org/wiki/Automatic_vectorization, see sections on control flow). Opting into the check gives you a way to easily check the validity of your code; but to be fast, the decision about whether to omit the safety check must be made at *compile time*.

The trick: if it's a package-wide setting rather than a type parameter, it has to depend on a setting that makes it into Julia's "is this precompiled cache valid given the settings for this session?" query. Otherwise Julia might end up picking, e.g., an "unsafe" precompiled cache in a project where you've deliberately enabled the check. `--check-bounds` is one of only a few options; the others correspond to `--pkgimages`, `-g` (aka `--debug-info`), `--inline`, `-O` (optimization level). Arguably `-g` or `-O` might make more sense.

There is one alternative to Julia's own settings: [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl), using a compile-time constant.

If you agree with the overall idea of setting this choice at compile time, which of these options seems best to you? Any of these could be activated as a special build step on CI, if you're prepared to wait for that one job (it will be a little slower, but not slower than it is now).

OlivierHnt:
>  [...] to never produce NaNs so you never have to check for them [...]

Then why not modifying `_unsafe_bareinterval` by removing the check for NaN entirely, and putting this check upstream in the `interval` and `bareinterval` constructor pipeline?

EDIT: to be clear; the idea would be to make `_unsafe_bareinterval` as close to a no-op as possible. The cost of checking for an input NaN when constructing an interval should be more than marginal.
Either the entire call `_normalisebound(a)` can be put upstream in the constructor and also perhaps where needed at specific location across the source code when the operation might "de-normalise". Or, if that's too much, and since you already check that NaN is already out of scope, then only the `_assert_not_nan(a)` stays a mere `isnan` call.

OlivierHnt:
> That check is expensive because it introduces branching [...]

Really? I mean it's just `isnan(x::AbstractFloat) = (x != x)::Bool`. Perhaps that's due to `::Bool`; that would be unfortunate.

OlivierHnt:
I did not try it, but I now convinced myself that we can just drop the `_assert_nan` entirely.. If `is_valid_interval` already produces an empty interval for `BareInterval` (and an NaI for `Interval` which is stored as an empty interval internally) then everything should be safe. (unless someone hacks into the private functions, but that is always strongly discouraged) 

What am I missing?

timholy:
It's completely optional. I added it as an escape hatch to catch regressions; if any future change in the package does start producing NaNs, turning that check on will catch them directly at the source rather than letting them cascade through a bigger computation. That's the only advantage it provides, and that may not be much of one since you can also just try to trace back once you observe NaNs. To use this to isolate the source of the NaNs would require starting a fresh Julia session with the option turned on.

> Really? I mean it's just isnan(x::AbstractFloat) = (x != x)::Bool. Perhaps that's due to ::Bool; that would be unfortunate.

It's not the evaluation of the Bool that's expensive; it's the change it causes in the compiled code. If you've never seen the "train switch station analogy," you might want to [read the first answer to this question](https://stackoverflow.com/questions/11227809/why-is-conditional-processing-of-a-sorted-array-faster-than-of-an-unsorted-array); it's entertaining and where I first learned about these issues. (Claude helped me find it again, and informed me that it's still the most upvoted stackoverflow question of all time!)

I'm happy to ditch the check altogether if that's what you prefer.

OlivierHnt:
Thx for the link. I thought that our `NaN` machinery (on `master`) was not branching (eg https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/17865d8f58634dfc604c60c0009e64181dc20e5a/src/intervals/interval_operations/numeric.jl#L47), we use `ifelse` as much as possible.
That's also the case for `normalisezero`, which here seems "free".

My vote is to remove the mechanism with `--check-bounds=true` (it feels too hacky for me)... let me mull over this for a bit 😅 

For now there is just my two minor comments on the PR.

timholy:
The branching is in `hull`: https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/17865d8f58634dfc604c60c0009e64181dc20e5a/src/intervals/interval_operations/set_operations.jl#L72

Here's a tally of the LLVM:
|                            | `br` | `select` | `fcmp` | IR lines |
| -------------------------- | ---: | -------: | -----: | -------: |
| `sup(::BareInterval)` before |    0 |        1 |      1 |        9 |
| `sup(::BareInterval)` after  |    0 |        0 |      0 |        7 |
| `inf(::BareInterval)` before |    0 |        2 |      2 |       12 |
| `inf(::BareInterval)` after  |    0 |        1 |      1 |       10 |
| `hull(bare, bare)` before    |    3 |        3 |      4 |       38 |
| `hull(bare, bare)` after     |    0 |        1 |      1 |       18 |
| `intersect_interval(bare, bare)` before | 3 |    7 |      8 |       42 |
| `intersect_interval(bare, bare)` after  | 3 |    1 |      2 |       26 |

OlivierHnt:
Ah of course, for some reason I got fixated on the normalizations involved in `inf` and `sup`... thanks!

timholy:
Glad it helps. Let me know what you decide about
> remove the mechanism with --check-bounds=true (it feels too hacky for me)... let me mull over this for a bit

and then I'll make any further changes. (I addressed your two review points in the most recent force-push.)

OlivierHnt:
@Kolaru is going to run some checks for IntervalRootFinding.jl just to be sure.

I think the mechanism `--check-bounds=true` will be removed before merging the PR.

Kolaru:
I ran the tests of IntervalRootFinding and TaylorModels with this PR and everything seems fine. So it looks like this PR doesn't break much downstream.

Regarding `--check-bounds`, I agree with Olivier that it is a bit strange: I would assume that this options refers exclusively to array bounds, mostly in the context of accessing part of the memory. Despite sharing the name for bounds, I think that we are doing something fundamentally different.

However, the idea could be added to the mechanism proposed in #637, together with other checks at construction (e.g. loss of guarantee).

OlivierHnt:
Fantastic thanks for running these checks.

@timholy once you removed the `--check-bounds` safeguard; we are good to go to merge the PR. 🎉

timholy:
Done! I'm happy with it once it passes CI, merge at will. I also added some notes to #637 and linked back here.

PR #773, opened 2026-08-06
Fast paths for `ExactReal` with `BareInterval`

timholy:
An `ExactReal` is a point, so `+`, `-`, `*`, `/` and `\` against a `BareInterval` of the same number type apply the operation to the two endpoints directly, rather than promoting the point to a thin interval and running the interval-interval operator on it.

Each method takes that route only where it is bitwise identical to the promotion, and delegates the rest: a non-finite point, whose thin interval is ill-formed; an unbounded interval under `*`, where `0 * Inf` is flavor dependent; and division by a zero point. Rounding still goes through `@round`, so the results are unchanged under every rounding mode. Mixed number types keep the promotion, which is where the point is rounded to the interval's number type.

The type parameter is bounded by `NumTypes` so that these methods are more specific than the generic `ExactReal`/`BareInterval` methods, whose own parameters are bounded by `Real` and `NumTypes`; the tests assert that dispatch selects them.

Point times interval costs 3.2 ns against the promotion route's 6.4-8.2 ns with rounding `:none`, and 6.7 ns against 9.1-10.2 ns with `:correct`.

Assisted-by: Claude Opus 5 (claude-opus-5) <noreply@anthropic.com>

### DIFF
--- a/src/intervals/exact_literals.jl
+++ b/src/intervals/exact_literals.jl
@@ -261,6 +261,57 @@ for f ∈ (:+, :-, :*, :/, :\, :^)
     @eval Base.$f(x::ExactReal, y::BareInterval) = $f(promote(x, y)...)
 end
 
+# These specializations improve efficiency, but always agree precisely with
+# the promoting fallbacks above. Several circumstances fall back to promotion: a
+# non-finite point (whose thin interval is not even valid), an unbounded interval
+# times a point (`0 * ∞` is flavor dependent), and division by zero (likewise).
+# The type parameter is bounded so that these methods are more specific than the
+# generic `Base.$f(::ExactReal, ::BareInterval)`, whose own parameters are bounded
+# by `Real` and `NumTypes`.
+
+function Base.:+(x::BareInterval{T}, y::ExactReal{T}) where {T<:NumTypes}
+    v = y.value
+    isfinite(v) || return +(promote(x, y)...)
+    isempty_interval(x) && return x
+    return @round(T, inf(x) + v, sup(x) + v)
+end
+Base.:+(x::ExactReal{T}, y::BareInterval{T}) where {T<:NumTypes} = y + x
+
+function Base.:-(x::BareInterval{T}, y::ExactReal{T}) where {T<:NumTypes}
+    v = y.value
+    isfinite(v) || return -(promote(x, y)...)
+    isempty_interval(x) && return x
+    return @round(T, inf(x) - v, sup(x) - v)
+end
+
+function Base.:-(x::ExactReal{T}, y::BareInterval{T}) where {T<:NumTypes}
+    v = x.value
+    isfinite(v) || return -(promote(x, y)...)
+    isempty_interval(y) && return y
+    return @round(T, v - sup(y), v - inf(y))
+end
+
+function Base.:*(x::BareInterval{T}, y::ExactReal{T}) where {T<:NumTypes}
+    v = y.value
+    isempty_interval(x) && return x # `isbounded` is true for the empty interval
+    (isfinite(v) & isbounded(x)) || return *(promote(x, y)...)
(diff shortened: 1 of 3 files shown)


OlivierHnt:
Can't all of this be captured by something in the vein of `Base.convert(::Type{BareInterval{T}}, x::ExactReal{T}) where {T} = _unsafe_bareinterval(T, x.value, x.value)`?

timholy:
Not quite; your approach saves a validation, but
```julia
julia> iv1, iv2 = bareinterval(3, 3), bareinterval(3, 4)
([3.0, 3.0], [3.0, 4.0])

julia> @code_lowered iv1*iv2
CodeInfo(
1 ── %1  = IntervalArithmetic.isempty_interval
│    %2  =   dynamic (%1)(x)
└───       goto #4 if not %2
2 ── %4  = x
└───       return %4
3 ──       goto #4
4 ┄─ %7  = IntervalArithmetic.isempty_interval
│    %8  =   dynamic (%7)(y)
└───       goto #7 if not %8
5 ── %10 = y
└───       return %10
6 ──       goto #7
7 ┄─ %13 = IntervalArithmetic.isthinzero
│    %14 =   dynamic (%13)(x)
└───       goto #10 if not %14
8 ── %16 = x
└───       return %16
9 ──       goto #10
10 ┄ %19 = IntervalArithmetic.isthinzero
│    %20 =   dynamic (%19)(y)
└───       goto #13 if not %20
11 ─ %22 = y
└───       return %22
12 ─       goto #13
13 ┄ %25 = IntervalArithmetic.:&
│    %26 = IntervalArithmetic.isbounded
│    %27 =   dynamic (%26)(x)
│    %28 = IntervalArithmetic.isbounded
│    %29 =   dynamic (%28)(y)
│    %30 =   dynamic (%25)(%27, %29)
└───       goto #16 if not %30
14 ─ %32 = IntervalArithmetic._mult
│    %33 = IntervalArithmetic.:*
│    %34 =   dynamic (%32)(%33, x, y)
└───       return %34
15 ─       goto #16
16 ┄ %37 = IntervalArithmetic._mult
│    %38 = IntervalArithmetic._unbounded_mul
│    %39 =   dynamic (%37)(%38, x, y)
└───       return %39
)
```
shows a lot of unnecessary work (it's not even showing the four-product min/max), given that a point only needs one signbit and two products. It cuts the time approximately in half, which is a game-changer for my application.

To be clear, it's totally OK if you say "no," I don't want to push you into something you don't want.

OlivierHnt:
I see. Oh no, I am not opposed in principle, but I always try to fix the root problems (otherwise we end up with too much band-aid...).
I'd say that these specialized functions if we end up having them, they should stay leave in exact.jl since that's where their generic counter parts live.

timholy:
To clarify, you mean move those methods to `src/intervals/exact_literals.jl`, right?

OlivierHnt:
Indeed, sorry. Next to the fallback ones.

timholy:
Done

OlivierHnt:
LGTM on my side. It seems to behave exactly as it does with the previous promotion system; @Kolaru @lbenet thoughts?

EDIT: while we are at it can we also make `+(x::Interval{T}, y::ExactReal{T})` etc. benefit from these methods? If it's not trivial, let's not over burden this PR and merge as is.

OlivierHnt:
I'll merge the PR to keep the ball rolling.

@timholy Just so you know, we'll have a triage meeting next week, and we will discuss your set of PRs before making the release. So if you have more PRs feel free to get them through; #770 is still pending the question of `Base.in`.

timholy:
I've been away (JuliaCon) and so haven't been working on this. I can pick it up again tomorrow.

OlivierHnt:
Indeed @Kolaru was there as well ; you two might have crossed path 🙂
On our side, we had our triage call and everyone is pretty happy with the recent PRs (now part of the new 1.0.11 release).

PR #775, opened 2026-08-24
Rework test suite

### LINKED ISSUE #774: Behaviour of `acot(::Interval)`
I stumbled on a bug
```Julia
julia> acot(bareinterval(0)) # current (v1.0.11)
[1.57079, -1.57079]
```
which I fixed on master so that
```julia
julia> acot(bareinterval(0)) # master
[-1.57079, 1.57079]
```

We did not make a custom method `acot(x::Interval)` since `Interval` is a `<:Real` so it calls `atan( inv( x ) )`. So we get

```Julia
julia> acot(interval(0)) # inconsistent with the above bare interval version
∅_trv

julia> acot(interval(0, 1))
[0.785398, 1.5708]_trv

julia> acot(interval(-1, 0))
[-1.5708, -0.785398]_trv
```

The question is should we make `acot` behave like how `inv` does as per the standard? (note that `acot` itself is not described in the standard)

1. If so, then `acot(bareinterval(0))` should return the empty interval; and we do not need to implement `acot(x::Interval)`.

2. If not, we then consider `acot` as a discontinuous function going to $\pm \pi/2$ as $x \to 0^{\pm}$. The decoration should demote to `def` whenever $0$ is in the input interval.

My first impulse is that Option 1. feels somewhat natural; after all `inv(interval(0))` does not return $(-\infty, \infty)$ but the empty interval.
Also, I wonder if the flavour has an impact on the decision here.

On that last point: (Section 9):
> The behavior of each operation outside the set of common evaluations is flavor-defined, subject to the general
flavor requirements in 7.4, 7.5.

and we have indeed https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/40c30a4eecae55220126502055b227babd4f9286/src/intervals/flavor.jl#L60-L67

OlivierHnt:
Complete rework of the test suite, and fixes for the bugs it surfaced in `src/`.

**Test suite**

- `test/` now mirrors the organization of `src/` and `ext/`. Every extension is covered (DiffRules, IntervalSets, Makie, RecipesBase and SparseArrays previously had no tests). The legacy `test/interval_tests/` is removed, and every assertion was preserved or strengthened.
- ITF1788: the reductions are translated with correctly rounded accumulation, removing the `dot_nearest` `@test_broken`. The `atan2` override is also removed, since the original ITL expectation `dac` is correct and now passes.
- Aqua tests reworked. The cross-dependency ambiguity check no longer depends on which packages earlier test files load.

**Fixes in `src/`**

- **Two-argument `atan`**: the decoration of a box touching the branch cut from above is capped at `dac`, since `com` requires continuity at each point of the box.
  ```julia
  julia> decoration(atan(interval(0.0, 1.0), interval(-2.0, -1.0)))
  dac # was com
  ```
- **`acot`**: uses the convention `acot(0) = π/2`, with the jump at 0 decorated like the branch cut of the two-argument `atan`. An explicit method for `Interval` replaces the Base fallback `atan(inv(x))`.
  ```julia
  julia> acot(interval(0))
  [1.57079, 1.5708]_dac # was ∅_trv
  ```
- **`:ulp` rounding mode**: functions without a CoreMath routine (`cot`, `sec`, `csc`, `acot`, `coth`, `sech`, `csch`, `acoth`) fall back to the `:correct` implementation.
  ```julia
  julia> IntervalArithmetic.configure(rounding = :ulp);

  julia> cot(interval(0.5))
  [1.83048, 1.83049]_com # was UndefVarError
  ```
- **`Piecewise`**:
  - the constructor throws the intended `ArgumentError`s on mismatched numbers of domains, functions or junctions (was `MethodError` or `UndefVarError`).
  - `in_domain` is rewritten. It wrongly returned `false` for inputs starting at a closed endpoint or lying in a later piece.
    ```julia
    julia> pgap = Piecewise(Domain{:closed,:closed}(0, 1) => Constant(1.0), Domain{:closed,:closed}(2, 3) => Constant(2.0));

    julia> decoration(pgap(interval(2.2, 2.5)))
    com # was trv
    ```
  - `isguaranteed` propagation: `Piecewise` and `Constant` no longer return guaranteed intervals from non-guaranteed inputs.
    ```julia
    julia> isguaranteed(pgap(convert(Interval{Float64}, 0.5)))
    false # was true
    ```
  - `intersect_domain`: at equal endpoint values, an open bound now wins (`[0,1] ∩ [0,1)` is `[0,1)`).
  - `Domain` bound validation throws `ArgumentError` (was `ErrorException`). Dead helpers removed and docstrings corrected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


OlivierHnt:
@Kolaru can you look closely at the changes in piecewise.jl? It looked fine to me, but since I do not use this feature I cannot tell for sure.

OlivierHnt:
- **`sample`**: supported for all `Rational` bound types. The exact fraction of the sampled float
  is rationalized, since it can overflow small integer types.
  ```julia
  julia> sample(interval(Int16(1)//Int16(2), Int16(3)//Int16(4)))
  13019//25050 # was InexactError

src/IntervalArithmetic.jl
```diff
@@ -272,7 +272,7 @@ function sample(rng::Random.AbstractRNG, x::Interval{T}) where {T<:NumTypes}
     β = rand(rng, float(T))
     lo = ifelse(lo == typemin(T), _value_min(T), lo)
     hi = ifelse(hi == typemax(T), _value_max(T), hi)
-    val = convert(T, (1 - β) * lo + β * hi)
+    val = _value_round(T, (1 - β) * lo + β * hi)
```
Kolaru: I am not sure that this is the correct fix.

`T` is only the bound type. In principle, it says nothing about the type of the numbers in between, it is fine to return a `Float64` for `Rational` bounds.

Furthermore, I don't know if `rationalize` keeps the distribution uniform.

If we would follow the convention that `Distributions.jl` uses for `Uniform`, then we would always return `Float64`, regardless of the bound type:

```julia
julia> using Distributions

julia> u = Uniform(0//1, 1//1)
Uniform{Rational{Int64}}(a=0//1, b=1//1)

julia> typeof(rand(u))
Float64

julia> u = Uniform(0.0f0, 1f0)
Uniform{Float32}(a=0.0f0, b=1.0f0)

julia> typeof(rand(u))
Float64
```

test/ITF1788_tests/libieeep1788_reduction.jl
```diff
@@ -1,45 +1,45 @@
 @testset "minimal_sum_test" begin
 
-    @test sum([1.0, 2.0, 3.0]) === 6.0
+    @test Float64(sum(big.([1.0, 2.0, 3.0]))) === 6.0
```
Kolaru: Why is this modified?

test/generate_ITF1788.jl
```diff
@@ -83,10 +83,11 @@ functions = Dict(
     "isMember" => x -> "in_interval($x)",
     "cancelPlus" => x -> "cancelplus($x)",
     "cancelMinus" => x -> "cancelminus($x)",
-    "sum_nearest" => x -> "sum($x)",
-    "dot_nearest" => x -> "sum(.*($x))",
-    "sum_abs_nearest" => x -> "sum(abs.($x))",
-    "sum_sqr_nearest" => x -> "sum($x.^2)"
+    # IEEE 1788-2015 requires correctly rounded reductions: accumulate in BigFloat, round once
```
Kolaru: If I understand correctly, with this change we are no longer testing the package functions.

Therefore, I believe that this is incorrect.

test/runtests.jl
```diff
@@ -1,45 +1,39 @@
...
+for entry ∈ ("IntervalArithmetic.jl", "intervals", "piecewise.jl", "display.jl", "symbols.jl", "ext")
+    if isdir(entry)
+        for (root, _, files) ∈ walkdir(entry), f ∈ files
+            push!(repo_tests, joinpath(root, f))
+        end
+    else
+        push!(repo_tests, entry)
+    end
+end
+for f ∈ repo_tests
```
Kolaru: Isn't that unnecessarily complicated?

I believe that the following does the same and won't ever require manual update:

```suggestion
for entry ∈ filter(isfile, walkdir(test))
    if entry[end-3:end] == ".jl" and entry != "runtests.jl"
        @testset "$f" begin
            include(f)
        end
    end
end
```

src/piecewise.jl
```diff
@@ -1,106 +1,71 @@
...
     val1, bound1 = upperbound(d1)
     val2, bound2 = lowerbound(d2)
-
-    val1 == val2 && return !(bound1 == bound2 == :closed)
+    val1 == val2 && return !(bound1 === bound2 === :closed)
     return val1 < val2
 end
 
-in_domain(x::Real, domain::Domain) = rightof(x, lowerbound(domain)) && leftof(x, upperbound(domain))
+in_domain(x::Real, d::Domain) = rightof(x, lowerbound(d)) && leftof(x, upperbound(d))
```
Kolaru: Why? ^^'

src/piecewise.jl
```diff
@@ -1,106 +1,71 @@
...
+function intersect_domain(d1::Domain, d2::Domain)
+    lo, L = _innermost(lowerbound(d1), lowerbound(d2), >)
+    hi, R = _innermost(upperbound(d1), upperbound(d2), <)
+    d = Domain{L,R}(lo, hi)
+    return isempty_domain(d) ? Domain() : d
+end
+# tightest of two like bounds; at equal values an open bound excludes the endpoint
+function _innermost((val1, bound1)::Tuple, (val2, bound2)::Tuple, isinner)
+    val1 == val2 && return (val1, ifelse(bound1 === bound2, bound1, :open))
+    return isinner(val1, val2) ? (val1, bound1) : (val2, bound2)
```
Kolaru: To me this design for `_innermost` is too confusing.

I would go with something like

```julia
function intersect_domain(d1::Domain, d2::Domain)
    lo, L = ifelse(rightof(lowerbound(d1), lowerbound(d2)), lowerbound(d1), lowerbound(d2))
   hi, H = ifelse(leftof(upperbound(d1), upperbound(d2)), upperbound(d1), upperbound(d2))
  return Domain{L, R}(lo, hi)
end
```
(which requires to add one method for `rightof` and `leftof`, but I think it will still be much more readable)

src/piecewise.jl
```diff
@@ -1,106 +1,71 @@
...
 
 """
     Constant(value)
 
-A constant function compatible with interval arithmetic.
-
-Return an interval containing only the value for an interval input,
-and the value directly otherwise.
+Constant function compatible with interval arithmetic: it returns an interval
+enclosing `value` for an interval input, and `value` itself otherwise. In
```
Kolaru: It doesn't technically "enclose" the value for interval inputs, it only convert it to interval (in the sense that there are no extra step taken, `Constant` just build a thin interval)

src/piecewise.jl
```diff
@@ -116,33 +81,46 @@ julia> c(22.2)
...
 """
 struct Constant{T}
-    value::T
+    value :: T
+end
+
+function (constant::Constant)(x::Interval)
+    y = interval(constant.value)
+    return _unsafe_interval(bareinterval(y), decoration(y), isguaranteed(x))
 end
```
Kolaru: I don't think that this is the right handling of the guarantee here:

It doesn't matter if the input interval is correct or not, the output will always be the same. If there is a doubt about the bound of the interval and whether a piece of the function should be used or not, it should go to `PieceWise` instead.

On the other hand, we are currently trusting the value inside the `Constant`... we could argue that this should return a NG interval, except for `Constant{ExactReal}`. I'm not sure what's best here.

test/piecewise.jl
```diff
@@ -0,0 +1,333 @@
...
+    @test ps[1] == (d1, Constant(1.0)) && ps[2] == (d2, identity)
+
+    @test discontinuities(p) == [1]
+    @test discontinuities(p, 1) == [1]
+    @test discontinuities(p2) == Int[]
+    @test discontinuities(p2, 1) == [1]
+    @test discontinuities(p2, 100) == [1]
+end
+
+@testset "domain_string and show" begin
```
Kolaru, line 195: I believe that a function of the form
```
p = Piecewise(
    Domain{:open, :open}(-1, 0) => identity,
    Domain{:open, :open}(0, 1) => identity
)
```
should also be tested here, to see if the single point out of the domain is found.

src/piecewise.jl
```diff
@@ -151,171 +129,118 @@ julia> myabs(-22.3)
...
+    for piece ∈ domains(piecewise)
+        pv, pb = upperbound(piece)
+        (pv < v || (pv == v && !(b === pb === :closed))) && continue
+        plv, plb = lowerbound(piece)
+        (plv < v || (plv == v && (plb === :closed || b === :open))) || return false
+        (hv < pv || (hv == pv && (pb === :closed || hb === :open))) && return true
+        v, b = pv, ifelse(pb === :closed, :open, :closed)
+    end
+    return false
+end
```
Kolaru, line 199: The naming convention suddenly (and drastically) changed in this function compared to the rest.

ext/IntervalArithmeticLinearAlgebraExt.jl
```diff
@@ -235,6 +235,20 @@ function LinearAlgebra.mul!(C::AbstractMatrix{<:RealOrComplexI}, A::AbstractVecO
...
+for (CT, AT, BT) ∈ (
+        (:(AbstractMatrix{<:RealOrComplexI}), :AbstractMatrix,    :_StructuredMatrix),
+        (:(AbstractMatrix{<:RealOrComplexI}), :_StructuredMatrix, :AbstractMatrix),
+        (:(AbstractMatrix{<:RealOrComplexI}), :_StructuredMatrix, :_StructuredMatrix),
+        (:(AbstractVector{<:RealOrComplexI}), :_StructuredMatrix, :AbstractVector))
+    @eval function LinearAlgebra.mul!(C::$CT, A::$AT, B::$BT, α::Number, β::Number)
+        size(A, 2) == size(B, 1) || return throw(DimensionMismatch("The number of columns of A must match the number of rows of B."))
+        return _mul!(IntervalArithmetic.default_matmul(), C, A, B, α, β)
+    end
+end
```
Kolaru: I am a bit confused here. It seems like that this is defining methods to store non-interval matrix multiplication into an interval matrix.

If the inputs (or some of the inputs) are supposed to be interval themselves, I feel like it would be better to type the method for them, and it should remove the ambiguity.

Kolaru (changes requested):
I didn't look into all tests in details (still I commented on what I caught), but I had a look at all the changes in the source files I believe.

src/IntervalArithmetic.jl
OlivierHnt: Ok let's output a float for either `AbstractFloat` or `Rational` bound type.

ext/IntervalArithmeticLinearAlgebraExt.jl
OlivierHnt: Once we drop 1.10 support, we can get rid of these.

src/piecewise.jl
OlivierHnt: Indeed, let's go for "wrapping `value` into an interval".

src/piecewise.jl
OlivierHnt: Ok let's revert this.

src/piecewise.jl
OlivierHnt:
```Julia
julia> f = Piecewise(Domain{:open,:open}(1, 2) => identity, Domain{:open,:open}(2, 3) => identity)
Piecewise function with 2 pieces:
  (1, 2) -> identity
  (2, 3) -> identity

julia> f(2)
ERROR: DomainError with 2:
piecewise function was called outside of its domain (1, 2) ∪ (2, 3)
Stacktrace:
 [1] (::Piecewise{2, 1, Tuple{Domain{…}, Domain{…}}, Tuple{typeof(identity), typeof(identity)}, Tuple{Int64}})(x::Int64)
   @ IntervalArithmetic ~/Documents/dev/IntervalArithmetic/src/piecewise.jl:183
 [2] top-level scope
   @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.

julia> f(interval(1.5, 2.5))
[1.5, 2.5]_trv

julia> f(interval(2))
∅_trv
```

test/ITF1788_tests/libieeep1788_reduction.jl
OlivierHnt: All sums were changed in the generate_ITF1788.jl

OlivierHnt: Oh wait I see. Let us get rid of all these tests since they do not pertain to interval arithmetic.

test/generate_ITF1788.jl
OlivierHnt: See other comment about dropping these tests.

test/piecewise.jl
OlivierHnt: Checked ; see other message.

src/piecewise.jl
Kolaru: We would need a `rightof` that looks something like

```julia
function rightof((a, A)::Tuple, (b, B)::Tuple)
    if a == b
        A === B && return false
        A === :closed && return true
        return false
    else
        return rightof(a, b)
    end
end
```

And the same think for `leftof`.

OlivierHnt: Thanks I'll make the change!

OlivierHnt:
Ok so there are two things worth discussing a bit more:
- I completely dropped support of `sample` for rational bounds because it's just weird with eg sample(interval(1//3))
- I wonder about @Kolaru comment on `Constant{<:ExactReal}` actually... that does feel like the most consistent way to go though adding a bit of inconvenience

Kolaru:
I'm fine with not having sample for Rational intervals.

Regarding `Constant`, the more I think about it the trickier I find it.

I believe that the main point of `Constant` is to avoid the NG flag. Otherwise, users can just use `Returns` and get a float, which will propagate the NG flag when interacting with an interval. It is also nice to have an interval output, but we could achieve that by converting the results in `Piecewise` when needed.

So I would actually be in favor of keeping the current convention that the constant is considered to be exact. I would however make it clear in the docstring and wrap the constant in `exact` at construction, so that it is clear when the function is shown or tested.

As a side note, it would be nice if `exact` and `Constant` supported a string as input and parse it to the best Float64/BigFloat/Interval when needed (then something like `Constant("0.1")` would be fully unambiguous).

Kolaru:
Or actually, now that I am thinking about it, we could deprecate `Constant` entirely in favor of `@exact Returns(x)`.

OlivierHnt:
That kind of make sense to me to drop `Constant` entirely.

PR #778, opened 2026-09-05
Support ForwardDiff duals over BareInterval

timholy:
This has two commits; the second depends on the first, but arguably they should be merged without squashing. I can also split into two separate (sequential) PRs if you prefer.

The main goal here is to support differentiation and arithmetic for bare intervals in nested duals. At present I think this is the last performance work I need to make it possible to adopt IA.


OlivierHnt:
Looks good to me. I wonder whether we just never got around implementing ForwardDiff for `BareInterval`, or if there was a deeper reason?
@Kolaru can you have a look?

ext/IntervalArithmeticForwardDiffExt.jl
```diff
@@ -32,15 +32,21 @@ Base.:<(x::Dual, y::Interval) = value(x) < y
...
 end
 _isthinzero(x::Interval) = isthinzero(x)
+_isthinzero(x::BareInterval) = isthinzero(x)
 _isthinzero(x::Real) = iszero(x)
 _isthinzero(d::Dual) = _isthinzero(value(d)) && all(_isthinzero, partials(d))
+# Use thin-zero checks for interval partials when ForwardDiff supports them.
+@static if isdefined(ForwardDiff, :unwrap_dual)
+    ForwardDiff._iszero_tuple(::Type{<:Union{Interval, BareInterval}}, tup::Tuple) =
+        all(_isthinzero, tup)
+end
```
Kolaru, line 49: I am not sure what is going on here. `unwrap_dual` doesn't seem to be defined on either ForwardDiff v1.0 nor v1.4, so it seems like this is doing nothing.

ext/IntervalArithmeticForwardDiffExt.jl
```diff
@@ -32,15 +32,21 @@ Base.:<(x::Dual, y::Interval) = value(x) < y
 # non-thin intervals. Use a recursive thin-zero test instead. `NestedInterval`
 # supports intervals nested in up to four `Dual` layers.
 const NestedInterval = let
-    U = Interval
+    U = Union{Interval, BareInterval}
```
Kolaru, line 35: Considering the definition of `NestedBareInterval` below, is this necessary?

To me it seems like this is only needed if we allow to nest Interval and BareInterval inside duals, which is unlikely to ever be correct.

ext/IntervalArithmeticForwardDiffExt.jl
```diff
@@ -170,4 +176,114 @@ end
 ForwardDiff.DiffRules._abs_deriv(x::Dual{T,<:Interval}) where {T} =
     Dual{T}(ForwardDiff.DiffRules._abs_deriv(value(x)), zero(partials(x)))
 
+# ForwardDiff support for `BareInterval`, which is not a subtype of `Number`.
+# Real constants and derivative coefficients are treated as exact. Use
+# decorated `Interval`s to track this assumption through `isguaranteed`.
+
```
Kolaru, line 182: This is in direct contradiction with the design philosophy of the package: BareInterval is supposed to give a higher level of guarantee than Interval by disallowing interaction with Real entirely. This would make BareInterval less safe than Interval.

I have not followed everything (busy busy), was this question discussed somewhere? Namely, is there a reason to believe that this is safe within ForwardDiff?

We were (probably) not supporting BareInterval previously for this exact reason: since ForwardDiff is bringing it's own numerical constants, it is not compatible with BareInterval on a quite fundamental level.

If safety can not be guaranteed with these conversions, then I believe that we need a different design here entirely. Right now, I can imagine either a FastInterval flavor that deactivate all rounding and checks (while being a Real), or upstream the concept of ExactReal and tag the exact numerica values in ForwardDiff, and then they should naturally be compatible.

I'm stopping my review here: we should first clarify if the design is valid.

Kolaru (commented):
Thanks a lot for the PR!

As I mention in comment, to me it looks like that this PR introduce a design that is not guaranteed. And (unfortunately T_T), I think we need to figure that out before we can move forward.

ext/IntervalArithmeticForwardDiffExt.jl
timholy: It was added in the hope that someday, someone might merge https://github.com/JuliaDiff/ForwardDiff.jl/pull/667. But I can delete it if desired.

ext/IntervalArithmeticForwardDiffExt.jl
timholy: It would probably be possible to rewrite for specific functions, basically vendoring what ForwardDiff does. I'll have to wait to tackle this until I'm back home and have access to the machine that has my research repo that I'm trying to switch over to IA.

timholy:
I've rebased on current `master` and hopefully addressed @Kolaru's concerns.

timholy:
If you do want to merge this as two commits, LMK and I can squash everything after the first commit together, leaving the first distinct.

